PR #2, opened 2021-07-06
Change locking implementation works

gammazero:
Needed to add items found in the previous cache to the current cache.

When an item is not found in the current cache but is found in the previous cache, that item needs to be put into the current cache.  Doing this requires a write lock on the cache. In order to avoid locking the entire cache, instead just acquire a lock for any key that would affect the same part of the tree.

Put needed to check for existing values in the previous cache.

Also added a unit test for cache rotation.

### DIFF
--- a/store/primary/primary.go
+++ b/store/primary/primary.go
@@ -9,28 +9,26 @@ import (
 	peer "github.com/libp2p/go-libp2p-core/peer"
 )
 
-var _ store.Storage = &rtStorage{}
+// numLocks is the lock granularity for radixtree. Must be power of two.
+const numLocks = 64
 
-// wraps a tree with its own mutex
-type tree struct {
-	lk   sync.RWMutex
-	tree *radixtree.Bytes
-}
+var _ store.Storage = &rtStorage{}
 
 // Adaptive Radix Tree primary storage
 type rtStorage struct {
-	lk        sync.Mutex
-	primary   *tree
-	secondary *tree
-	sizeLimit int
+	branchLocks []sync.Mutex
+	current     *radixtree.Bytes
+	previous    *radixtree.Bytes
+	sizeLimit   int
 }
 
 // New creates a new Adaptive Radix Tree storage
 func New(size int) store.Storage {
 	return &rtStorage{
-		primary:   &tree{tree: radixtree.New()},
-		secondary: nil,
-		sizeLimit: size,
+		branchLocks: make([]sync.Mutex, numLocks),
+		current:     radixtree.New(),
+		previous:    nil,
+		sizeLimit:   size / 2,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


store/primary/primary.go
```diff
@@ -39,25 +37,34 @@ func (s *rtStorage) Get(c cid.Cid) ([]store.IndexEntry, bool) {
...
+	// Search previous if not found in current
+	v, found = s.previous.Get(k)
+
 	// If nothing has been found return nil
 	if !found {
 		return nil, false
 	}
+
+	// Put the value found in the previous tree into the current one.
+	s.current.Put(k, v)
```
adlrocha, line 66: I completely missed this slight detail. Thanks! 🙏

store/primary/primary.go
```diff
@@ -69,47 +76,87 @@ func (s *rtStorage) Get(c cid.Cid) ([]store.IndexEntry, bool) {
...
+		k := string(cids[i].Hash())
+		s.lockBranch(k)
+
+		if s.put(k, in) {
+			addedEntry = true
+		}
+		s.unlockBranch(k)
+	}
+
+	if addedEntry {
```
adlrocha, line 149: Here we are checking the size of the cache after putting the batch, right? Shouldn't we check it after each individual put better? If the batch is too large we may not realize that we are at capacity until we've finished putting every key.

store/primary/primary.go
```diff
@@ -126,38 +173,20 @@ func duplicateEntry(in store.IndexEntry, old []store.IndexEntry) bool {
...
-		s.lk.Lock()
-		// Create new tree and make primary = secondary
-		s.secondary = s.primary
-		s.primary = &tree{tree: radixtree.New()}
-		s.lk.Unlock()
+func (s *rtStorage) lockBranch(k string) {
+	var idx int
+	if k != "" {
+		// bitwise modulus requires that s.locks is power of 2
+		idx = int(k[0]) & (len(s.branchLocks) - 1)
```
adlrocha, line 180: Wouldn't it be better to use `k[len(k)-1]` instead of `k[0]` here? Multihashes use prefixes that may be biasing the lock that we lock, right? (see https://github.com/multiformats/multihash).

Maybe I am wrong, is just a sanity-check.

adlrocha:
This looks great, @gammazero. I was completely missing this detail. Thanks!
I just left a minor comment/sanity-check.

store/primary/primary.go
gammazero: That would be better, but...  We need to lock a "branch" of the tree so that one lock prevents any operation in the same section of the tree, so we need to use the **first** bits of the key.  There is an implicit relationship between this locking strategy and how a radix tree works.

That being said, there is a way to solve both problems at the same time: Store the key backwards.

adlrocha: Oh, now I get it! Really smart! 

I don't think we need to store the key backwards right now, but worth having it in mind (adding a NOTE or something?) in case we start seeing some performance overhead as a result of using the multihash prefix for the lock.

gammazero: I think we want to store the key forwards because that automatically compresses out any common prefix.  Another way to solve this is to use multiple trees, and compute a fast hash of the key to see which tree to use.  Then just lock that tree. I will revisit this problem.

gammazero: Added comment

PR #4, opened 2021-07-07
Use multiple caches to for concurrent access

gammazero:
Instead of having to release a read lock and acquire a write lock and redo a lookup, whenever an entry is put into the current cache from the previous, just exclusively lock the cache.  To lessen lock contention with concurrent access use multiple cache instances over which keys are evenly distributed.

Implement a method to remove single entry (provider-ID + piece-ID) from the cache.

### DIFF
--- a/store/primary/primary.go
+++ b/store/primary/primary.go
@@ -9,26 +9,41 @@ import (
 	peer "github.com/libp2p/go-libp2p-core/peer"
 )
 
-// numLocks is the lock granularity for radixtree. Must be power of two.
-const numLocks = 64
+// concurrrency is the lock granularity for radixtree. Must be power of two.
+const concurrency = 16
 
 var _ store.Storage = &rtStorage{}
 
+// syncCache is a rotatable cache. Multiples instances may be used to decrease
+// concurrent access collision.
+type syncCache struct {
+	current   *radixtree.Bytes
+	previous  *radixtree.Bytes
+	mutex     sync.Mutex
+	rotations int
+}
+
 // Adaptive Radix Tree primary storage
 type rtStorage struct {
-	branchLocks []sync.Mutex
-	current     *radixtree.Bytes
-	previous    *radixtree.Bytes
-	sizeLimit   int
+	cacheSet   []*syncCache
+	rotateSize int
 }
 
 // New creates a new Adaptive Radix Tree storage
 func New(size int) store.Storage {
+	cacheSetSize := concurrency
+	if size < 256 {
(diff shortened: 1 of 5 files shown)


store/primary/primary.go
```diff
@@ -109,55 +173,53 @@ func (s *rtStorage) put(k string, entry store.IndexEntry) bool {
...
+	v, found := cache.Get(k)
+	if !found {
+		return false
+	}
 
-		if s.put(k, in) {
-			addedEntry = true
+	values := v.([]store.IndexEntry)
+	for i := range values {
+		if entry.PieceID == values[i].PieceID &&
```
adlrocha, line 197: Reading this I just realized something. I used a list of tuples for `IndexEntry` because I assumed several providers wouldn't be storing the same CID, and by using a list I prevented having to loop through a map to return every value (I recall adding a NOTE somewhere). However, this assumption may be wrong as CIDs may be representing small blocks that are indexed as part of several larger pieces of content. This may be adding an unnecessary overhead by repeating the providerID in every tuple of the list. 

Again, I don't have real data to support this claim, but just wanted to share it here in case you feel it is worth adding a comment somewhere or filing an issue.

store/primary/primary.go
```diff
@@ -109,55 +173,53 @@ func (s *rtStorage) put(k string, entry store.IndexEntry) bool {
...
-	if addedEntry {
-		// Check if rotation needed
-		s.lockBranch("")
-		if s.current.Len() >= s.sizeLimit {
-			s.unlockBranch("")
-			s.rotateCache()
-		} else {
-			s.unlockBranch("")
-		}
+func (c *syncCache) removeAll(k string) bool {
```
adlrocha, line 212: Is this function being used somewhere? I don't see it exposed through the interface nor used in any other function.

adlrocha (approved):
Looks great! Thanks!

store/primary/primary.go
gammazero: No currently.  I was anticipating its use, but I will remove it since I do not think it will get used since I do not know under what conditions we will want to remove a CID without knowing which provider provides it.

store/primary/primary.go
gammazero: I suppose all situations are possible:
1.  CID is provided by one provider, but with multiple piece IDs
2. CID is provided by multiple providers, with same piece IDs
3. CID is provided by multiple providers, with different piece IDs

If we need to support all of these, then a CID will need to map to a list of (providerID, pieceID) pairs.

PR #5, opened 2021-07-07
Storethehash persistence. Kicking off benchmark

adlrocha:
This PR:
- Adds a storethehash storage for persistence.
- Adds support for the generation of synthetic data for a certain size of CIDs.
- Adds a basic benchmark for single thread `gets` for different file sizes.
- Modifies slightly the`Storage` interface with `Close()` and `Size()`

Next steps after this PR:
- Exhaustive benchmark of storethehash, gather more metrics for benchmark.
- More iteration on `Storage` interface.

//cc @gammazero


store/persistent/storethehash/storethehash.go
```diff
@@ -0,0 +1,136 @@
...
+const DefaultIndexSizeBits = uint8(24)
+const DefaultBurstRate = 4 * 1024 * 1024
+const DefaultSyncInterval = time.Second
+
+type sthStorage struct {
+	dir   string
+	store *sth.Store
+}
+
+func New(dir string) (store.Storage, error) {
```
gammazero, line 28: IMO, it is preferable to return a pointer to the concrete type (`*sthStorage`) here.  That way the caller can decide what interface they want to cast it to if there are multiple interfaces that it may satisfy.

In general, [accept interfaces, return types](https://mycodesmells.com/post/accept-interfaces-return-struct-in-go)

gammazero (commented):
Please see the comments.  Also, you are going to need to rebase before merge (sorry).

store/persistent/storethehash/storethehash.go
adlrocha: It makes total sense! Thank you for the reference, I wasn't aware of this proverb :)

store/store.go
```diff
@@ -17,11 +19,42 @@ type IndexEntry struct {
...
+func Unmarshal(b []byte) ([]IndexEntry, error) {
+	li := []IndexEntry{}
+	err := json.Unmarshal(b, &li)
+	return li, err
+}
+
+// DuplicateEntry checks if the entry already exists in the index. An entry
+// for the same provider but a different piece is not considered
+// a duplicate entry (at least for now)
+func DuplicateEntry(in IndexEntry, old []IndexEntry) bool {
```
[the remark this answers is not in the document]
adlrocha: You are right! Thanks!

adlrocha:
No problem. Thank you for the comments, they were really in point. Let me know if there is anything else before merge.

gammazero (approved):

PR #9, opened 2021-07-09
Intern entries to conserve memory

gammazero:
Store only one copy of each distinct `IndexEntry`, per cache.  This reduces memory consumption as many identical `IndexEntry` values may be mapped to by different CIDs.

- Added `Stats` function to get cache stats. 
- Perform some operations concurrently
- Added Memory usage tests and benchmarks

### DIFF
--- a/store/primary/cache.go
+++ b/store/primary/cache.go
@@ -11,14 +11,27 @@ import (
 // syncCache is a rotatable cache. Multiples instances may be used to decrease
 // concurrent access collision.
 type syncCache struct {
-	current   *radixtree.Bytes
-	previous  *radixtree.Bytes
-	curEnts   *radixtree.Bytes
-	prevEnts  *radixtree.Bytes
+	// CID -> IndexEntry
+	current  *radixtree.Bytes
+	previous *radixtree.Bytes
+
+	// IndexEntery interning
+	curEnts  *radixtree.Bytes
+	prevEnts *radixtree.Bytes
+
 	mutex     sync.Mutex
 	rotations int
 }
 
+type CacheStats struct {
+	Cids           int
+	Values         int
+	UniqueValues   int
+	InternedValues int
+	Rotations      int
+}
+
+// newSyncCache created a new syncCache instance
 func newSyncCache() *syncCache {
 	return &syncCache{
 		current: radixtree.New(),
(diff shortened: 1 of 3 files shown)


store/primary/cache.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+func (c *syncCache) rotate() {
+	c.previous, c.current = c.current, radixtree.New()
+	c.prevEnts, c.curEnts = c.curEnts, radixtree.New()
+	c.rotations++
+}
+
+func (c *syncCache) internEntry(entry *store.IndexEntry) *store.IndexEntry {
+	k := string(entry.ProvID) + cidToKey(entry.PieceID)
```
adlrocha, line 220: We may need to change this in the future. If we end up allowing providers to add any kind of metadata related to the CID, we need an alternative way of identifying the internEntry key, as a pieceIDs may not be provided.

willscott: we should expect to continue having strict constraints on the amount and structure of metadata, since such data has relatively high cost in its need to be replicated for content routing. We should expect large number of cids in an update all of which are allowed to point to one 'record' of the provider issuing them, and ideally metadata that is opaque to the indexing subsystem but which serializes to, say, less than 100 bytes.

store/primary/cache.go
```diff
@@ -0,0 +1,312 @@
...
+			continue
+		}
+		if ent == nil {
+			ent = c.internEntry(&entry)
+		}
+		c.current.Put(k, append(old, ent))
+		count++
+	}
+
+	if c.current.Len() > rotateSize {
```
adlrocha, line 180: Shouldn't we check this after every put instead of doing it at the end?

store/primary/primary.go
```diff
@@ -72,25 +62,18 @@ func (s *rtStorage) Put(c cid.Cid, providerID peer.ID, pieceID cid.Cid) error {
 // PutCheck stores a provider-piece entry for a CID if the entry is not already
 // stored.  New entries are added to the entries that are already there.
 // Returns true if a new entry was added to the cache.
+//
+// Only rotate one cache at a time. This may leave older entries in other
```
adlrocha, line 66: After reading the code (and understanding what you mentioned in our sync) I think the benefits of this implementation outgrow the overhead of leaving some outdated values behind memory-wise. 

What is less clear to me is if by adding an additional cache for entries, and consequently having to access two/three trees to get a value (vs. one/two in the implementation) we are adding a time overhead. Did you have a chance to test this?

Worst case scenario, we can configure the use of internValues as an option (that way we can have caches with and without entry deduplication) in case we want to compare both approaches.

adlrocha (approved):
I think this looks great, @gammazero. Thanks!

My only question is related to the time overhead v.s. memory improvements compared to the previous approach (just in case you had a chance to test it).

store/primary/primary.go
gammazero: The overhead to having additional trees, as compared to additional branches in the same tree, is insignificant.  I had in fact looked at this and considered prefixing different types of data with a different byte to make separate branches, but that did not save anything.

store/primary/cache.go
gammazero: I will change the implementation in a follow-up PR

store/primary/cache.go
gammazero: Yes... since there is no telling how many they might be adding.

gammazero: Fixed

PR #10, opened 2021-07-09
Implement Remove functions for Storethehash storage

adlrocha:
Implements https://github.com/filecoin-project/storetheindex/issues/7
Depends on: https://github.com/ipld/go-storethehash/pull/6

PR #12, opened 2021-07-12
Pogreb-backed persistence

adlrocha:
Depends on: https://github.com/filecoin-project/storetheindex/pull/10

This PR: 
- Implements `StorageFlusher` interface for Pogreb.
- Organizes the code so persistent storage implementation can reuse core unit tests and benchmarks code.

__Sidenote regarding storetheindex storage interface: Is true that pogreb and storethehash are quite alike, but implementing a new k-v storage for our interface has been quite straightforward and no changes have been needed in the interface. It may be worth doing this same exercise with a full-fledge database like PostgreSQL to further validate this.

Also, worth checking this out for alternatives to Pogreb: https://github.com/akrylysov/pogreb/issues/38#issuecomment-850852472__

//cc @gammazero

### DIFF
--- a/store/persistent/benchmarks.go
+++ b/store/persistent/benchmarks.go
@@ -16,41 +16,6 @@ import (
 const testDataDir = "../../test_data/"
 const testDataExt = ".data"
 
-/*
-func BenchmarkSingle10KB(b *testing.B) {
-	benchSingleGet("10KB", b)
-}
-
-func BenchmarkSingle10MB(b *testing.B) {
-	benchSingleGet("10MB", b)
-}
-
-func BenchmarkSingle100MB(b *testing.B) {
-	benchSingleGet("100MB", b)
-}
-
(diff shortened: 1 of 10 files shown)


store/persistent/benchmarks.go
```diff
@@ -0,0 +1,134 @@
...
+	for c := range out {
+		now := time.Now()
+		_, found, err := s.Get(c)
+		if err != nil || !found {
+			b.Errorf("cid not found")
+		}
+		m.getTime.add(time.Since(now).Microseconds())
+
+	}
+	err = <-errOut
```
gammazero, line 69: I really like this pattern, collect results in range loop and only check error after channel is closed.
👍

gammazero (approved):
I will probably copy/reuse you benchmarks for the cache, so that we are benchmarking identically.

PR #13, opened 2021-07-13
API takes IndexEntry instead of providerID + data

gammazero:
The information associated with a CID will depend on the provider protocol.  This has been changed from storing PieceID to storing a byte slice that starts with a uvarint  to encode a protocol number that will indicate how to decode the following bytes.

- IndexEntry does not have PieceID, but now has uvarint + data
- Add Memory use tests for primary (run with `TEST_MEM_USE=1 go test -v`)
- Rename IndexEntry.ProvID -> IndexEntry.ProviderID

### DIFF
--- a/store/primary/cache.go
+++ b/store/primary/cache.go
@@ -20,15 +20,15 @@ type syncCache struct {
 	prevEnts *radixtree.Bytes
 
 	mutex     sync.Mutex
-	rotations int
+	rotations uint64
 }
 
 type CacheStats struct {
-	Cids           int
-	Values         int
-	UniqueValues   int
-	InternedValues int
-	Rotations      int
+	Cids           uint64
+	Values         uint64
+	UniqueValues   uint64
+	InternedValues uint64
+	Rotations      uint64
 }
 
 // newSyncCache created a new syncCache instance
@@ -50,36 +50,36 @@ func newSyncCache() *syncCache {
 // items have been removed from cache.
 func (c *syncCache) stats() CacheStats {
 	unique := make(map[*store.IndexEntry]struct{})
-	var cidCount, valCount, interned int
+	var cidCount, valCount, interned uint64
 
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
 	walkFunc := func(k string, v interface{}) bool {
 		values := v.([]*store.IndexEntry)
-		valCount += len(values)
+		valCount += uint64(len(values))
 		for _, val := range values {
 			unique[val] = struct{}{}
 		}
 		return false
 	}
 
-	cidCount = c.current.Len()
+	cidCount = uint64(c.current.Len())
 	c.current.Walk("", walkFunc)
 	if c.previous != nil {
-		cidCount += c.previous.Len()
+		cidCount += uint64(c.previous.Len())
 		c.previous.Walk("", walkFunc)
 	}
 
-	interned = c.curEnts.Len()
+	interned = uint64(c.curEnts.Len())
 	if c.prevEnts != nil {
-		interned += c.prevEnts.Len()
+		interned += uint64(c.prevEnts.Len())
 	}
 
 	return CacheStats{
 		Cids:           cidCount,
 		Values:         valCount,
-		UniqueValues:   len(unique),
+		UniqueValues:   uint64(len(unique)),
 		InternedValues: interned,
 		Rotations:      c.rotations,
 	}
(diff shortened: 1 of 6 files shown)


store/primary/cache.go
```diff
@@ -218,7 +218,7 @@ func (c *syncCache) rotate() {
 }
 
 func (c *syncCache) internEntry(entry *store.IndexEntry) *store.IndexEntry {
-	k := string(entry.ProvID) + cidToKey(entry.PieceID)
+	k := string(entry.ProviderID) + string(entry.Metadata)
```
adlrocha, line 221: Quick idea (maybe it makes no sense at this point, feel free to disregard it): what about using the first e.g. 20 bytes of Metadata for the key? Just as a sanity-check in case we start storing a lot of metadata per CID in the system, to prevent from ending up with arbitrarily large keys here.

store/store.go
```diff
@@ -1,16 +1,57 @@
...
+}
+
+// PutData writes the protocol ID and the encoded data to IndexEntry.Metadata
+func (ie *IndexEntry) PutData(protocol uint64, data []byte) {
+	ie.Metadata = encodeMetadata(protocol, data)
+}
+
+// GetData returns the protocol ID and the encoded data from the IndexEntry.Metadata
+func (ie *IndexEntry) GetData() (uint64, []byte, error) {
+	protocol, len, err := varint.FromUvarint(ie.Metadata)
```
adlrocha, line 35: Quick question to see if I understood this right. The protocolID determines how the data has been encoded, right? How do you envision the flow of integrating new protocols? Who determines what protocolID to use, the provider or the underlying transport protocol used to attest data to the index?

store/store.go
```diff
@@ -22,13 +63,13 @@ type Storage interface {
...
+	Put(c cid.Cid, entry IndexEntry) error
 	// PutMany stores the provider-piece entry for multiple CIDs
-	PutMany(cs []cid.Cid, providerID peer.ID, pieceID cid.Cid) error
+	PutMany(cs []cid.Cid, entry IndexEntry) error
 	// Remove removes a provider-piece entry for a CID
-	Remove(c cid.Cid, providerID peer.ID, pieceID cid.Cid) error
+	Remove(c cid.Cid, entry IndexEntry) error
 	// RemoveMany removes a provider-piece entry from multiple CIDs
-	RemoveMany(cids []cid.Cid, providerID peer.ID, pieceID cid.Cid) error
+	RemoveMany(cids []cid.Cid, entry IndexEntry) error
```
adlrocha, line 72: *Not a blocker by any means, I can address it myself in my next PR, but wanted to open the discussion*

While working in the pass-through for the cache+persistence integration I realized that it may be convenient to add a `PutEntries/Move` (choose your preferred name) function for the cache so we can bulk put/move an `[]IndexEntry` from persistent to cache without having to loop through the entries putting them individually in the cache (which can be quite an overhead).

My suggestion would be to add:
```
type CacheStorage interface{
    Storage
     PutEntries(c cid.Cid, entries []IndexEntry) error
}
```
That way we have `StorageFlusher` (we maybe should consider changing the name) for the persistence layer, and `CacheStorage` for the cache.

(Although, I am happy to add `PutEntries` directly in `Storage` and implement it in all storages as it may also come handy for the persistence layer).

adlrocha (approved):

adlrocha:
Looks great, thanks @gammazero. Just a few follow-up questions if you don't mind.

store/primary/cache.go
willscott: hashing the indexentry to get its cid as a key would be preferable to a prefix, but this seems fine for now

gammazero: If we do not use the entirety of the metadata, and the metadata is opaque, then something with only slightly different metadata (in last few bytes) might not be seen as different when it needs to be.  If keys are getting too big, then instead of using the metadata, we can use a hash of the metadata.

If keys are similar, then any common prefixes in the data get compressed out by the radix tree, so depending on what the data looks like, long keys could take up very little space.

store/store.go
gammazero: I think the provider will choose this, but I do not think anyone has decided exactly this would work.  I am thinking that if a provider has certain information that helps them look up data for a CID, (deal id, region, storage data, special magic number), then they will register a protocol number and some specification of what data accompanies a retrieve CID request.

store/store.go
gammazero: Since you are already benchmarking this, it probably makes more sense for you to implement this so that you can tune it for maximum awesomeness.

PR #15, opened 2021-07-13
Finish propagating interface changes through code

gammazero:
This is a follow-up PR to finish propagating the interface changes, that use IndexEntry instead of provider and piece ID.


adlrocha (approved):
Wow! That was super fast. Thanks @gammazero. LGTM.
I will rebase tomorrow and push the passthrough PR (I still need to add some more tests).

PR #17, opened 2021-07-14
Implemented node storage. Fixed synthetic. Interface changes

adlrocha:
- Add `nodeStorage`, which wraps the combination of a cache and a persistent storage and implements the `Storage` interface. It includes the pass-through logic, and is the backend storage for `node`. It has been decoupled from the daemon so this logic can be used as a library instead of embedding it in the node's daemon. 
- Add a slight modification to `Put` and `Remove` in `Storage` (changing output from `err` to `(bool, err)`) for convenience.
- Renamed `StorageFlusher` to `PersistentStorage` in case we end up adding (as briefly discussed) a `CacheStorage` interface.
- Refactored `synthetic`commands.
- Fixed issue with flags, and storage config in daemon.
- Update README.
- Additional benchmarks for persistent storage.

Next: 
- Add stats for `nodeStorage` (cache hits/miss, persistence miss, cid distribution, etc.)
- Improve `nodeStorage`'s `get` function adding a `PutEntries` function to `primary so we can move a list of entries efficiently between persistent storage and cache without having to loop through entries (is this worth?)

### DIFF
--- a/node/storage.go
+++ b/node/storage.go
@@ -26,17 +26,28 @@ func (ns *nodeStorage) Get(c cid.Cid) ([]store.IndexEntry, bool, error) {
 	if ns.primary != nil {
 		// Check if CID in primary storage
 		v, found, err := ns.primary.Get(c)
+		if err != nil {
+			return nil, false, err
+		}
 
-		if ns.persistent != nil && (!found || err != nil) {
+		if !found && ns.persistent != nil {
 			v, found, err = ns.persistent.Get(c)
+			if err != nil {
+				return nil, false, err
+			}
 			// TODO: What about adding a CacheStorage interface that includes
 			// putEntries(cid, []IndexEntry) function
 			// so we don't need to loop through IndexEntry to move from
 			// one storage to another?
-			if found && err == nil {
+			if found {
 				// Move from persistent to cache
 				for i := range v {
-					ns.primary.Put(c, v[i])
+					_, err := ns.primary.Put(c, v[i])
+					if err != nil {
+						// Only log error since request has been satisified
+						log.Errorw("failed to put entry into primary storage", "err", err)
+						break
+					}
 				}
 			}
 		}
@@ -54,29 +65,36 @@ func (ns *nodeStorage) Put(c cid.Cid, entry store.IndexEntry) (bool, error) {
 }
 
 func (ns *nodeStorage) put(c cid.Cid, entry store.IndexEntry) (bool, error) {
-	var ok bool
-	var err error
-	// If there is persistent storage we put in persistent storage
-	if ns.persistent != nil {
-		ok, err = ns.persistent.Put(c, entry)
-		if err != nil {
-			return false, err
-		}
-	} else {
-		// If not we put in primary storage right away
+	// If there's only primary storage put right away
+	if ns.persistent == nil {
 		return ns.primary.Put(c, entry)
 	}
 
-	// If we have both, we need to check if there is an entry for
-	// the cid in primary and update it accordingly to keep both
-	// storages consistent
+	// If both check first if cid already in cache
 	if ns.primary != nil {
-		_, found, _ := ns.primary.Get(c)
-		if found && ok {
-			return ns.primary.Put(c, entry)
+		v, found, err := ns.primary.Get(c)
+		if err != nil {
+			return false, err
+		}
+		// If found, check if entry already exists in cache as if this
+		// is the case there's no need to put anything new.
+		if found {
+			for i := range v {
+				// If exists, no put needed here nor in persistent
+				if v[i].Equal(entry) {
+					return false, nil
+				}
+			}
+			// Put in primary
+			_, err := ns.primary.Put(c, entry)
+			if err != nil {
+				log.Errorw("failed to put entry into primary storage", "err", err)
+			}
 		}
 	}
-	return ok, err
+
+	// If persistent and not in cache we always put
+	return ns.persistent.Put(c, entry)
 }
 
 // PutMany stores entry for multiple CIDs
--- a/node/storage.go
+++ b/node/storage.go
@@ -1,0 +1,188 @@
+package node
+
+import (
+	"fmt"
+
+	"github.com/filecoin-project/storetheindex/store"
+	"github.com/ipfs/go-cid"
+	peer "github.com/libp2p/go-libp2p-core/peer"
+)
+
+var _ store.Storage = &nodeStorage{}
+
+// nodeStorage combines a cache and a persistent storage
+type nodeStorage struct {
+	primary    store.Storage
+	persistent store.PersistentStorage
+}
+
+// NewStorage creates new nodeStorage from a primary and persistent storages
+func NewStorage(primary store.Storage, persistent store.PersistentStorage) *nodeStorage {
+	return &nodeStorage{primary, persistent}
+}
+
+// Get retrieves IndexEntries for a CID
+func (ns *nodeStorage) Get(c cid.Cid) ([]store.IndexEntry, bool, error) {
+	v := []store.IndexEntry{}
+	var found bool
+	var err error
+
+	if ns.primary != nil {
+		// Check if CID in primary storage
+		v, found, err = ns.primary.Get(c)
+
+		if ns.persistent != nil && (!found || err != nil) {
+			v, found, err = ns.persistent.Get(c)
+			// TODO: What about adding a CacheStorage interface that includes
+			// putEntries(cid, []IndexEntry) function
+			// so we don't need to loop through IndexEntry to move from
+			// one storage to another?
+			if found && err == nil {
+				// Move from persistent to cache
+				for i := range v {
+					ns.primary.Put(c, v[i])
+				}
+			}
+		}
+		return v, found, err
+	}
+
+	// If no primary storage, get from persistent
+	return ns.persistent.Get(c)
+}
+
+// Put stores entry for a CID if the entry is not already
+// stored.  New entries are added to the entries that are already there.
+func (ns *nodeStorage) Put(c cid.Cid, entry store.IndexEntry) (bool, error) {
+	return ns.put(c, entry)
+}
+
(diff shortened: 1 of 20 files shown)


store/persistent/benchmarks.go
```diff
@@ -105,6 +107,76 @@ func BenchParallelGet(s store.StorageFlusher, size string, b *testing.B) {
...
+	cids, _ = utils.RandomCids(4096)
+	s.PutMany(cids, entry)
+
+	for testCount := 1024; testCount < 10240; testCount *= 2 {
+		b.Run(fmt.Sprint("Get", testCount), func(b *testing.B) {
+			b.ReportAllocs()
+			b.ResetTimer()
+			m := initMetrics()
+			for i := 0; i < b.N; i++ {
+				for j := 0; j < testCount; j++ {
```
gammazero, line 128: These two loops are the equivalent of:
```go
for i := 0; i < b.N * testCount; i++ {
    ...
}
```
Makes more sense to let `b.N` alone vary the number of queries.  That way the benchmark can vary `b.N` to see how long it takes to do a query on average.

store/persistent/benchmarks.go
```diff
@@ -105,6 +107,76 @@ func BenchParallelGet(s store.StorageFlusher, size string, b *testing.B) {
...
+		panic(err)
+	}
+	p, _ := peer.Decode("12D3KooWKRyzVWW6ChFjQjK4miCty85Niy48tpPV95XdKu1BcvMA")
+
+	entry := store.MakeIndexEntry(p, protocolID, cids[0].Bytes())
+
+	cids, _ = utils.RandomCids(4096)
+	s.PutMany(cids, entry)
+
+	for testCount := 1024; testCount < 10240; testCount *= 2 {
```
gammazero, line 122: Each test run is changing the number of queries done, but so is `b.N`.  I see that the intent is to benchmark how long it takes to do some number of queries, and each test run changes the number of queries.  However, effectively what is being done is to make the benchmark less sensitive / less accurate with each run, since the same values of `b.N` are not used by different runs.  The benchmarker changes `b.N` only as much as it needs to in order to get to a stable benchmark.

If you use `b.N` as the limit in the outer loop, then you can get a benchmark of how long it takes to do some fixed number of queries.

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+func NewStorage(primary store.Storage, persistent store.PersistentStorage) *nodeStorage {
+	return &nodeStorage{primary, persistent}
+}
+
+// Get retrieves IndexEntries for a CID
+func (ns *nodeStorage) Get(c cid.Cid) ([]store.IndexEntry, bool, error) {
+	if ns.primary != nil {
+		// Check if CID in primary storage
+		v, found, err := ns.primary.Get(c)
+
```
gammazero, line 29: Need to check `err` first.  I do not think we want to hide the error just because the value might be retrievable from persistent storage, because an error at this point means something is broken.
```suggestion
if err != nil {
	return nil, false, err
}
```

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+		if ns.persistent != nil && (!found || err != nil) {
```
gammazero, line 30:
```suggestion
		if !found && ns.persistent != nil {
```

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+			v, found, err = ns.persistent.Get(c)
+			// TODO: What about adding a CacheStorage interface that includes
```
gammazero, line 32: As above, need to check `err` first:
```go
if err != nil {
	return nil, false, err
}
```

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+			// putEntries(cid, []IndexEntry) function
+			// so we don't need to loop through IndexEntry to move from
+			// one storage to another?
+			if found && err == nil {
```
gammazero, line 36:
```suggestion
			if found && ns.primary != nil {
```
Need to check that there is a primary storage.

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+				// Move from persistent to cache
+				for i := range v {
+					ns.primary.Put(c, v[i])
```
gammazero, line 39: Need to check `err` here, but only log it here:
```go
err := ns.primary.Put(c, v[i])
if err != nil {
    // Only log error since request has been satisified
    log.Errorw("failed to put entry into primary storage", "err", err)
    break
}
```

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+	// If there is persistent storage we put in persistent storage
+	if ns.persistent != nil {
+		ok, err = ns.persistent.Put(c, entry)
+		if err != nil {
+			return false, err
+		}
+	} else {
+		// If not we put in primary storage right away
+		return ns.primary.Put(c, entry)
+	}
```
gammazero, line 68: Minor:  Code would look a little cleaner if the check is `if ns.persistent == nil`, because then the `else` goes away and the flow is visually easier to follow.

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+// Put stores entry for a CID if the entry is not already
+// stored.  New entries are added to the entries that are already there.
+func (ns *nodeStorage) Put(c cid.Cid, entry store.IndexEntry) (bool, error) {
+	return ns.put(c, entry)
+}
+
+func (ns *nodeStorage) put(c cid.Cid, entry store.IndexEntry) (bool, error) {
+	var ok bool
+	var err error
+	// If there is persistent storage we put in persistent storage
```
gammazero, line 59: Should the primary be checked first?  If there is a guarantee that anything that is in the primary is already in persistent storage, then that would avoid unneeded updates without having to query persistent storage, for values already cached.

Doing this would probably require a new API, `HasValue(cid.Cid, store.IndexEntry) bool` or would mean looking at the value(s) returned by `Get`.  Is it worth it?

node/storage.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// Put stores entry for a CID if the entry is not already
+// stored.  New entries are added to the entries that are already there.
+func (ns *nodeStorage) Put(c cid.Cid, entry store.IndexEntry) (bool, error) {
+	return ns.put(c, entry)
+}
+
+func (ns *nodeStorage) put(c cid.Cid, entry store.IndexEntry) (bool, error) {
+	var ok bool
```
gammazero, line 57: maybe change `ok` to `isNew` for clarity on what it means.  it makes sense since given how this variable is used at the end of the function.

gammazero (changes requested):
There are some changes that are needed:

- Some changes to error handling and other code changed are recommended.
- A couple of the benchmarks are broken.  If it is not clear why from the comments LMK.

node/storage.go
adlrocha: This is already checked at the [beginning of this block](https://github.com/filecoin-project/storetheindex/pull/17/files#diff-059931e6619ad69718bd9674f123a7445e26ef0badd0a68118ff23bdc5b48b7fR26), right? We can't reach this code if there is no primary storage. In any case, the `err` check is redundant.

node/storage.go
adlrocha: Good point. My reasoning was that maybe we want the indexer to still be able to give service even if the primary storage breaks, but maybe this is too strong of an assumption. And we can always change it in the future.

node/storage.go
adlrocha: You are right. This is a better approach. Fixed. Thanks!

store/persistent/benchmarks.go
```diff
@@ -105,6 +107,76 @@ func BenchParallelGet(s store.StorageFlusher, size string, b *testing.B) {
...
+				go func(wg *sync.WaitGroup) {
+					m := initMetrics()
+					for i := 0; i < b.N; i++ {
+						for j := 0; j < testCount; j++ {
+							now := time.Now()
+							_, ok, _ := s.Get(cids[j%len(cids)])
+							if !ok {
+								panic("missing cid")
+							}
+							m.getTime.add(time.Since(now).Microseconds())
```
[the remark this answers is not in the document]
adlrocha: Indeed, it was slightly modifying the results reported by the benchmark, but as I was disregarding the benchmark results all along and using my average measurement, the final outcome was the same 😅

adlrocha:
@gammazero, thank you for the thorough review (specially with benchmarks, I was mixing/overthinking/abusing them. I choose to simplify them so others can replicate them more easily following a similar approach to the one you used for `primary`).

Let me know if there's anything else that needs to be addressed before merge. Thanks!

store/persistent/benchmarks.go
```diff
@@ -37,72 +42,146 @@ func prepare(s store.Storage, size string, b *testing.B) {
...
+		}
+	})
+
+	// Test time to fetch certain amount of requests
+	for testCount := 1024; testCount < 10240; testCount *= 2 {
+		b.Run(fmt.Sprint("Get", testCount), func(b *testing.B) {
+			b.ReportAllocs()
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				for j := 0; j < testCount; j++ {
```
gammazero, line 132: It looks like these two loops are the same as doing:
```go
for i := 0; i < b.N * testCount; i++ {
    _, ok, _ := s.Get(cids[i%len(cids)])
}
```
We can revisit the benchmarks later, but let's get these in for now if they are generating meaningful results.

gammazero (approved):

gammazero:
@adlrocha Note that some of the unit tests are failing in CI: `... storethehash.data: The process cannot access the file because it is being used by another process.`

This can probably be fixed by waiting for shutdown to finish at the end of the test.

adlrocha:
> @adlrocha Note that some of the unit tests are failing in CI: `... storethehash.data: The process cannot access the file because it is being used by another process.`
> 
> This can probably be fixed by waiting for shutdown to finish at the end of the test.

Fixed the Windows CI error. For future reference, apparently there's a bug in Windows with `testing.TempDir` that makes it fail when cleaning up the test. Changing it to `ioutil.TempDir` made all test pass.

PR #20, opened 2021-07-19
Assets to run load tests using locust

adlrocha:
- Include assets to run load tests against a `storetheindex` daemon using https://locust.io/
- Add README to guide through how to replicate the tests

### DIFF
--- a/utils/load/load.py
+++ b/utils/load/load.py
@@ -1,0 +1,46 @@
+"""
+This locust test randomly gets CIDs from a
+storetheindex daemon. CIDs are chosen randomly following a 
+ZIPF distribution. 
+"""
+
+import numpy as np
+
+import time
+from locust import HttpUser, task, between
+import random
+
+# Initializing Zipf distribution
+a = 2.  # Zipf parameter (feel free to change it)
+
+# Read from a list of imported cids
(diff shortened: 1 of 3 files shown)


utils/load/load.py
```diff
@@ -0,0 +1,46 @@
...
+import time
+from locust import HttpUser, task, between
+import random
+
+# Initializing Zipf distribution
+a = 2.  # Zipf parameter (feel free to change it)
+
+# Read from a list of imported cids
+def read_cids(filepath):
+    file1 = open(filepath, 'r')
```
gammazero, line 18: You need to close the file.
```suggestion
    with open(filepath) as file1:
```

Also, the 'r' is not needed as that is the default for opening.

utils/load/load.py
```diff
@@ -0,0 +1,46 @@
...
+    lines = file1.readlines()
+    cids = []
+    for line in lines:
+        cids.append(line.strip())
+    return cids
```
gammazero, line 23: minor:  Just use a list comprehension for all code starting at `cids = []`:
```python
return [line.strip() for line in lines]
```

gammazero (approved):
Couple of comments to see before merge.

PR #21, opened 2021-07-19
Fix storethehash periodic flush

adlrocha:
- Fixes a bug where `storethehash` persistence wasn't starting its sync routine and was not periodically flushing from memory.
- Adds unit test.

### DIFF
--- a/store/persistent/storethehash/storethehash.go
+++ b/store/persistent/storethehash/storethehash.go
@@ -216,7 +216,6 @@ func (s *sthStorage) removeEntry(k []byte, entry store.IndexEntry, stored []stor
 
 // Close stops all storage-related routines, and flushes
 // pending data
-// NOTE: Add it to storage interface?
 func (s *sthStorage) Close() error {
 	return s.store.Close()
 }
--- a/store/persistent/storethehash/storethehash.go
+++ b/store/persistent/storethehash/storethehash.go
@@ -42,6 +42,7 @@ func New(dir string) (*sthStorage, error) {
 	if err != nil {
 		return nil, err
 	}
+	s.Start()
 	return &sthStorage{dir: dir, store: s}, nil
 }
 
@@ -211,4 +212,11 @@ func (s *sthStorage) removeEntry(k []byte, entry store.IndexEntry, stored []stor
 		}
 	}
 	return false, nil
+}
+
+// Close stops all storage-related routines, and flushes
+// pending data
+// NOTE: Add it to storage interface?
+func (s *sthStorage) Close() error {
+	return s.store.Close()
 }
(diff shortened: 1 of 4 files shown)


store/persistent/storethehash/storethehash.go
```diff
@@ -212,3 +213,10 @@ func (s *sthStorage) removeEntry(k []byte, entry store.IndexEntry, stored []stor
 	}
 	return false, nil
 }
+
+// Close stops all storage-related routines, and flushes
+// pending data
+// NOTE: Add it to storage interface?
+func (s *sthStorage) Close() error {
+	return s.store.Close()
```
gammazero, line 221: It probably makes sense to expose this through the Storage interface to make sure data is flushed at shutdown.  Would there be any utility in exposing a `Flush` method so that a library user could call that after finishing some bulk ingestion?

store/persistent/storethehash/storethehash_test.go
```diff
@@ -40,3 +44,60 @@ func TestRemoveMany(t *testing.T) {
...
+	for _, c := range cids[1:] {
+		_, err = s.Put(c, entry)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// Sleep for 2 sync Intervals to ensure that data is flushed
+	time.Sleep(2 * storethehash.DefaultSyncInterval)
+	s.Close()
```
gammazero, line 83: I think that calling `Close` should cause a flush immediately and that there should not be any need to wait before closing.  If there is a need to wait before closing then that should be built into the `Close`

gammazero (changes requested):
Requesting change, because there must not be a need to wait before calling `Close` to make sure data is synced.  Calling `Close` should force this immediately, or is not possible then should have the wait built in.

store/persistent/storethehash/storethehash_test.go
adlrocha: This `s.Close()` is not necessary here. I forgot to remove it from the test after doing some sanity-checks. `s.Close()` calls storethehash's `Close` method which is the one [responsible for immediately flushing](https://github.com/ipld/go-storethehash/blob/1ea50cc878b8f7f10f79207a8060d950edb8a0e8/store/store.go#L106). Already fixed it. Thanks!

store/persistent/storethehash/storethehash.go
adlrocha: I exposed `Close` in the persistent storage interface. I haven't removed `Flush` because it is being used by unit tests to avoid having to wait for the next sync tick. I checked pogreb's API and it exposes both, Close and Flush, but let me know if you think it'd be better to remove Flush from the interface.

adlrocha:
> Requesting change, because it must not be necessary to wait before calling Close to make sure data is synced. Calling Close should force this immediately, or if not possible, then Close should have the wait built-in.

It is not necessary to wait in order to flush data before Close. The storethehash `Close` method is the one responsible for flushing if there is outstanding data in memory. Let me know if there is any additional comment to be addressed. Thanks!

gammazero (approved):

PR #26, opened 2021-08-02
Add client interface with libp2p and http implementations

adlrocha:
The PR includes:
- Client interface with `Get` and `GetBatch` functions.
- Models for request and response protocol messages.
- Implementation of a reference `client` and `server` for the client querying protocol over `http` and `libp2p`
- Integration of existing `import` and `admin` endpoints over the `http` server.
- Integration of server implementations of client protocol over node process, and of the client reference implementations in the CLI.

We can now start an indexer node listening to end user requests over http and libp2p.

_Future work:_
-  _No authentication has been included yet_ 
- _Once we are comfortable with the protocol, we should add specs to allow other client implementations_


willscott:
should we separate out the import and admin functionality that are logically separate - import for the provider side interface rather than 'client', and 'admin' probably running on a separate endpoint that can be protected.

adlrocha:
> should we separate out the import and admin functionality that are logically separate - import for the provider side interface rather than 'client', and 'admin' probably running on a separate endpoint that can be protected.

Sure, that makes sense. Actually I will add `import` endpoints to the admin API for now, but once the actual ingestion protocol is in place we may not need them anymore. They are just handy for debugging/testing.

client/libp2p/client/message_sender.go
```diff
@@ -0,0 +1,229 @@
...
+var readMessageTimeout = 10 * time.Second
+
+// ErrReadTimeout is an error that occurs when no message is read within the timeout period.
+var ErrReadTimeout = fmt.Errorf("timed out reading response")
+
+// messageSenderImpl is responsible for sending requests and messages to peers
+type messageSenderImpl struct {
+	host      host.Host // the network services we need
+	smlk      sync.Mutex
+	strmap    map[peer.ID]*peerMessageSender
```
gammazero, line 28: A more descriptive name would be good here.

client/libp2p/client/message_sender.go
```diff
@@ -0,0 +1,229 @@
...
+	go func(r msgio.ReadCloser) {
+		defer close(errc)
+		bytes, err := r.ReadMsg()
+		defer r.ReleaseMsg(bytes)
+		if err != nil {
+			errc <- err
+			return
+		}
+		errc <- mes.Unmarshal(bytes)
+	}(ms.r)
```
gammazero, line 216: Instead of starting a new goroutine for every read, consider having a read goroutine that is started when the message sender is created.  The read goroutine would wait on a channel that takes a request object.  The request object contains a `*pb.Message` and an error chan to respond on.

commands/find.go
```diff
@@ -33,6 +69,9 @@ func getCidCmd(c *cli.Context) error {
 		return err
 	}
 
-	return cl.Get(ctx, ccid)
+	resp, err := cl.Get(ctx, ccid, end)
+	log.Info("Response:")
```
gammazero, line 73: Log something more informative?

commands/find.go
```diff
@@ -33,6 +69,9 @@ func getCidCmd(c *cli.Context) error {
...
+	resp.PrettyPrint()
```
gammazero, line 74: Pretty-printed output should be an option, not the default IMO.  If this is JSON, then people can just run it through `jq` anyway to make it pretty.

gammazero (commented):
Review is not complete, but publishing a few comments so far.

adlrocha:
@gammazero, I think we have a first iteration of the refactor with what we discussed in our sync. All tests pass and I manually checked the commands for which we don't have unit tests yet. 

Let me know if there is anything else you want to get into this PR, if not let's keep this as a baseline and iterate in future PRs (this one is getting gigantic :) ). Thanks!

gammazero:
I think that there should be either a `client/` or an `api/` directory but not both, because these do not seem like they can be separated.  Can I have a v1 and a v2 protocol with the same client?  If these cannot be separated, then maybe decide which name you want to keep and move everything into that directory.

If these can be separated, then this is fine and maybe I just did not understand how that would work.  Let me know what you think so we can get this merged soon.

adlrocha:
Added `client` to `api/v1`. I think at this stage we can consider clients version-specific. Thanks!

gammazero (approved):

PR #27, opened 2021-08-03
Add config data structure that can be saved and loaded

gammazero:
- Add command to create initial configuration file with defaults


config/discovery.go
```diff
@@ -0,0 +1,9 @@
+package config
+
+// Discoveryu stores the pubsub router address to subscribe to, and a list of
+// allowed discovery clients
+type Discovery struct {
+	PubsubRouter      string
+	AllowClients      []string
```
adlrocha, line 7: I guess this is a placeholder and will change in the future, but a few questions:
* With `PubsubRouter` you mean the PubSub topic the storetheindex node will be listening for discovery/announcements? We don't expect indexers to have to listen to more than one topic, right?
* `AllowClients` is referring to providers? Is the list of whitelisted providers the indexer is giving service to. If this is the case what about using `AllowProviders`?
* And maybe this will make sense discussing it when we start building this part of the system, but what piece of information were you considering to identify entities in the AllowList (IPs, multiaddr, peer.ID)?

Thanks!

config/init.go
```diff
@@ -0,0 +1,83 @@
...
+	conf := &Config{
+		// setup the node's default addresses.
+		Addresses: Addresses{
+			Admin:     "/ip4/0.0.0.0/tcp/3001",
+			Discovery: "/ip4/0.0.0.0/tcp/3002",
+			Ingest:    "/ip4/0.0.0.0/tcp/3003",
+			Finder:    "/ip4/0.0.0.0/tcp/3000",
+		},
+
+		Datastore: Datastore{
```
adlrocha, line 32: I think I don't understand the datastore config (I may be missing something). Are we planning to support different types and according to the type a different datastore is initialized?

adlrocha (approved):
LGTM. I just added a few non-blocking questions. Thanks!

config/init.go
gammazero: I was thinking that we would support different datastores, but really there is no need to.  I should probably remove his part of the config.

config/discovery.go
gammazero: - The `PubsubRouter` is a placeholder, and I was thinking there would be some gossip address to connect to and subscribe.
- `AllowClients` is a list of client IDs to allow discovery clients to connect to the TBD discovery API and tell the indexer about new providers.  For example, the storage broker could be one such client and tell the indexer about providers it knows about.  Maybe this does not make sense to have a placeholder for yet. How about if I just have a filename to read to discover providers? 
- The allow/block list is not yet part of the config, because I had not decided whether or not it was indexer config, or some other file.  I think I will make this part of the config, but put it in a separate section called "Providers" or "Ingest".

WDYT about:
```json
"Providers": {
  "DefaultPolicy": "block",
  "Except": [provider1_peerID, provider2_peerID],
}
```

And the discovery mechanism is to drop a file into `.storetheindex/discovery/` where the file is a list of `peer.AddrInfo`?

willscott: for pubsub we probably have a string topic as a config, and also will want a list of default bootstrap peers we should connect to (e.g. ipfs gateway bootstraps) at startup, as a place to then discover other pubsub peers and mesh.

The rest of the plan here sounds good

adlrocha: Now I get it. That all sounds good.

PR #28, opened 2021-08-03
Service should not know about cli commands

gammazero:


command/daemon.go
```diff
@@ -21,36 +21,47 @@ var DaemonCmd = &cli.Command{
 	Action: daemonCommand,
 }
 
-func daemonCommand(c *cli.Context) error {
-	ctx, cancel := context.WithCancel(ProcessContext())
-	defer cancel()
+func daemonCommand(cctx *cli.Context) error {
+	// TODO: get these from config file
```
adlrocha, line 25: Does this mean that we'll remove CLI flags or that we are going to make them optional as a way to overwrite the current config?

gammazero: We will keep them to override configured values where it makes sense.

command/daemon.go
```diff
@@ -21,36 +21,47 @@ var DaemonCmd = &cli.Command{
...
 
-	go func() {
-		// Wait for SIGINT (CTRL-c), then close server and exit.
-		sigint := make(chan os.Signal, 1)
-		signal.Notify(sigint, os.Interrupt)
-		<-sigint
+	n.Start()
+
+	// Wait for SIGINT (CTRL-c), then close server and exit.
+	sigint := make(chan os.Signal, 1)
```
adlrocha, line 42: This approach makes way more sense than what I was doing. I can't remember why I included this in the routine in the first place 🤦‍♂️

internal/node/node.go
```diff
@@ -32,88 +31,96 @@ type Node struct {
...
 	if p2pEnabled {
+		ctx, cancel := context.WithCancel(context.Background())
+
 		// NOTE: We are creating a new flat libp2p host here because no other
 		// process in the indexer node needs a libp2p host. In the future, when
 		// the indexer node starts using other libp2p protocols to interact with
 		// miners and other indexers, we may need to initialize it before and
 		// use it here so we have a single libp2p host giving service to the whole indexer.
-		h, err := libp2p.New(ctx)
+		p2pHost, err := libp2p.New(ctx)
```
adlrocha, line 90: Just a reminder. We should create the libp2p host using the keys generated in the config. Let's maybe include it already in this PR so we don't forget in the future.

Also, this libp2p host will be shared by the ingestion and discovery modules of the indexer, so it won't be exclusively initialized with `p2pEnabled` (this is not the case for now).

gammazero: Yes, that willl be in the next PR that combines the config PR and this one.

adlrocha (approved):
LGTM. I like the new code structure after the reorg.

PR #30, opened 2021-08-05
Change API version to v0 since it will be changing

gammazero:
- Add skeletal provider registry
- Updated find response population
- numberous misc changes


api/v0/finder/models/models.go
```diff
@@ -87,37 +88,49 @@ func UnmarshalResp(b []byte) (*Response, error) {
...
 	for i := range cids {
-		v, f, err := e.Get(cids[i])
+		values, found, err := engine.Get(cids[i])
 		if err != nil {
-			// If error and the request is not for a batch, continue without
-			// sending an error as if the errored CID wasn't found
-			if len(cids) < 1 {
-				return nil, err
-			}
+			return nil, fmt.Errorf("failed to query cid %q: %s", cids[i], err)
```
adlrocha, line 104: Just a note: do we expect the `Get` for a small set of CIDs of a batch to be able to fail when querying the indexer engine? If this is the case I would not return an error right away, and continue as if the CID wasn't found. I think a batch query shouldn't err just because a one CID in the batch has err. If we don't expect this to happen, I am happy with how it is now.

adlrocha: Actually, I think this is fine how it is. If a `Get` fails it is probably because something is wrong with the indexer-core and we should be able to trace it. Nevermind :)

adlrocha (approved):
LGTM

PR #33, opened 2021-08-13
Ingest server stub

gammazero:


internal/providers/registry.go
```diff
@@ -1,52 +1,318 @@
...
 }
 
+// ProviderInfo is an immutable data sturcture that holds information about a
+// provider.  A ProviderInfo instance is never modified, but rather a new one
+// is created to update its contents.  This means existing references remain
+// valid.
 type ProviderInfo struct {
-	Addresses []ma.Multiaddr
+	AddrInfo      peer.AddrInfo
+	FcAddress     string
```
willscott, line 42: better name?

internal/providers/registry.go
```diff
@@ -1,52 +1,318 @@
...
+	err := r.syncNeedDiscover(fcAddr)
+	if err != nil {
+		r.syncEndDiscover(fcAddr, nil, err, errCh)
+		return
+	}
+
+	// Mark discovery as in progress
+	r.discoTimes[fcAddr] = time.Time{}
+
+	r.pendingDisco++
```
willscott, line 206: i'm confused by the semantics of pendingDisco

internal/providers/registry.go
```diff
@@ -1,52 +1,318 @@
...
+// NewRegistry creates a new provider registry, giving it a Discovery interface
+// and provider policy configuration.
+func NewRegistry(providerCfg config.Providers, disco discovery.Discovery, dstore datastore.Datastore) (*Registry, error) {
+	// Create policy from config
+	providerPolicy, err := policy.New(providerCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	r := &Registry{
```
willscott, line 56: consider adding a worker pool for actions - it would be good to have a pool to know how many will be going on at once

internal/providers/registry.go
```diff
@@ -1,52 +1,318 @@
...
+
+		// Check if last discovery completed too recently
+		if !r.policy.CanRediscover(completed) {
+			return ErrTooSoon
+		}
+	}
+	return nil
+}
+
+func (r *Registry) discover(fcAddr string, signature, signed []byte) (*discovery.Discovered, error) {
```
willscott, line 275: the mix of `fcAddr` vs peer addresses, where it's unclear how much of this is filecoin specific, and what is generic to all providers should probably get clarified / delineated better

internal/providers/registry.go
gammazero: For limiting concurrency to some max number (workers in pool), or simply for tracking how many actions are running concurrently?

Running actions here is a substitute for mutex-locking the sections of code run as an action and allows the caller to decide whether or not to wait for the code to finish running.  So the maximum running actions are, the number of concurrent requests being processed + any discoveries in progress.  There will never be multiple discoveries for the same provider within the same `RediscoverWait` time.

internal/providers/registry.go
gammazero: That is the number of discoveries that are currently in progress. This is used to determine who many to wait for to finish when doing a graceful shutdown.

internal/providers/registry.go
gammazero: Is there a more generic name for the address that we identify a miner with to lotus (e.g., `t01000`)?  Should I call this `MinerAddress`, `OnChainID`, `DiscoveryID`?

api/v0/finder/models/models.go
```diff
@@ -54,7 +25,7 @@ type CidData struct {
 // Response used to answer client queries/requests
 type Response struct {
 	Cids      []CidData
-	Providers []ProviderData
+	Providers []peer.AddrInfo
```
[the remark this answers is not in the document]
gammazero: That is already included in each `CidData`. (I think `Cids` needs to be renamed `CidResponses`).  Each `CidData` is:
```go
type CidData struct {                                                                                                              
    Cid     cid.Cid                                                                                                                
    Entries []entry.Value                                                                                                          
}                                                                                                                                  
```
Where each `entry.Value` consists of a protocol ID and body of metadata.

config/providers.go
```diff
@@ -12,11 +16,18 @@ type Providers struct {
...
 	Trust []string
 	// PollInterval is the amount of time to wait without getting any updates
 	// from a provider, before sending a request for the latest advertisement.
 	// Values are a number ending in "s", "m", "h" for seconds. minutes, hours.
-	PollInterval string
+	PollInterval Duration
+
+	// DiscoveryTimeout is the maximum amount of time that the indexer will
+	// spend trying to verify a new provider.
+	DiscoveryTimeout Duration
```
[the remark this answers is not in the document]
gammazero: Discovery timeout is a configurable value, and when discovery does not complete within this time, the in-progress attempt is canceled and is considered a failure to discover the requested miner.  If the timeout is 0, then there is no discovery timeout and discovery will continue until failure or success, unless it hangs forever.

The config file is initialized with a value of 2 minutes unless otherwise specified.

internal/providers/registry.go
gammazero: I need to use `fcAddr` (e.g. `t01000`) to discover the miner via lotus.  From that, I can get the provider's `peer.ID` and its list of `multiaddr`.  Maybe this should be a new type of `Multiaddr` where the protocol is `filecoin`, and I name this value `DiscoveryID`?

api/v0/finder/models/models.go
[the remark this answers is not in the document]
gammazero: Renamed `Response.Cids` to `Response.CidResults`

internal/providers/registry.go
gammazero: If this were an IPFS provider to discover, then this would be a Multiaddr and would usually be the same one of the discovered addresses (which might include multiple addresses). So, I think a better name is DiscoverAddr. What is missing is where/how the discovery was done.

gammazero: Renamed `FcAddress` to `DiscoveryAddr`

internal/providers/registry.go
gammazero: Renamed `fcAddr` to `discoverAddr` and commented on the need for additional data to tell the indexer where/how to perform the discovery.

internal/providers/registry.go
gammazero: Changed to use `sync.WaitGroup` instead.  It was not strictly necessary to synchronize this since it, since incr and decr of counter was done in same goroutine, but using `sync.WaitGroup` makes the code cleaner and more obvious.

api/v0/finder/models/models.go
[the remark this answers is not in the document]
willscott: that seems weird though? how do i know which protocol in a ciddata response corresponds with a given addrInfo?
Are the lists of cids and providers 1:1?

PR #37, opened 2021-08-16
Add Dockerfile and entrypoint script

gammazero:
Alos added Makefile to make building easier

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,80 @@
+FROM golang:1.16-stretch as builder
+LABEL maintainer="Andrew Gillis <andrew.gillis@protocol.ai>"
+
+# Install deps
+RUN apt-get update && apt-get install -y \
+  ca-certificates
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+RUN apt-get update && apt-get install -y \
+  ca-certificates
+
+ENV SRC_DIR /storetheindex
+
+RUN mkdir /storetheindex
+WORKDIR /storetheindex
+
+# Download packages first so they can be cached.
+COPY go.mod go.sum /storetheindex/
```
willscott, line 14: this can just copy them to '.' since that's the default WORKDIR, right?

willscott (approved):

Dockerfile
gammazero: No.  If you have multiple sources then the destination needs to be a directory name followed by a /.

PR #38, opened 2021-08-17
Discover provider using lotus

gammazero:
This looks up a miner using the miner short address (e.g. ”t01000") and gets the miner's peerID and addresses.  If the requested peer id matches the one returned and the signature in the request is verified, the miner is consider discovered and my participate in indexing.


internal/lotus/client.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("api token: %w", err)
+	}
+
+	apiInfo := cliutil.APIInfo{
+		Addr:  ma.String(),
+		Token: token,
+	}
+
+	//return openFillNode(ctx, apiInfo)
```
adlrocha, line 59: I think there's a typo here :)

```suggestion
	//return openFullNode(ctx, apiInfo)
```

gammazero: I think I will just delete this.  I don't think the indexer will ever need more than the gateway interface provides.

PR #40, opened 2021-08-18
Use correct filecoin-ffi stub and update to latest lotus

gammazero:


willscott:
we need all of this to link against the JSON-RPC API on chain.love?

gammazero:
@willscott We need this new ffi-stub to build against the newer lotus (v1.11.1) and will also need it when v1.11.2 is out.  We will need the newest lotus version to build with a newer version of graphsync and ipld-prime.  That second part is not available yet, but when it is that will allow storetheindex to generate ipld code that does make go vet complain (and I am sure has many other fixes).

willscott:
Why do we need lotus though?
We can (and should) use go-data-transfer directly for syncing with providers. lotus depends on data-transfer, not the other way around

gammazero:
@willscott Using lotus for the gateway client API.  If this is not the correct way to do this, then lets change it.
https://github.com/filecoin-project/storetheindex/blob/main/internal/lotus/client.go#L29

willscott:
Yep. that makes sense, it's unfortunate it's such a heavy dependency to call the JSON-RPC API.

I don't know if there's a lighter way to get that interface without pulling in all of lotus / having to stub out FFI so no opposition to doing this as the asnwer for now

willscott (approved):

PR #41, opened 2021-08-19
Registry dstore path and register command

gammazero:
- Changed dstore path for providers saved.  This is in anticipation of different types of provider information being stored.
- Flush dstore after each update.
- HTTP server returns useful error message when applicable, such as "invalid signature" as opposed to only a BadRequest response code.
- Common structure for error messages
- Added `register` command to register trusted providers.  This only requires signature validation but does not need to verify provider on chain.  This is needed to register test providers and other off-chain type of providers.
- Write messages intended for interactive CLI users to stdout, instead of to log.


internal/providers/registry.go
```diff
@@ -58,12 +59,17 @@ type ProviderInfo struct {
...
+	return datastore.NewKey(path.Join(providerKeyPath, p.AddrInfo.ID.String()))
+}
+
 // NewRegistry creates a new provider registry, giving it provider policy
 // configuration, a datastore to persist provider data, and a Discovery
 // interface.
 //
 // TODO: It is probably necessary to have multiple discovery interfaces
 func NewRegistry(cfg config.Discovery, dstore datastore.Datastore, disco discovery.Discovery) (*Registry, error) {
+	cfg.Policy.Trust = append(cfg.Policy.Trust, "12D3KooWSG3JuvEjRkSxt93ADTjQxqe4ExbBwSkQ9Zyk1WfBaZJF")
```
willscott, line 72: leftover debugging?

command/register.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/filecoin-project/storetheindex/api/v0/ingest/models"
+	"github.com/filecoin-project/storetheindex/config"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/multiformats/go-multiaddr"
+	"github.com/urfave/cli/v2"
+)
+
+var RegisterCmd = &cli.Command{
+	Name:   "register",
+	Usage:  "Register trusted provider with indexer",
```
willscott, line 23: this is description, not usage

command/flags.go
```diff
@@ -113,6 +111,26 @@ var InitFlags = []cli.Flag{
...
 
+var RegisterFlags = []cli.Flag{
+	&cli.StringFlag{
+		Name:     "config",
+		Usage:    "Config file containing provider's peer ID and private key",
+		Required: true,
+	},
+	&cli.StringFlag{
+		Name:     "indexer-addr",
+		Usage:    "Indexer address:port to register at",
```
willscott, line 122: maybe call this the 'control port' of the indexer?

command/flags.go
```diff
@@ -113,6 +111,26 @@ var InitFlags = []cli.Flag{
 	},
 }
 
+var RegisterFlags = []cli.Flag{
+	&cli.StringFlag{
+		Name:     "config",
+		Usage:    "Config file containing provider's peer ID and private key",
```
willscott, line 117: why is the providers _private_ key needed?

willscott (commented):
I think i'm a bit confused about how this command is supposed to be used?

It's set up as though the provider is the one running 'register' (since they're the only one with access to their private key, but it seems to control the permissions for ingestion in a way that would be controlled by the operator of the indexing node

command/flags.go
gammazero: The register command is run by someone wanting to be a provider, and the request to register needs to be signed by the provider who is registering themselves and the indexer may be on a remote machine.  The provider being registered must be trusted by the indexer (indexer already knows indexer peerID), and the indexer needs to verify the request comes from the provider which the message signature does.  Trust just means that a provider can be registered without having to verify on-chain.

command/flags.go
gammazero: Well... Need the address too because this could be talking to an indexer running somewhere else.

This is more a simple client than a command.

internal/providers/registry.go
gammazero: oops, yes

command/register.go
gammazero: Changed wording a bit

gammazero:
I think I really should move some of the code that is in the command into `api/v0/client/`, so that the command simply calls a client method.

willscott (approved):

api/v0/client/libp2p/client.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	protocols := []protocol.ID{pid}
+
+	// Start a client
+	return &p2pclient{
+		ctx:       ctx,
+		host:      h,
+		self:      h.ID(),
+		peerID:    peerID,
+		protocols: protocols,
```
willscott, line 40: Does protocols need to be duplicated in both the client and messageSender? who is responsible for knowing it? would one finder client plausibly send on multiple protocols?

command/find.go
```diff
@@ -27,61 +27,55 @@ var GetCmd = &cli.Command{
...
-		/// for long-running peers.
+		// for long-running peers.
 		var host host.Host
 		host, err = libp2p.New(ctx)
 		if err != nil {
 			return err
 		}
-		cl, err = p2pclient.New(ctx, host)
+
+		peerID, err := peer.Decode(cctx.String("finder-host"))
```
willscott, line 62: should this not be 'indexer-host'?

gammazero: yep

api/v0/client/libp2p/client.go
gammazero: Looks like that was something left over from [before](https://github.com/filecoin-project/storetheindex/blob/main/api/v0/client/libp2p/finder.go#L46-L51), and protocols only is used in messageSender.

PR #42, opened 2021-08-20
Add utils to generate, sign and verify advertisements

adlrocha:
This PR adds all the utils required by providers and indexers as part of the ingestion protocol to generate IPLD-aware local indices, generate their corresponding advertisements, sign, and verify them.

_We need `go-ipld-prime > 0.11.0`  for this code to work. I'll keep it as a drat without rebasing until we perform the upgrade. I also may add minor changes while I integrate with the reference provider_


adlrocha:
@gammazero, JFYI, I am not blocked so no rush, but let me know when you upgrade `go-ipld-prime` so I can rebase my branch. Let me know if you need support with anything. Thanks!

api/v0/ingest/schema/envelope.go
```diff
@@ -0,0 +1,101 @@
...
+	"errors"
+
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	crypto "github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/record"
+	mh "github.com/multiformats/go-multihash"
+)
+
+const (
+	adSignatureCodec  = "/indexer/ingest/adSginature"
```
willscott, line 14:
```suggestion
	adSignatureCodec  = "/indexer/ingest/adSignature"
```

api/v0/ingest/schema/gen.go
```diff
@@ -70,16 +70,21 @@ func main() {
 		// ID is a multihash of IndexID+Provider+Previous
 		schema.SpawnStructField("ID", "Bytes", false, false),
 		schema.SpawnStructField("IndexID", "Link_Index", false, false),
-		schema.SpawnStructField("Previous", "Link_Advertisement", true, false),
+		// Empty bytes for the first advertisement
+		schema.SpawnStructField("PreviousID", "Bytes", false, false),
```
willscott, line 74: why bytes rather than an optional link?

api/v0/ingest/schema/envelope.go
```diff
@@ -0,0 +1,101 @@
...
+	adSignatureDomain = "indexer"
+)
+
+type advSignatureRecord struct {
+	domain *string
+	codec  []byte
```
willscott, line 20: are these for the smart/composable record format?

api/v0/ingest/schema/envelope.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+	bindex := lindex.(cidlink.Link).Cid.Bytes()
+	// Signature data is indexID+provider+previousAdvID
+	sigData := make([]byte, len(bindex)+len([]byte(provider))+len(previousAdvID))
+	copy(sigData[:len(bindex)], bindex)
+	copy(sigData[len(bindex):], []byte(provider))
+	copy(sigData[len(bindex)+len(provider):], previousAdvID)
+
+	return mh.Encode(sigData, mhCode)
```
willscott, line 60: consider using a codec / ipld encoder over a struct rather than doing this manually?

internal/utils/utils.go
```diff
@@ -64,3 +65,7 @@ func HasValue(values []indexer.Value, v indexer.Value) bool {
 	}
 	return false
 }
+
+func TestPeerID() (peer.ID, error) {
+	return peer.Decode("12D3KooWKRyzVWW6ChFjQjK4miCty85Niy48tpPV95XdKu1BcvMA")
```
willscott, line 70: ℹ️  complete or remove

api/v0/ingest/schema/envelope.go
adlrocha: They aren't, this comes from `go-libp2p-core/envelope`. @gammazero suggested using libp2p envelopes for signatures. It's the first time I've used them, so I may be using them wrong.

api/v0/ingest/schema/gen.go
adlrocha: When a `legSubscriber`sees an advertisement we don't want it to lfollow advertisement links, as it is only interested in the data. Indices point to their previous index so from an advertisement we can directly fetch all our pending updates.

This previousID will be used for the paginated ingestion protocol, or for indexers to request individual advertisements from a provider, thus the bytes. This is also the most convenient structure to use the `stopAt` selector.

Let me know if something doesn't make sense.

api/v0/ingest/schema/envelope.go
adlrocha: Can you elaborate on this a bit? I don't think there is any encoding to do here, I am just concatenating the three fields in bytes to compute their multihash in order to use the result as the ID for the advertisement. Thanks!

willscott: you have a serialization format of how to go from a structured object to a byte-stream, and then get a cid of it. That's what a codec does in getting a link out of an ipld object, but in a general way, i think

adlrocha: Ok, gotcha! Thanks. 

As AdvertisementID != AdvertisementLink because the the latter includes the signature and AdvertisementID only needs three of the fields of the node and is used to create the signature, I thought it made sense to do it this way, but I'll check if it is better to have an IPLD node with the three fields, and use its encoded link to create the signature.

adlrocha:
hey, @gammazero, we need `go-ipld-prime > 0.11.0` for this PR to work. I couldn't see any issue/PR tracking this so I don't know if you are approaching this work already. 

Do you have an estimate for when are we going to able to make this upgrade? Do you want me to start working on it or help you in any way? We can discuss it later today in the COLO. Thanks!

//cc @masih

PR #43, opened 2021-08-24
Remove all lotus imports and use raw jsonrpc over https instead

gammazero:
- Update to latest IPLD to generate code that works with go vet
- Simplify Discoverer interface and Discover function


adlrocha (approved):

internal/lotus/discovery.go
```diff
@@ -4,86 +4,123 @@ import (
...
-	tipSetKey := tipSet.Key()
+	u.Scheme = "https"
+	u.Path = "/rpc/v1"
 
+	return &Discoverer{
+		gatewayURL: u.String(),
+	}, nil
+}
+
+func (d *Discoverer) Discover(ctx context.Context, peerID peer.ID, minerAddr string) (*discovery.Discovered, error) {
```
adlrocha, line 57: Why are we using the peer.ID as an argument? Aren't we getting this info from lotus? (I don't have a deep knowledge of Lotus, just want to understand how it works :) )

PR #45, opened 2021-08-25
Added support for CAR entries in schema and utils

adlrocha:
This PR:
- Adds support for `CarEntries` in schema and utils.
- Removes graphSupport field.


masih (approved):
Nice 🍻

PR #46, opened 2021-08-27
Ingestion protocol

adlrocha:
_Work in progress_
Depends on: https://github.com/willscott/go-legs/pull/11

This PR:
- Adds `go-legs`-backed ingester using a graphsync hook to index data  received.
- Integrates ingester in daemon and exposes admin interface in http admin API.

Pending:
- [x] Integration with main process.
- [x] Fix flaky tests with multiple subscriptions.
- [ ] Ingester commands in CLI
- [ ] `Sync` implementation. (deferring until reference-indexer implements libp2p server for advertisement fetch)


internal/ingest/ingest.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+		return cid.Undef, err
+	}
+	_, c, err := cid.CidFromBytes(b)
+	return c, err
+}
+
+// Tracks latest sync for a specific peer.
+func (i *legIngester) putLatestSync(p peer.ID, c cid.Cid) error {
+	return i.ds.Put(datastore.NewKey(syncPrefix+p.String()), c.Bytes())
```
gammazero, line 228:
```suggestion
	if !c.Defined() {
		return nil
	}

	return i.ds.Put(datastore.NewKey(syncPrefix+p.String()), c.Bytes())
```
I think you want to ignore undefined cids because otherwise you get a huge number of them and will try to store them.  Put a debug print there and see.  Is that what should be expected?

gammazero (commented):
Review not complete, just a few comments so far

adlrocha:
@gammazero, after the rebase this PR is getting huge. I am still missing the `Sync` implementation, and a way to handle ingester subscriptions from CLI commands. `Sync` requires from implementing the libp2p server in the reference-provider, so if you are OK with it, I'd suggest merging this, and add that in a future PR. Thanks!

gammazero:
It is OK to merge it to avoid the PR growing forever.  Is there any chance of fixing the multiple subscriber problem first?

PR #54, opened 2021-09-01
implement provider sync

### LINKED ISSUE #50: Use t.Cleanup instead of defer in tests
Example: https://github.com/filecoin-project/storetheindex/blob/dbd6da01bfba0a641a6e1e8a136ef57b4a542d30/internal/ingest/ingest_test.go#L113

adlrocha:
This PR:

* Implements sync with providers.
* Integrates sync into admin API
* Fix #50


internal/ingest/ingest.go
```diff
@@ -147,14 +179,38 @@ func (i *legIngester) listenUpdates(ctx context.Context, s *sub) {
...
+
 // Unsubscribe to stop listening to advertisement from a specific provider.
 func (i *legIngester) Unsubscribe(ctx context.Context, p peer.ID) error {
 	i.sublk.Lock(p)
 	defer i.sublk.Unlock(p)
 	// Close subscriber
 	i.subs[p].ls.Close()
-	// Run cancel
-	i.subs[p].cncl()
+	// Check if we are subscribed
```
gammazero, line 209: Seems that there should be a check to make sure `i.subs[p]` returns a value.  Consider is `Unsubscribe` is called a second time.

gammazero (approved):
Had one comment

PR #57, opened 2021-09-03
Use libp2p envelope for all signed requests

gammazero:
- Include timestamp sequences with signed requests
- Veridy sequences in signed requests
- Registry performs periodic cleanup
- Added function to do provider polling periodically
- Defined SysError type to hold status code and error message
- Added IndexContent API to clients
- Moved ingest request handler functionality, common to all protocols, into single lib
- Removed internal/signature lib, now replaces by libp2p envelope


internal/providers/registry.go
```diff
@@ -56,6 +61,10 @@ type ProviderInfo struct {
 	LastIndex cid.Cid
 	// LastIndexTime is the time indexed data was added to the indexer
 	LastIndexTime time.Time
+
+	// lastContactTime is the time of last contact or attempted contact with
```
adlrocha, line 65: Remove?

api/v0/ingest/models/discover_request.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// ReadDiscoverRequest unmarshals a DiscoverRequest from bytes, verifies the
+// signature, and returns the DiscoverRequest
+func ReadDiscoverRequest(data []byte) (*DiscoverRequest, error) {
+	_, untypedRecord, err := record.ConsumeEnvelope(data, DiscoverRequestEnvelopeDomain)
+	if err != nil {
+		return nil, fmt.Errorf("cannot consume discover request envelope: %s", err)
+	}
+	rec, ok := untypedRecord.(*DiscoverRequest)
```
adlrocha, line 75: Out of a curiosity: we check here that the signature in the record is valid. Do we need to check also the source (i.e. that the signer is the right one?). The same applies to the other records.

api/v0/ingest/pb/ingest.proto
```diff
@@ -20,6 +20,8 @@ message IngestMessage {
         REGISTER_PROVIDER_RESPONSE=8;
         REMOVE_PROVIDER = 9;
         REMOVE_PROVIDER_RESPONSE=10;
+        INDEX_CONTENT=11;
+        INDEX_CONTENT_RESPONSE=12;
```
adlrocha, line 24: Just an FYI: in the reference_provider we chose to create an `UNSET = 0` type so by default the message type is `UNSET`. Just so it doesn't default to `ERROR_MESSAGE`. Its a nit-pick and I don't think we need to change it here, but so you know.

internal/handler/ingest_handler.go
```diff
@@ -0,0 +1,112 @@
...
+		AddrInfo: peer.AddrInfo{
+			ID:    peerRec.PeerID,
+			Addrs: peerRec.Addrs,
+		},
+	}
+	return h.registry.Register(info)
+}
+
+func (h *IngestHandler) ListProviders() ([]byte, error) {
+	infos := h.registry.AllProviderInfo()
```
adlrocha, line 67: We are not populating the `lastIndex` info for providers here, right? This is data is available in the datastore, so we could directly fetch it from there: https://github.com/filecoin-project/storetheindex/blob/94fbb1becd2018a2bd83b366f01eea8740740979/internal/ingest/ingest.go#L276

adlrocha (approved):
LGTM. Left a few comments but no blockers. I really like how you are using envelopes, so I'll switch to your approach for advertisement signatures.

api/v0/ingest/models/discover_request.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+type DiscoverRequest struct {
+	ProviderID    peer.ID
+	DiscoveryAddr string
+	Seq           uint64
+}
+
+// DiscoverRequestEnvelopeDomain is the domain string used for discover requests contained in a Envelope
+const DiscoverRequestEnvelopeDomain = "indexer-discover-request-record"
```
adlrocha, line 19: nit-pick: I don't know what is the standard in the libp2p world (I just found [this example from tests](https://github.com/libp2p/go-libp2p-core/blob/ea6cbf9ff2894c96aa014006c475207e01d6780a/record/record_test.go#L5)), but I think we should use a path format for domains: `indexer/discover/request/record`

adlrocha (approved):

api/v0/ingest/models/discover_request.go
gammazero: I think we have to rely on the source keeping their private key private.  The source's ID is the public key that is tied to that private key, and that peer ID is how we identify any source.

That being said, perhaps we need a way for a source to change their ID in case the think their private key has been compromized.  The current mechanism for that would be to do `DeleteProvider` and then do discovery using the new ID.

api/v0/ingest/pb/ingest.proto
gammazero: Do we have or need to have an error message for the reference provider?  If so, maybe change that to 0?

internal/handler/ingest_handler.go
gammazero: Seems like we should set the value in the registry here.  If we are not going to do that, then we should remove it from the registry and only fetch it from the datastore.  What do you think is better?

api/v0/ingest/models/discover_request.go
gammazero: I was following the example here: https://github.com/libp2p/go-libp2p-core/blob/v0.9.0/peer/record.go#L23

:shrug:

internal/providers/registry.go
gammazero: I am going to need it for polling, but it may end up being a timer... so will remove for now.

internal/handler/ingest_handler.go
gammazero: Added TODO comment

adlrocha: > The lastIndex and iastIndexTime are to record when an entire advertised index has been updated, so that the indexer can know if it can ignore advertisements. I think we probably want to update something else when individual pieces of content are updated. Maybe set a last update time for the provider?

We should link the logic in the ingestion protocol (that is where we are tracking this) with the provider info. It should be straightfoward with what we have now. Let's file an issue and decide what data we want to track. Thanks!

PR #68, opened 2021-09-14
Create place for finder handler functionality common to all protocols

gammazero:


adlrocha (approved):
Nice! This is something that we discussed about re-orging in the on-site. Thanks!

willscott (approved):

PR #71, opened 2021-09-16
Add a provider client

gammazero:
storetheindex should provide a provider-client that is used to communicate with providers.  storetheindex will use this client, and if someone wants to write their own indexer, they are welcome to also use this provider-client.

Simplified the libp2pclient so that there a separate client instance is created for each different peer to communicate with.  This means that there is no need to maintain and synchronize a map of different peers within a single client.  THis is advantageous when the app or goroutine only connects to a single peer.


internal/ingest/ingest.go
```diff
@@ -135,7 +136,13 @@ func (i *legIngester) Sync(ctx context.Context, p peer.ID, opts ...ingestion.Syn
 }
 
 func (i *legIngester) getLatestAdvID(ctx context.Context, p peer.ID) (cid.Cid, error) {
-	res, err := i.client.GetLatestAdv(ctx, p)
+	client, err := i.newClient(ctx, i.host, p)
```
adlrocha, line 139: What happens if two routines call a function for the same peer simultaneously? We will instantiate two different clients instance to the same peer and attempt to establish a stream from scratch? I don't think this is a problem because IIRC the libp2p host will check if it is already connected to the peer and leverage the existing stream. 

What I am not that sure is what would happen if we try to create a new stream while a previous one is still being negotiated. This was the reason for the message sender. Anyway, I think this can be easily tested.

adlrocha (commented):
Everything LGTM. Just a minor concern which may end up not being so, but worth checking.

internal/ingest/ingest.go
gammazero: I think this is a fairly normal occurrence in libp2p, when two goroutines try to talk to the same peer using the same protocol ID.  Either the existing stream gets reused or a new stream is created.

If we try to create a new stream while a previous one is still being negotiated, then either libp2p will wait until the first is created, or it will create a new stream to the same host with the same protocol ID.  I would guess that it would create a new stream, so that either stream could be reset independently.

This has got to be a common occurrence because different goroutines sharing the same host will try to make connections to some peer/proto.  The different goroutines should not have to coordinate which host they connect to.

This should be documented behavior, but lets test anyway.

adlrocha (approved):

internal/libp2pclient/client.go
```diff
@@ -35,93 +33,141 @@ type Client struct {
...
-	return ms
+
+	return nil
 }
 
-func (c *Client) messageSenderForPeer(ctx context.Context, peerID peer.ID) (*peerMessageSender, error) {
-	ms := c.peerSender(peerID)
+func (c *Client) closeStream() {
+	_ = c.stream.Reset()
+	_ = c.r.Close()
```
gammazero, line 142: I think this `c.r.Close()` may be unnecessary since I think `c.stream.Reset()` takes care of that.  Can this be confirmed?

internal/ingest/ingest.go
adlrocha: I ran a test with a few dozen clients connecting to the same server in parallel and nothing broke, so I am happy with this :)

internal/libp2pclient/client.go
adlrocha: I think you are right. We create the `ReadCloser` from the stream in `c.r = msgio.NewVarintReaderSize(nstr, network.MessageSizeMax)`, and `stream.Reset()` closes the stream (and underlying readers) and tells the other end to close its own side.

PR #84, opened 2021-09-22
Initial structure for prometheus metrics on admin server

willscott:


gammazero (commented):
This all looks good, but hold off for a day on merging it until I have time to do the equivalent with OpenTelemety.  @Stebalien was much in favor of using OT.  The code should actually be very similar, and we can export metrics to an agent that will make them available for prometheus.

gammazero (approved):
Yes, this is exactly what I was looking for.

internal/metrics/server.go
```diff
@@ -0,0 +1,69 @@
...
+// Global Tags
+var (
+	Version, _ = tag.NewKey("version")
+
+	Method, _ = tag.NewKey("method")
+)
+
+// Measures
+var (
+	IngestCIDs    = stats.Int64("ingest/cids", "Number of CIDs ingested", stats.UnitDimensionless)
```
gammazero, line 24: Change "CIDs" to "multihashes"

internal/metrics/server.go
```diff
@@ -0,0 +1,69 @@
...
+	IngestChange  = stats.Int64("ingest/change", "Number of ingest triggers received", stats.UnitDimensionless)
+	ProviderCount = stats.Int64("provider/count", "Number of know (registered) providers", stats.UnitDimensionless)
+)
+
+// Views
+var (
+	IngestCIDView = &view.View{
+		Measure:     IngestCIDs,
```
gammazero, line 32: Here also, `IngestCIDs` => `IngestMultihashes`

PR #86, opened 2021-09-28
Rename packages to improve names and follow naming conventions

gammazero:
A PR that will follow this will make additional breaking interface changes to support context ID and communicate whether or not double-hashing is enabled, so this PR is getting other unrelated interface changes done.

- Renamed internal/utils => test/util
- Renamed internal/providers => internal/registry
- Renamed api/v0/finder/models => api/v0/finder/model
- Renamed api/v0/ingest/models => api/v0/indest/model


willscott (approved):

gammazero:
I would like to first merge #86 and a corresponding PR in the reference provider first. Then I will merge this if it is ready.

PR #90, opened 2021-10-06
Supply ContextID to indexer as part of Value

gammazero:
* Advertisements now contain a ContextID that serves as the provider's key to metadata
* storetheindex uses ContextID from advertisement to compose value to index
* Read content linked from advertisement and store in indexer asynchronously to reading from provider
* Move admin client methods out of finder client

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -2,12 +2,11 @@ package ingest
 
 import (
 	"bytes"
-	"encoding/base64"
 	"io"
 	"net/http"
 
 	"github.com/filecoin-project/go-indexer-core"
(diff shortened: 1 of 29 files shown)


internal/ingest/linksystem.go
```diff
@@ -246,6 +250,14 @@ func (i *legIngester) processEntries(adCid cid.Cid, p peer.ID, nentries ipld.Nod
 		log.Errorf("Error decoding entries: %s", err)
 		return err
 	}
+
+	const batchSize = 64
```
willscott, line 254: consider making configurable?

willscott (approved):

PR #92, opened 2021-10-12
Use new indexer.Metadata type

gammazero:
Instead of passing a protocol ID and a byte slice, or an ID encoded into a byte slice, storetheindex APIs use a `indexer.Metadata` type that has explicit `ProtocolID` and `Data` members.

### DIFF
--- a/api/v0/finder/model/model.go
+++ b/api/v0/finder/model/model.go
@@ -17,13 +17,14 @@ type FindRequest struct {
 	Multihashes []multihash.Multihash
 }
 
+// ProviderResult is a one of possibly multiple results when looking up a
+// provider of indexed context.
 type ProviderResult struct {
-	// ContextID identifies the metadata that is part of this value
+	// ContextID identifies the metadata that is part of this value.
 	ContextID []byte
-	// Metadata is serialized data that provides information about retrieving
-	// data, for the indexed CID, from the identified provider.
-	Metadata []byte `json:",omitempty"`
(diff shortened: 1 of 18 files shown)


api/v0/finder/model/model.go
```diff
@@ -17,13 +17,14 @@ type FindRequest struct {
...
-	// ContextID identifies the metadata that is part of this value
+	// ContextID identifies the metadata that is part of this value.
 	ContextID []byte
-	// Metadata is serialized data that provides information about retrieving
-	// data, for the indexed CID, from the identified provider.
-	Metadata []byte `json:",omitempty"`
-	// Privider is the peer ID of the provider and its multiaddrs
+	// Metadata contains information for the provider to use to retrieve data.
+	Metadata indexer.Metadata
+	// Privider is the peer ID of the provider and its multiaddrs.
```
willscott, line 27:
```suggestion
	// Provider is the peer ID and addresses of the provider.
```

willscott (approved):

api/v0/finder/model/model.go
```diff
@@ -56,9 +57,11 @@ func (pr ProviderResult) Equal(other ProviderResult) bool {
 }
 
 func ProviderResultFromValue(value indexer.Value, addrs []multiaddr.Multiaddr) ProviderResult {
+	metadata, _ := indexer.DecodeMetadata(value.MetadataBytes)
```
mvdan, line 60: do we want to ignore this error?

PR #95, opened 2021-10-13
api/v0: reject Metadata.ProtocolID == 0

mvdan:
(see commit message)


server/admin/http/handler.go
```diff
@@ -133,10 +133,18 @@ func (h *adminHandler) importManifest(w http.ResponseWriter, r *http.Request) {
...
+		log.Errorf("Error putting entries in indexer: %s", err)
+		http.Error(w, "", http.StatusInternalServerError)
+		return
+	}
+
 	value := indexer.Value{
 		ProviderID:    provID,
 		ContextID:     contextID,
-		MetadataBytes: v0.Metadata{}.Encode(),
+		MetadataBytes: encMetadata,
```
mvdan, line 147: Loking for input as to what these should do instead. Evidently they aren't covered by our tests, because that MarshalBinary call should always error.

I see two options:

1) Keep MetadataBytes empty. On the receiver side, do not decode MetadataBytes if it is empty.
2) Use a real Metadata with a valid ProtocolID.

Ultimately, I think this depends on what the "zero metadata" here is meant to be useful for. I'm not sure I understand that.

gammazero: This particular handler was not implemented by me, so I cannot say for certain, but I believe that the intent was to provide some metadata at a later point, which was not done.

Minimally, the metadata should always specify a protocol ID, even if there is no data component to the metedata.  IMO what is needed here is some additional input to supply a `v0.Metadata`

api/v0/metadata.go
```diff
@@ -17,30 +27,47 @@ type Metadata struct {
...
+
 // Equal determines if two Metadata values are equal.
 func (m Metadata) Equal(other Metadata) bool {
 	return m.ProtocolID == other.ProtocolID && bytes.Equal(m.Data, other.Data)
 }
 
-// EncodeMetadata serializes Metadata to []byte.
-func (m Metadata) Encode() []byte {
+// MarshalBinary implements encoding.BinaryMarshaler.
+func (m Metadata) MarshalBinary() ([]byte, error) {
```
gammazero, line 41: 🎉

server/admin/http/handler.go
mvdan: Now leaves MarshalBytes empty.

PR #99, opened 2021-10-15
add first version of an end-to-end test

mvdan:
(see commit message)

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -184,14 +184,12 @@ func daemonCommand(cctx *cli.Context) error {
 
 		p2pfinderserver.New(ctx, p2pHost, indexerCore, registry)
 		p2pingestserver.New(ctx, p2pHost, indexerCore, registry)
-		log.Infow("libp2p servers initialized", "host_id", p2pHost.ID(), "multiaddr", p2pmaddr)
 
-		// Initialize ingester if libp2p enabled.
+		// Initialize ingester.
 		ingester, err = legingest.NewLegIngester(ctx, cfg.Ingest, p2pHost, indexerCore, registry, dstore)
 		if err != nil {
 			return err
 		}
-		log.Info("libp2p ingester initialized")
 
 		// Subscribe to pubsub channel if a pubsub host is configured
 		if cfg.Ingest.PubSubPeer != "" {
@@ -204,6 +202,8 @@ func daemonCommand(cctx *cli.Context) error {
 				log.Errorf("Cannot subscribe to provider", "err", err)
 			}
 		}
+
+		log.Infow("libp2p servers initialized", "host_id", p2pHost.ID(), "multiaddr", p2pmaddr)
 	}
 
 	// Create admin HTTP server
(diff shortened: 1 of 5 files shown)


command/daemon.go
```diff
@@ -232,6 +232,10 @@ func daemonCommand(cctx *cli.Context) error {
 		errChan <- ingestSvr.Start()
 	}()
 
+	// Wait for http servers to start.
+	time.Sleep(time.Second)
```
mvdan, line 236: surely we don't need a sleep to know that the http server is ready? :)

e2e_test.go
```diff
@@ -148,17 +148,25 @@ func TestEndToEndWithReferenceProvider(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Initialized provider ID: %s", cfg.Identity.PeerID)
 
-	cmdIndexer := e.start(indexer, "daemon")
+	e.run(indexer, "init", "--pubsub-peer", cfg.Identity.PeerID)
+	cfg, err = config.Load(filepath.Join(e.dir, ".storetheindex", "config"))
+	if err != nil {
+		t.Fatal(err)
```
mvdan, line 179: please use quicktest for consistency

command/daemon.go
gammazero: The HTTP servers are started in separate goroutines, and there is not any channel reporting when they are ready.  This was to be reasonably sure that those servers were ready to accept requests before logging the "Indexer ready" message.

Will remove as it is not necessary.

e2e_test.go
```diff
@@ -14,7 +14,6 @@ import (
 	"time"
 
 	"github.com/filecoin-project/storetheindex/config"
-	qt "github.com/frankban/quicktest"
```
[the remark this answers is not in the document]
gammazero: There was an error that `go get` was needed to get the package, yet your PR did not modify go.mod so instead of determining why that complaint only occurred on some platforms, I just eliminated the issue.

I have reverted the change that removed quicktest.

e2e_test.go
```diff
@@ -41,7 +40,9 @@ func (e *e2eTestRunner) run(name string, args ...string) []byte {
 	cmd := exec.CommandContext(e.ctx, name, args...)
 	cmd.Env = e.env
 	out, err := cmd.CombinedOutput()
-	qt.Assert(e.t, err, qt.IsNil, qt.Commentf("output: %s", out))
+	if err != nil {
+		e.t.Fatalf("output: %s", out)
```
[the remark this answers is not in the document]
gammazero: Fixed.

gammazero:
The changes that made the test pass were:
- Configure storetheindex to subscribe to the provider's pubsub messages.  Necessary for indexer to receive pubsub message.
- Download the sample car file.  Necessary to fix "no such file" error.

mvdan:
Ah, the test doesn't require the "find" commands to succeed, just yet. Doing that last change.

PR #104, opened 2021-11-04
Remove indexed content by context ID

gammazero:
When a remove advertisement does not contain a list of entries, then remove all entries that have the context ID specified by the advertisement.

Other changes included:
- Update last advertisement in registry
- Record last contact time for each provider in registry (in prep for polling)
- Periodically update storage size metric
- Simplify batch update logic (only need to do remove or put, not both)
- Do not use "i" as the name of a receiver
- Provider info API returns LastAdvertisement instead of LastIndex
- Misc. code cleanup

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -104,7 +104,8 @@ func NewLegIngester(ctx context.Context, cfg config.Ingest, h host.Host,
 	return li, nil
 }
 
-// metricsUpdate periodically updates certains metrics
+// metricsUpdate periodically updates metrics.  This goroutine exits when the
+// sigUpdate channel is closed, when Close is called.
 func (li *legIngester) metricsUpdater() {
 	var hasUpdate bool
 	t := time.NewTimer(time.Minute)
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -55,6 +55,7 @@ type legIngester struct {
 	sublk *keymutex.KeyMutex
 
 	batchSize int
+	sigUpdate chan struct{}
 }
 
 // subscriber datastructure for a peer.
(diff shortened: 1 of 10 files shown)


internal/ingest/linksystem.go
```diff
@@ -345,86 +324,61 @@ func (i *legIngester) processEntries(adCid cid.Cid, p peer.ID, nentries ipld.Nod
...
-// error channel to indicate completion.
-func batchIndexerEntries(batchSize int, putChan, removeChan <-chan multihash.Multihash, value indexer.Value, idxr indexer.Interface) <-chan error {
+// The goroutine exits when the input channel is closed.  It closes the error
+// channel to indicate completion.
+func (li *legIngester) batchIndexerEntries(mhChan <-chan multihash.Multihash, value indexer.Value, isRm bool) <-chan error {
+	var indexFunc func(indexer.Value, ...multihash.Multihash) error
+	var opName string
+	if isRm {
+		indexFunc = li.indexer.Remove
+		opName = "remove"
```
willscott, line 342: can we have constant / enum for these rather than strings?

willscott (approved):

internal/ingest/ingest.go
```diff
@@ -91,37 +92,70 @@ func NewLegIngester(ctx context.Context, cfg config.Ingest, h host.Host,
...
 		sublk:     keymutex.New(0),
 		batchSize: cfg.StoreBatchSize,
+		sigUpdate: make(chan struct{}, 1),
 	}
 
 	// Register storage hook to index data as we receive it.
 	lms.GraphSync().RegisterIncomingBlockHook(li.storageHook())
 	log.Debugf("LegIngester started and all hooks and linksystem registered")
+
+	go li.metricsUpdater()
```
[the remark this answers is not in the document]
gammazero: The `li.sigUpdate` channel is closed at shutdown, causing the goroutine to exit (will add comment).

internal/ingest/linksystem.go
gammazero: It is only for a log message, here only, so no.

PR #105, opened 2021-11-15
Upgrade to latest `go-legs`

masih:
Upgrade dependency to `go-legs`. The upgrade introduces a breaking
change where a subscriber needs to also specify a selector. The selector
specified aims to recursively sync the graph of advertisements and their
chain of entry chunks. This is equivalent to how the indexer node
operated before selectors existed in legs.

Relates to:
 - https://github.com/filecoin-project/indexer-reference-provider/pull/81


internal/ingest/ingest.go
```diff
@@ -71,7 +71,7 @@ func NewLegIngester(ctx context.Context, cfg config.Ingest, h host.Host,
 	idxr *indexer.Engine, reg *registry.Registry, ds datastore.Batching) (LegIngester, error) {
 
 	lsys := mkLinkSystem(ds, reg)
-	lms, err := legs.NewMultiSubscriber(ctx, h, ds, lsys, cfg.PubSubTopic)
+	lms, err := legs.NewMultiSubscriber(ctx, h, ds, lsys, cfg.PubSubTopic, nil)
```
masih, line 74: @gammazero @willscott I'd love your input on whether there are any plans to utilise selectors in storetheindex. For now setting this to `nil` so that the work on dependency restructuring can carry on in reference provider.

willscott, line 74: we should specify a selector that pulls the advertisements but not the car indexes themselves here.

willscott: Yes. this interface change came from an ask we got in integration:
* The primary legs sync should only sync the advertisments. Each list of CIDs getting pulled over backed by a car should be it's own graphsync session. Otherwise, you have a risk of memory exhaustion on the server, since one transaction will open and retain references to all deal files since graphsync holds onto previously transferred blocks in order to manage deduplication within a session.

masih: Selector matcher with `Has_field` condition is not implemented in IPLD.

Can you think of a way to work around it here (other than implementing the selector condition in IPLD)?

willscott: I think you want a recursive selector but instead of a 'explore all' in the recursion you want an ['explore fields'](https://github.com/ipld/go-ipld-prime/blob/master/traversal/selector/exploreFields.go#L20)

masih: sigh... thank you

config/ingest.go
```diff
@@ -14,4 +15,6 @@ type Ingest struct {
 	// StoreBatchSize is the number of entries in each write to the value
 	// store.  Specifying a value less than 2 disables batching.
 	StoreBatchSize int
+	// SelectorRecursionLimitDepth sets the maximum recursion depth for the selector used to select advertisements.
+	SelectorRecursionLimitDepth int
```
gammazero, line 19: How do I set this to unlimited?  Specify `0` or `-1`?

masih: > Also, maybe there should be something similar to this:
> https://github.com/filecoin-project/index-provider/blob/main/config/config.go#L97-L98

Off topic: I am curious why call this `defaults`? would something like `reset` or even a constructor for the config struct type like `newIngestConfig` that constructs the type with default values set make sense?

masih: > How do I set this to unlimited? Specify 0 or -1?

It looks like this depends on `RecursionLimit` mode: There is a mode dedicated to `RecursionLimit_None` i.e. no limits to recursion. It is unclear to me how to set the limit on _depth_ to unlimited.

I'll add some gluecode here to switch the mode to `None` if `SelectorRecursionLimitDepth` is set to `-1`.

internal/ingest/ingest.go
masih: @willscott  would you mind taking a look at my selector to check it's the right one? 🍻

willscott: that looks like the right thing to me 👍

masih:
@gammazero I'd love your help in understanding why tests fail to find keys after the selector is added assuming the selector is correct

willscott:
@masih the selector change now means that only advertisements are synced by the default legs sync, not the lists of CIDs yourself. with the more limited selector, you'll need to kick off explicit `legs.Sync` calls each time you want to access the individual advertisement cid lists from the subscriber.

config/ingest.go
gammazero: > why call this defaults? would something like ... `newIngestConfig` that constructs the type with default values set make sense?

The `defaults` function is run _after_ the config is loaded, to replace any zero-valued items with the appropriate real default value (when it is different from an item's zero-value).  So, a constructor would not be quite right because the config object is already built/loaded.

If `SelectorRecursionLimitDepth` is not configured in the config file, I would expect a default value to be used.  Similarly, if I set it to the zero-value (0 in this case), I would also expect some reasonable default to be used.

gammazero: minor:  It sounds better to me to call this `SelectorRecursionDepthLimit`.  "LimitDepth" sounds more like a boolean deciding whether to limit the depth, whereas "DepthLimit" sounds like a maximum depth.

willscott:
The other point in syncing the contents is something I'm not totally sure what the answer is, and maybe @hannahhoward can comment on:

* we may already have the cid for the contents of the advertisement in our local data store. eg. if we got it from a different provider already. we shouldn't in those cases re-fetch it.
* we may have part of an advertisement linked-list chain - we started the graphsync fetch, got the root cid and part of the list, but it got interrupted.. Is there a 'cheap' way to know if we got the full linked list, or if we need to resume the fetch of it?

gammazero:
> we may already have the cid for the contents of the advertisement in our local data store. eg. if we got it from a different provider already. we shouldn't in those cases re-fetch it.

If we have the same advertisement CID, but from a different provider (seems like something that would not happen), then we would still need to get all the multihashes for that advertisement so that the indexer could associate them with another provider.  There is no good way[1] to search for these in the indexer, so fetching the content for the advertisement would be necessary anyway.

> we may have part of an advertisement linked-list chain ...

The linksystem cache may prevent fetching chunks that we already have, but it is a cache so not guaranteed.  I think a better approach is for the indexer to store not only the latest advertisement but also the latest advertisement that has been synced.

[1] Requires iterating all values to find values with matching contextID, and then iterating all indexes to find each that map to any of those values.

config/ingest.go
masih: Note that I have removed this config completely from this PR, and captured #123 which depends on go-legs changes to expose limit.

internal/ingest/ingest.go
```diff
@@ -71,7 +77,24 @@ func NewLegIngester(ctx context.Context, cfg config.Ingest, h host.Host,
...
+		recLimit = selector.RecursionLimitNone()
+	} else {
+		recLimit = selector.RecursionLimitDepth(int64(cfg.SelectorRecursionLimitDepth))
+	}
+	advSelectorNode := ssb.ExploreRecursive(
+		recLimit,
+		ssb.ExploreFields(
+			func(efsb builder.ExploreFieldsSpecBuilder) {
+				efsb.Insert("PreviousID", ssb.ExploreRecursiveEdge())
+			})).Node()
```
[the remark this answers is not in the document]
masih: Already handled on go-legs side; [added docs to go-legs](https://github.com/filecoin-project/go-legs/blob/main/multisubscribe.go#L48-L56) to clarify.

internal/ingest/ingest.go
```diff
@@ -173,7 +197,7 @@ func (li *legIngester) Sync(ctx context.Context, peerID peer.ID, opts ...SyncOpt
 	// Start syncing. Notifications for the finished
 	// sync will be done asynchronously.
 	log.Debugf("Started syncing process with provider %s", sub)
-	watcher, cncl, err := sub.ls.Sync(ctx, peerID, c)
+	watcher, cncl, err := sub.ls.Sync(ctx, peerID, c, li.advSelectorNode)
```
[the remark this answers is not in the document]
masih: Discussed in 1:1 meetings. Summary of approach: this PR introduces a selector that would sync an advertisement _and_ its linked chunks, a similar behaviour to how it worked before selector was introduced. This means we do not need to re-engineer the storage hook mechanism in storetheindex link system. Instead, approach that in future PRs along with other opportunities for optimisation, e.g. batch processing of advertisements instead of one-by-one.

masih:
@willscott @gammazero  this PR is now ready for review. The approach I have taken here is:
to introduce a selector that selects both advertisements and their chunks. This means the storage hook mechanism that handles the processing of chunks does not need to be re-engineered, which allows us to provide a thiner iteration of working indexer node and move optimisations to a separate PR. There are lots of things that we _could_ optimise in the DAG resolution and it is not clear at this stage which to prioritise.

Example optimisations we can do include:
* batch processing of advertisements _if we receive many at once_ say when a provider has not advertised for a while.
* reusing entry chunks that are already resolved across more than one provider ID
* resume resolution of entry chunks

I will capture an issue for optimisations.

masih:
> we may already have the cid for the contents of the advertisement in our local data store. eg. if we got it from a different provider already. we shouldn't in those cases re-fetch it.

I am curious does graphsync not already check local datastore via link system for presence of a link before making a remote call to fetch it? Datastore across all moving parts in `legIngester` is shared.

Aside from the block resolution, it seems there are [internal mappings of chunk entry link to ad](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/linksystem.go#L387) that are updated on block persistence and entry processing. That mapping fundamentally assumes a chunk entry link uniquely maps to a single advertisement. We need to update those to when introducing this optimisation.

willscott:
@masih I don't think graphsync does do that check because the remote side can't know what's in your local blockstore. You need to tell it (via the extension of 'don't send <cid>' currently) if you want to skip specific blocks in the traversal. Since there isn't a graphsync extension yet for 'don't send <cid> or it's subdag' we instead handle that currently by baking it into the selector.

internal/ingest/ingest.go
```diff
@@ -71,7 +73,18 @@ func NewLegIngester(ctx context.Context, cfg config.Ingest, h host.Host,
...
 
 	lsys := mkLinkSystem(ds, reg)
-	lms, err := legs.NewMultiSubscriber(ctx, h, ds, lsys, cfg.PubSubTopic)
+
+	// Construct a selector that recursively looks for nodes with field
+	// "PreviousID" as per Advertisement schema.
+	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
+	advAndChunkSel := ssb.ExploreFields(
+		func(efsb builder.ExploreFieldsSpecBuilder) {
+			efsb.Insert("Entries", ssb.ExploreRecursiveEdge())
```
willscott, line 82: i thought we didn't want the basic advertisement sync/selector to include the fetch of entries / next.

masih: The rationale is to have that work done in a separate PR because it's beyond a simple dependency upgrade and involves changing the way storage hooks work among other changes.

I have explained more here: https://github.com/filecoin-project/storetheindex/pull/105#issuecomment-979249687

willscott (approved):

masih:
>  I don't think graphsync does do that check because the remote side can't know what's in your local blockstore. You need to tell it (via the extension of 'don't send ' currently) if you want to skip specific blocks in the traversal. Since there isn't a graphsync extension yet for 'don't send or it's subdag' we instead handle that currently by baking it into the selector.

Oh I see. Thank you for pointing that out. 
So the entire dag is transferred back if the caller asks for the root. 
I thought, or maybe wished, the caller is consulted on each link traversal before the block that corresponds to it is transferred.

PR #106, opened 2021-11-16
adapt e2e test to new index-provider Go module

mvdan:
(see commit message)


masih (approved):

mvdan:
Looks like I hit a TestSubscribe flake. Merging.

mvdan:
By which I mean, that test uses a libp2p-pubsub sleep, so some flakes are to be expected.

PR #125, opened 2021-11-28
Bootstrap

gammazero:
- Add bootstrap functionality.  This connects the indexer to configured peers to enable the exchange of gossip pubsub.
- Configure list of PubSub peers.  This allows the indexer to accept pubsub messges that _originate_ from these peers.   This is a temporary fix until go-legs can use the indexer's allow/deny policy to determine whether or not to accept a pubsub message.


config/bootstrap.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	defaultMinimumPeers     = 1
+	defaultBootstrapPeriod  = 30 * time.Second
+	defaulConnectionTimeout = defaultBootstrapPeriod / 3
+)
+
+// defaultBootstrapAddresses are the hardcoded bootstrap addresses.
+var defaultBootstrapAddresses = []string{}
```
willscott, line 19: are you planning to load this in via the `go build` or some other way?

willscott (approved):

config/bootstrap.go
gammazero: No.  This is just a place holder for if/when we want to hardcode some bootstrap addresses as IPFS does.

willscott: we should find a way to default to the filecoin bootstrap nodes, probably

willscott: namely we should use https://github.com/filecoin-project/lotus/blob/7a55e8e892a2ed7a76ae4a75c1e33580688c3d0e/build/bootstrap/mainnet.pi

gammazero: Added hardcoded values.  These are written into the config file during `./storetheindex init`

PR #132, opened 2021-12-08
Implement the ability to change log level at runtime

masih:
Add admin server and cli functionality to set log level for each logging
subsystem or batch set level by matching subsystem name via regex.


api/v0/admin/client/http/client.go
```diff
@@ -111,26 +113,71 @@ func (c *Client) Unsubscribe(ctx context.Context, provID peer.ID) error {
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u, nil)
+	if err != nil {
+		return err
+	}
+
+	resp, err := c.c.Do(req)
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
```
gammazero, line 177: 🎉

command/config.go
```diff
@@ -0,0 +1,87 @@
...
+	"errors"
+	"fmt"
+
+	httpclient "github.com/filecoin-project/storetheindex/api/v0/admin/client/http"
+	"github.com/urfave/cli/v2"
+)
+
+var ConfigCmd = &cli.Command{
+	Name:  "config",
+	Usage: "Dynamically modifies and shows the current configuration",
```
gammazero, line 13: Looks like this pertains to the log levels config.  Is this comment correct, because it seems to say that it can modify any part of the config.

gammazero (approved):
LGTM, with one question

command/config.go
masih: I envisioned eventually the `config` command does that. For now it just modifies the log level.

Happy to change that description to be less aspirational

PR #138, opened 2021-12-14
Use the built-in go-legs head publisher to syn latest avertisement ID

### LINKED ISSUE #137: Use go-legs head publisher to get the latest advertisement ID
... [instead of the libp2p server exposed by provider](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/ingest.go#L208) because it would be good to converge on common mechanisms to  do the same thing and that would reduce the number of endpoints exposed by the provider.

masih:
Instead of connecting to the provider's libp2p server use the mechanism
built into go-legs that automatically gets head if unspecified and uses
default selector.

Fixes #137


gammazero (approved):
LGTM

PR #139, opened 2021-12-15
Selectively sync advertisements then entries only if absent

### LINKED ISSUE #136: Sync advertisements DAG selectively
When syncing an advertisement use a selector, such that:
* the traversal stops at the latest seen advertisement
* the chunks are excluded

Exclude the chunks so that:
* linksystem can be checked for its presence before pull channel is opened.
* we reduce unnecessary traffic in cases where multiple providers provide the same CAR content.

For syncing the chain of entries a full traversal via explore all selector is permitted, as long as the link system is checked for the root of entries chain prior to open pull channel.

masih:
Configure the background leg sync to only sync advertisement chain. Sync
entries separately via storage hook by first checking if they are
present already.

Note that the semantics of explicit Sync is changed by this commit where
Sync returns when the advertisement is synced _excluding_ entries.

Change ingest tests to assert that expected multihashes are indexed
eventually. Remove evil `time.Sleep`s where possible in existing tests.

Fixes #136

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -72,15 +72,15 @@ func NewLegIngester(ctx context.Context, cfg config.Ingest, h host.Host,
 
 	// Construct a selector that recursively looks for nodes with field
 	// "PreviousID" as per Advertisement schema.
+	// Note that the entries within an advertisement are synced separately triggered by
+	// storage hook, so that we can check if a chain of chunks exist already before syncing it.
 	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
-	advAndChunkSel := ssb.ExploreFields(
+	adSel := ssb.ExploreFields(
 		func(efsb builder.ExploreFieldsSpecBuilder) {
-			efsb.Insert("Entries", ssb.ExploreRecursiveEdge())
 			efsb.Insert("PreviousID", ssb.ExploreRecursiveEdge())
-			efsb.Insert("Next", ssb.ExploreRecursiveEdge())
 		}).Node()
 
-	lms, err := legs.NewMultiSubscriber(ctx, h, ds, lsys, cfg.PubSubTopic, advAndChunkSel)
+	lms, err := legs.NewMultiSubscriber(ctx, h, ds, lsys, cfg.PubSubTopic, adSel)
 	if err != nil {
 		log.Errorf("Failed to state LegTransport in ingester: %s", err)
 		return nil, err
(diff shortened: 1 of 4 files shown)


internal/ingest/ingest.go
```diff
@@ -135,38 +135,44 @@ func (li *legIngester) metricsUpdater() {
...
+// Sync syncs the latest advertisement from peerID.
+// This is done by first fetching the latest advertisement ID from the given peer and sync it
+// until the traversal reaches the last seen advertisement.
+//
+// Note that the multihash entries corresponding to the advertisement are synced in the background.
+// The completion of advertisement sync does not necessarily mean that the entries corresponding to
+// the advertisement are synced.
 func (li *legIngester) Sync(ctx context.Context, peerID peer.ID, opts ...SyncOption) (<-chan multihash.Multihash, error) {
-	log.Debugf("Syncing with peer %s", peerID)
+	log := log.With("peerID", peerID)
```
gammazero, line 146: 👍

internal/ingest/ingest_test.go
```diff
@@ -66,10 +66,10 @@ func TestSubscribe(t *testing.T) {
 	// Test with two random advertisement publications.
 	_, mhs := publishRandomAdv(t, i, lph, lp, lsys, false)
 	// Check that the mhs have been indexed correctly.
-	i.checkMhsIndexed(t, lph.ID(), mhs)
+	i.checkMhsIndexedEventually(t, lph.ID(), mhs, 2*time.Second, 10*time.Second)
```
gammazero, line 69: The timeouts should be defined as consts or made configurable.

If they are always the same for each call to `i.checkMhsIndexedEventually`, then set them when the `Ingester` is instantiated.

internal/ingest/ingest_test.go
```diff
@@ -307,12 +307,40 @@ func (i *legIngester) checkMhsIndexed(t *testing.T, p peer.ID, mhs []multihash.M
...
+
+func (i *legIngester) checkMhsIndexedEventually(t *testing.T, p peer.ID, mhs []multihash.Multihash, interval time.Duration, timeout time.Duration) {
+	requireTrueEventually(t, func() bool { return i.providesAll(p, mhs...) }, interval, timeout, "multihashes were not indexed")
+}
+
+func (i *legIngester) providesAll(p peer.ID, mhs ...multihash.Multihash) bool {
+	for _, mh := range mhs {
+		values, exists, err := i.indexer.Get(mh)
+		if err != nil || !exists {
+			return false
```
gammazero, line 319: Log error here?

internal/ingest/ingest_test.go
```diff
@@ -333,3 +361,19 @@ func publishRandomAdv(t *testing.T, i *legIngester, lph host.Host, lp legs.LegPu
...
+	defer cancel()
+	for {
+		if attempt() {
+			return
+		}
+		select {
+		case <-ctx.Done():
+			require.FailNow(t, "timed out awaiting eventual success", msgAndArgs...)
+			return
+		case <-time.After(interval):
```
gammazero, line 376: Create a `time.Timer` before the loop, and reset it here instead of using `time.After()` in the loop.

gammazero (changes requested):
Looks good, but have one required change: "Create a time.Timer before the loop,..."

internal/ingest/linksystem.go
```diff
@@ -147,67 +149,122 @@ func verifyAdvertisement(n ipld.Node) (schema.Advertisement, error) {
...
 		if err != nil {
-			log.Errorf("Error deleting index from datastore: %s", err)
+			log.Errorw("Error deleting index from datastore", "err", err)
 		}
 	}
 }
 
+func (li *legIngester) syncAdEntries(ctx context.Context, from peer.ID, ad schema.Advertisement) error {
+	log := log.With("peerID", from)
+	sub, err := li.newPeerSubscriber(ctx, from)
```
gammazero, line 227: After creating this subscriber to sync the ad entries, does it need to be shutdown when the sync is done?

gammazero (approved):

internal/ingest/linksystem.go
gammazero: After the legs broker stuff is available, we will just call Sync() here and not have to create a subscriber at all.  The broker will handle using the existing handler (or creating one if needed).

masih: I think the name `newPeerSubscriber` is misleading; it actually gets or creates a new subscriber and there is long-lived subscriber map in ingest engine. See existing implementation [here](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/ingest.go#L267).

internal/ingest/ingest.go
```diff
@@ -72,15 +72,15 @@ func NewLegIngester(ctx context.Context, cfg config.Ingest, h host.Host,
 
 	// Construct a selector that recursively looks for nodes with field
 	// "PreviousID" as per Advertisement schema.
+	// Note that the entries within an advertisement are synced separately triggered by
+	// storage hook, so that we can check if a chain of chunks exist already before syncing it.
 	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
-	advAndChunkSel := ssb.ExploreFields(
+	adSel := ssb.ExploreFields(
 		func(efsb builder.ExploreFieldsSpecBuilder) {
-			efsb.Insert("Entries", ssb.ExploreRecursiveEdge())
```
MarcoPolo, line 78: ah yes, I was wondering earlier why these were here. Thanks for clarifying!

internal/ingest/ingest.go
```diff
@@ -135,38 +135,44 @@ func (li *legIngester) metricsUpdater() {
...
 }
 
-// Sync with a data provider up to latest ID.
+// Sync syncs the latest advertisement from peerID.
+// This is done by first fetching the latest advertisement ID from the given peer and sync it
+// until the traversal reaches the last seen advertisement.
+//
+// Note that the multihash entries corresponding to the advertisement are synced in the background.
+// The completion of advertisement sync does not necessarily mean that the entries corresponding to
+// the advertisement are synced.
```
MarcoPolo, line 144: 💯 this is good to point out.

PR #143, opened 2021-12-22
Update to new version of go-legs

gammazero:
The new `go-legs` works significantly differently than the previous version.  It does not require explicitly subscribing to peers from which to receive advertisements.  Instead, this is now controlled by the configured discovery policy: advertisements are read and state is kept for any providers that are authorized.  This also means that the `Subscribe` and `Unsubscribe` admin APIs no longer exist.  Instead, there are now Allow and Block to accept or reject provider advertisements.

Syncing can be done to an HTTP provider if there is an explicit sync to an HTTP multiaddress.  The sync command now has an optional flag to specify the multiaddr of the provider to sync to.  This is used for an HTTP sync or a sync where the address is not already known to the indexer through discovery.

The peer that publishes advertisements (publisher) does not need to be the same peer listed as providing content (provider) in the advertisements.  The peer ID that is indexed is what is read from the advertisement, and not the peer with whom the datatransfer session is with.

- Update comments to clarify differences between publisher and provider.
- Allow/trust policy pertains to both publishers and providers.
- Verifying an advertisement signature returns the signer's peer ID so that it can be checked if it is allowed.
- Check that the advertisement is signed by the provider specified in the advertisement.
- Update ingest tests to check that indexed data has the provider ID from the advertisement, not the peer ID of the host that is publishing the advertisement.
- When syncing advertisement links, wait for the previous advertisement to finish syncing.  This prevents a context remove or update advertisement (which does not have entries to sync) from being undone as the previous advertisement finishes syncing.


go.mod
```diff
@@ -7,15 +7,13 @@ require (
 	github.com/filecoin-project/go-address v0.0.5
 	github.com/filecoin-project/go-dagaggregator-unixfs v0.2.0
 	github.com/filecoin-project/go-indexer-core v0.2.6
-	github.com/filecoin-project/go-legs v0.1.0
+	github.com/filecoin-project/go-legs v0.1.1-0.20211222094616-bfa51a79453a
```
willscott, line 10: should we cut a tag?

internal/ingest/interface.go
```diff
@@ -9,11 +9,11 @@ import (
 
 // Ingester is the interface implemented to subscribe to
 // advertisements from a provider
-type Ingester interface {
+type XIngester interface {
```
willscott, line 12: is the X here intentional?

willscott (approved):
great to see how much simpler this gets!

internal/ingest/interface.go
gammazero: Meant to delete that interface.  Done

go.mod
gammazero: Done

gammazero: Yes. Done

internal/ingest/interface.go
gammazero: The X meant it was marked for removal... just need to delete it. Done.

internal/ingest/lockchain.go
```diff
@@ -0,0 +1,78 @@
...
+			delete(lc.linkLocks, curCid)
+			lc.mutex.Unlock()
+		}
+	} else {
+		unlockFunc = func() {}
+	}
+
+	if prevCid != cid.Undef {
+		// Lookup the previous wait channel.
+		prevChan = lc.linkLocks[prevCid]
```
willscott, line 68: can we make this not panic if the specified prefCid isn't in the array?

gammazero: Oops, thanks!  Fixed

api/v0/ingest/schema/envelope.go
```diff
@@ -124,17 +126,23 @@ func VerifyAdvertisement(ad Advertisement) error {
...
 	genID, err := signaturePayload(&previousID, provider, addrs, entries, metadata, isRm)
 	if err != nil {
-		return err
+		return peer.ID(""), err
 	}
 
 	// Consume envelope
 	rec := &advSignatureRecord{}
-	_, err = record.ConsumeTypedEnvelope(sig, rec)
+	env, err := record.ConsumeTypedEnvelope(sig, rec)
```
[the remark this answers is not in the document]
gammazero: Fixed

PR #146, opened 2022-01-04
Update version

gammazero:
Update the version to `v0.2.0`

This updates the version from v0.1 to v0.2 since the Subscribe and Unsubscribe APIs have been removed.  There is no to explicitly subscribe and unsubscribe to separate peers to receive pubsub messages from them.

This version has the following changes:
* Depend on go-legs v0.2.1
* Sync command takes optional multiaddress to sync to.
* Sync command allows explicit sync to HTTP multiaddress.
* New command to reload policy from config file
* Sync advertisements in order so that removal not done before/while removed content is being added.
* Handle advertisements that remove content by context ID 
* Allow publisher and provider to be different peers
* Update comments to clarify differences between publisher and provider.
* Allow/trust policy pertains to both publishers and providers.
* Verifying the advertisement signature returns the signer's peer ID so that it can be checked if it is allowed.
* Check that the advertisement is signed by the provider specified in the advertisement.
* Do not start sync of advertisement that is currently being synced


gammazero:
Do we want to wait to include any more changes before bumping the version?

willscott:
My sense is we're still in a phase of rapid iteration, and shouldn't delay too much

willscott (approved):

PR #148, opened 2022-01-06
Remove by contextID for removal ads that have contextID

gammazero:
If a removal advertisement has a context, then remove content by contextID, regardless of whether the add includes an entries link or not.  Individual entries will only be removed if an empty context ID is given.

Other changes:
- Added removal test to end-to-end test.
- Do not output raw binary contextID from find command.

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -3,6 +3,7 @@ package ingest
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


internal/ingest/linksystem.go
```diff
@@ -283,42 +284,55 @@ func (ing *Ingester) syncAdEntries(from peer.ID, ad schema.Advertisement, adCid
...
+		err = ing.indexer.RemoveProviderContext(providerID, contextID)
 		if err != nil {
-			log.Error("Failed to removed content by context ID", "contextID", contextID)
+			log.Error("Failed to removed content by context ID")
 			return
 		}
-		log.Infow("Removed content by context ID", "contextID", contextID)
+		log.Infow("Removed content by context ID")
+		return
+	}
```
willscott, line 323: should we warn if isRm is set and there is no contextID?

willscott (approved):

internal/ingest/linksystem.go
gammazero: The is literally done on the next line (see `for_removal`).  If `isRm` is set and there is a context ID, execution does not get this far.

PR #150, opened 2022-01-10
Update to latest datastore with contexts

gammazero:
Also updates other dependencies to latest, including libp2p.


internal/ingest/ingest.go
```diff
@@ -161,7 +161,7 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
 func (ing *Ingester) watchSyncFinished(onSyncFin <-chan legs.SyncFinished) {
 	for syncFin := range onSyncFin {
 		// Persist the latest sync
-		err := ing.ds.Put(datastore.NewKey(syncPrefix+syncFin.PeerID.String()), syncFin.Cid.Bytes())
+		err := ing.ds.Put(context.Background(), datastore.NewKey(syncPrefix+syncFin.PeerID.String()), syncFin.Cid.Bytes())
```
willscott, line 164: do we want a timeout on these operations?

internal/registry/registry.go
```diff
@@ -102,7 +102,7 @@ func NewRegistry(cfg config.Discovery, dstore datastore.Datastore, disco discove
 		dstore: dstore,
 	}
 
-	count, err := r.loadPersistedProviders()
+	count, err := r.loadPersistedProviders(context.Background())
```
willscott, line 105: should this one bubble up?

internal/registry/registry.go
```diff
@@ -402,37 +402,36 @@ func (r *Registry) syncStartDiscover(peerID peer.ID, discoAddr string, errCh cha
...
+			if discoErr != nil {
+				errCh <- discoErr
+				return
+			}
+
+			info := &ProviderInfo{
+				AddrInfo:      discoData.AddrInfo,
+				DiscoveryAddr: discoAddr,
+			}
+			if err := r.syncRegister(context.Background(), info); err != nil {
```
willscott, line 419: might be good also for this to have a timeout?

internal/ingest/ingest.go
gammazero: Not for this one.  There is nothing waiting for this to complete, so there is no reason to consider a slow write as too long and need to cancel it.

internal/registry/registry.go
gammazero: This should use the configured discovery timeout.  Fixed.

internal/registry/registry.go
gammazero: Probably makes sense, but only to cancel this if a graceful shutdown is requested while the indexer is still starting up.
Done.

willscott (approved):

PR #151, opened 2022-01-13
Fix incorrect log messages

gammazero:
Cache advertisement while processing entry chunks for advertisement to prevent reloading re-validating the same data.


willscott (approved):

willscott:
it looks like the other thing to look like is in lotus integration what's happening is we're seeing config passed in as an empty/newly-initialized config struct. We need those values, when 0, to get interpreted as the default values.

PR #164, opened 2022-01-21
Side effect import to register the dagjson codec

MarcoPolo:
The previous change removed the dagjson import which unregistered the codec.


willscott:
lets add dag cbor as well while we're at it. that's the other likely format to support

willscott (approved):

PR #166, opened 2022-01-22
Process content entries in earliest to latest order

gammazero:
When handling a chain of entries, it was still possible for content entries to be processed out of order if a previous ad took too long to show up.  This add a timer to wait for a previous ad if there is one that has not already been handled by the indexer.

When import an advertisement chain that looks like this:
```
                    HEAD
                    |
                    V
Ad-A <--- Ad-B <--- Ad-C
|         |         |
V         V         V
Ents-a    Ents-b    Ents-c
```
The sequence of events is  as follows, with content syncs started in the wrong order:
```
StorageWriteOpener: Recv Ad C
StorageHook: go syncAdEntries(C)

StorageWriteOpener: Recv Ad B
StorageHook: go syncAdEntries(B)

StorageWriteOpener: Recv Ad A
StorageHook: go syncAdEntries(A)

syncAdEntries(C): StartSync Ents-c
syncAdEntries(B): StartSync Ents-b
syncAdEntries(A): StartSync Ents-a
```
This PR corrects this so that the content syncs are ordered:
```
syncAdEntries(A): StartSync Ents-a
syncAdEntries(B): StartSync Ents-b
syncAdEntries(C): StartSync Ents-c
```


internal/ingest/linksystem.go
```diff
@@ -289,9 +297,23 @@ func (ing *Ingester) storageHook(pubID peer.ID, c cid.Cid) {
 }
 
 // syncAdEntries fetches all the entries for a single advertisement
-func (ing *Ingester) syncAdEntries(from peer.ID, ad schema.Advertisement, adCid cid.Cid, prevWait <-chan struct{}, unlock context.CancelFunc) {
-	<-prevWait
-	defer unlock()
+func (ing *Ingester) syncAdEntries(from peer.ID, ad schema.Advertisement, adCid, prevCid cid.Cid) {
+	if !ing.adWaiter.wait(prevCid) {
+		// If there was a previous ad, but it was not waitiable, then check if
```
MarcoPolo, line 302: I don't love this. Is there some other solution we can come up with that is better?

gammazero: I think I can do better by building a linked list of ads as they arrive.  The indexer can then wait until all the ads in the chain are received (when the initial ad with no previous link is received), or until the chain links to a previously processed ad.

masih: Am I right to say without this PR the index processing is broken on the indexer node? If so, It would be great to push this out sooner rather than later if you folks think the implementation is correct and improve later.

masih (approved):

internal/ingest/linksystem.go
gammazero: @masih Yes, it is effectively broken without this.  Will merge and we can provide improvements in following PR

PR #167, opened 2022-01-24
Limit the depth of recursion when syncing advertisement entries

masih:
Add a configuration value that allows limiting the recursion depth when
syncing the advertisement entries with default value of 400. The default
of 400 in conjunction with default chunk size of 16384 in index-provider
would accommodate a default total of 6,553,600 multihashes, which seems
like a reasonably large number.

### DIFF
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -1,6 +1,10 @@
 package config
 
-import "time"
+import (
+	"time"
+
+	"github.com/ipld/go-ipld-prime/traversal/selector"
+)
 
 // Ingest tracks the configuration related to the ingestion protocol.
 type Ingest struct {
@@ -16,14 +20,19 @@ type Ingest struct {
 	// or a chain of advertisement entries.  The value is an integer string
 	// ending in "s", "m", "h" for seconds. minutes, hours.
 	SyncTimeout Duration
+
+	// The recursion depth limit when syncing advertisement entries.
+	// The value -1 means no limit. Defaults to 400.
+	EntriesDepthLimit int64
 }
 
 // NewIngest returns Ingest with values set to their defaults.
 func NewIngest() Ingest {
 	return Ingest{
-		PubSubTopic:    "/indexer/ingest/mainnet",
-		StoreBatchSize: 256,
-		SyncTimeout:    Duration(2 * time.Hour),
+		PubSubTopic:       "/indexer/ingest/mainnet",
+		StoreBatchSize:    256,
+		SyncTimeout:       Duration(2 * time.Hour),
+		EntriesDepthLimit: 400,
 	}
 }
 
@@ -40,4 +49,16 @@ func (c *Ingest) populateUnset() {
 	if c.SyncTimeout == 0 {
 		c.SyncTimeout = def.SyncTimeout
 	}
+	if c.EntriesDepthLimit == 0 {
+		c.EntriesDepthLimit = def.EntriesDepthLimit
+	}
+}
+
+// EntriesRecursionLimit returns the recursion limit of advertisement entries.
+// See Ingest.EntriesDepthLimit
+func (c *Ingest) EntriesRecursionLimit() selector.RecursionLimit {
+	if c.EntriesDepthLimit == -1 {
+		return selector.RecursionLimitNone()
+	}
+	return selector.RecursionLimitDepth(c.EntriesDepthLimit)
 }
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/query"
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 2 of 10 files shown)


internal/ingest/ingest.go
```diff
@@ -70,6 +73,14 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr *indexer.Engine, reg *regi
 			efsb.Insert("PreviousID", ssb.ExploreRecursiveEdge())
 		}).Node()
 
+	// Construct the selector used when syncing entries of an advertisement with the configured
+	// recursion limit.
+	entSel := ssb.ExploreRecursive(cfg.EntriesRecursionLimit(),
+		ssb.ExploreFields(func(efsb builder.ExploreFieldsSpecBuilder) {
+			efsb.Insert("Next", ssb.ExploreRecursiveEdge())    // Next field in EntryChunk
+			efsb.Insert("Entries", ssb.ExploreRecursiveEdge()) // Entries field in advertisement
```
MarcoPolo, line 81: Why do we need to expore the `Entries` field? Aren't we already saying we are in an entrychunk, so we wouldn't expect to run into an advertisement.

MarcoPolo (commented):
One question. Can we also add a test here?

internal/ingest/ingest.go
masih: Only added so that this works even if given CID of an advertisement. Happy to remove if you think that's better.

MarcoPolo: Yes please. I think it's a bit simpler to reason about and that way it fails if someone sends us an entry chunk that links to an advertisement (it would be weird if that worked!).

config/ingest.go
```diff
@@ -16,14 +20,19 @@ type Ingest struct {
...
 // NewIngest returns Ingest with values set to their defaults.
 func NewIngest() Ingest {
 	return Ingest{
-		PubSubTopic:    "/indexer/ingest/mainnet",
-		StoreBatchSize: 256,
-		SyncTimeout:    Duration(2 * time.Hour),
+		PubSubTopic:                "/indexer/ingest/mainnet",
+		StoreBatchSize:             256,
+		SyncTimeout:                Duration(2 * time.Hour),
+		EntriesRecursionLimitDepth: 400,
```
gammazero, line 35: Should this be called `EntriesRecursionLimit` or `EntriesDepthLimit`?  Seems like Recursion and Depth or redundant terms.

config/ingest.go
```diff
@@ -41,3 +50,12 @@ func (c *Ingest) populateUnset() {
...
 	}
 }
+
+// EntriesRecursionLimit returns the recursion limit of advertisement entries.
+// See Ingest.EntriesRecursionLimitDepth
+func (c *Ingest) EntriesRecursionLimit() selector.RecursionLimit {
+	if c.EntriesRecursionLimitDepth == 0 {
+		return selector.RecursionLimitNone()
+	}
+	return selector.RecursionLimitDepth(c.EntriesRecursionLimitDepth)
```
gammazero, line 60: Another option is to use `0` as a default value, which will get transformed into an actual default by `populateUnset` above, and then use `-1` to mean unlimited.

gammazero (approved):

masih:
Failing windows tests may be resolved by #166. Awaiting merge of that PR before marking as ready for review.

internal/ingest/ingest_test.go
```diff
@@ -124,6 +130,51 @@ func TestSync(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, lcid, c1)
+	case <-ctx.Done():
+		t.Fatal("sync timeout")
+	}
+
+	// Assert the maximum number of ingestable multihashes from original published ad are indexed.
+	checkMhsIndexedEventually(t, i.indexer, providerID, allMhs[maxIngestableMhCount:])
+
+	// Assert any remaining published multihashes are no indexed.
```
gammazero, line 170:
```suggestion
	// Assert any remaining published multihashes are not indexed.
```

internal/ingest/ingest_test.go
```diff
@@ -124,6 +130,51 @@ func TestSync(t *testing.T) {
...
+	for _, mh := range allMhs[:maxIngestableMhCount] {
```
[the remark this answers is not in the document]
masih: See line 168 where the second half is asserted for indexing. 

Here we check the first half is not indexed. The rationale for reverse order is that the tailing multihashes are added to earlier chunks and traversal should suspend traversing deeper chunks, i.e. first half of the slice.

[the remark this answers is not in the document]
gammazero: Seems like line 168 should be:
```go
remainingMhsCount := len(allMhs) - maxIngestableMhCount
checkMhsIndexedEventually(t, i.indexer, providerID, allMhs[remainingMhsCount:])
```
and line 171 should be:
```go
for _, mh := range allMhs[:remainingMhsCount] {
```

PR #169, opened 2022-01-25
Fix error check is ad cache

### LINKED ISSUE #168: `loadAdData` does not use cache when indexing block data
`hasNext` is always false when [mapping of entry CID to advertisement](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/linksystem.go#L449) made which means the advertisement data is never cached while ad entries are being processed [here](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/linksystem.go#L453).

https://github.com/filecoin-project/storetheindex/blob/bc23aad609943fc1ca6a4fe5ba53bef4f6bc1089/internal/ingest/linksystem.go#L512

gammazero:
Code error was preventing ad caching.

Fixes #168 

🤦

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -500,7 +500,7 @@ func (ing *Ingester) setNextCidToAd(nchunk schema.EntryChunk, adCid cid.Cid) (bo
 
 	lnk, err := nchunk.Next.AsNode().AsLink()
 	if err != nil {
-		return false, fmt.Errorf("error decoding next chunk link: %s", err)
+		return false, fmt.Errorf("error decoding next chunk link: %w", err)
 	}
 	err = pushCidToAdMapping(context.Background(), ing.ds, lnk.(cidlink.Link).Cid, adCid)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/ingest/linksystem.go
```diff
@@ -494,16 +494,17 @@ func (ing *Ingester) indexContentBlock(adCid cid.Cid, pubID peer.ID, nentries ip
...
 	lnk, err := nchunk.Next.AsNode().AsLink()
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("error decoding next chunk link: %s", err)
 	}
 	err = pushCidToAdMapping(context.Background(), ing.ds, lnk.(cidlink.Link).Cid, adCid)
-	if err == nil {
-		return false, err
+	if err != nil {
+		return false, fmt.Errorf("error storing reverse map for next chunk: %s", err)
```
masih, line 507:
```suggestion
		return false, fmt.Errorf("error storing reverse map for next chunk: %w", err)
```

masih (approved):

PR #171, opened 2022-01-26
Add command to list provider information

gammazero:
List current information about registered providers.  Example:
```
./storetheindex ingest providers --indexer 18.116.210.184

Provider 12D3KooWE8yt84RVwW3sFcd6WMjbUdWrZer2YtT4dmtj3dHdahSZ
    Addresses: [/ip4/192.168.0.109/tcp/45399]
    LastAdvertisement: baguqeeqql7kdkkburx32mv6nc7uhsqdfjy
    LastAdvertisementTime: 2022-01-24T16:31:56Z
Provider 12D3KooWNUC8jEwuwH464tCEDfk1Y8SdaUZ7mJgjBU9NfjpBrghq
    Addresses: [/ip4/3.143.149.162/tcp/3103]
    LastAdvertisement: baguqeeqqufvyqiuqyfwtguwo2tg2o66myi
    LastAdvertisementTime: 2022-01-25T19:14:16Z
```

### DIFF
--- a/command/ingest.go
+++ b/command/ingest.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	httpclient "github.com/filecoin-project/storetheindex/api/v0/admin/client/http"
+	ingclient "github.com/filecoin-project/storetheindex/api/v0/ingest/client/http"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/multiformats/go-multiaddr"
 	"github.com/urfave/cli/v2"
@@ -37,6 +38,13 @@ var reload = &cli.Command{
 	Action: reloadPolicyCmd,
 }
 
+var providers = &cli.Command{
+	Name:   "providers",
+	Usage:  "Shown information about known providers",
+	Flags:  ingestProvidersFlags,
+	Action: providersCmd,
+}
+
 var IngestCmd = &cli.Command{
 	Name:  "ingest",
 	Usage: "Admin commands to sync indexer with a provider",
@@ -45,6 +53,7 @@ var IngestCmd = &cli.Command{
 		allow,
 		block,
 		reload,
+		providers,
 	},
 }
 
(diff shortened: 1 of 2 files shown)


command/ingest.go
```diff
@@ -37,6 +38,13 @@ var reload = &cli.Command{
 	Action: reloadPolicyCmd,
 }
 
+var providers = &cli.Command{
+	Name:   "providers",
+	Usage:  "Shown information about known providers",
```
willscott, line 43:
```suggestion
	Usage:  "Show information about known providers",
```

command/ingest.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	httpclient "github.com/filecoin-project/storetheindex/api/v0/admin/client/http"
+	ingclient "github.com/filecoin-project/storetheindex/api/v0/ingest/client/http"
```
willscott, line 7: why is this an ingest rather than an admin api action?
It seems like this is a listing that the operator of the storetheindex node would be able to do, and so would be exposed on the admin API. what's the distinction?

gammazero: I agree.  That was where it was for some time, probably never got moved when things were reorganized.  I will move it.  As far as I can tell it was not previously used anywhere so moving it will not affect anything.

gammazero: After giving it more thought, I think we want `ListProviders` and `GetProvider` to be part of the public ingest API.  That allows providers/publishers to confirm that their information is up-to-date at any indexer, and allows anyone to see if an indexer is indexing content for a particular provider.

willscott: i think 1 level of abstraction meta from that:
when we say ‘the ingest api’ it would be great if we mean 1 specific thing:
the protocol for publishers to create and submit their chain of advertisements.

that we also have this http thing that’s used for interactive querying about the state of advertisements, but isn’t part of the ingest protocol overloads that term and is confusing.

This is not part of ‘the ingest api’ because it’s not available if i’m doing gossip sub + graphsync for transmission.

can we name it something else? ‘publisher status api’?

gammazero: There are three different services and each is available on a different port and potentially a different network interface:
- admin: 127.0.0.1:3002
- finder: 0.0.0.0:3000
- ingest: 0.0.0.0:3001

I was considering an API to be paired with each service. Finder and ingest were on different servers because of an anticipation that someone running an indexer for themselves may not want to expose the ingest service, but only the finder service, although these are combined in the libp2p service.

It is looking like I probably want to combine finder and ingest into one service, and then add the publisher status API to that service. 

- admin: 127.0.0.1:3002
- http: 0.0.0.0:3001

WDYT?

willscott: * the HTTP should stay on :3000 as that's what infra just did a bunch of work to route differently
* I like the  http vs admin naming (or maybe 'public' vs 'admin' as the two services?) since these APIs are different from the ingest (and soon find) interfaces that we expect others to implement directly.

gammazero: I think "admin" and "indexer" make sense.  Indexer is served over http and libp2p.

gammazero:
replaced by #186

PR #176, opened 2022-01-26
Limit the recursion depth when syncing advertisements

masih:
By default sync at most 400 unseed advertisements and expose ingest
config parameter to override it.


willscott:
if we get more than 400 behind the head, will this cause them all to get synced but over multiple sessions, ore will we end up loosing the old ones?

masih:
The latter

gammazero (approved):

gammazero (changes requested):
Need to fix test

gammazero (approved):

PR #177, opened 2022-01-27
New Ingester proposal

MarcoPolo:
Here's a proposal for a new way to ingest data in the indexer. Please take a look and give me your thoughts


doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+7. The staging area is represented as a list of stacks of Ads that need to be
+   processed. One staging area per provider.
+8. To make progress on the staging area a worker:
+   1. Picks the stack at the head of the list (the first thing put in the
+      staging area).
+   2. In the stack, pop the stack (last ad put on the stack) and process it
+     (by fetching the ad's referenced entries and indexing them). Repeat until
+     stack is empty.
+   3. Remove the head from the staging area.
+   4. Repeat 1 -> 3 with the rest of the list.
```
masih, line 68: * Maybe document an unhappy path and how this proposal would resume partially "staged" entries.
* cover a case where syncing of entries fails such that some of the entries are written into the datastore but not all.

masih (commented):
Would love to see more details. 

Great start nevertheless.

MarcoPolo:
What should happen when the ad chain is very long and the syncer does not sync the whole chain?
Similarly what should happen if the entry chunk chain is too long and the syncer does not sync the whole entry chunk?

We can do a couple things:
1. Pause our work and resume when this provider has more "gas" (equivalently when their rate tokens replenish). This means a new indexer will eventually get up to parity to an established indexer (assuming the absence of multiple provider chains).
2. Drop the ads not processed. (I don't think we should do this since it conflicts with constraint 1.)

doc/ingest-proposal.md
MarcoPolo: Could you elaborate on what you mean by _partially "staged" entries_. Is it bullet point 2 or something else?

doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+1. If a an Ad is applied all older ads this ad references have also been
+   applied. For example, given some chain `A <- B <- C`, the indexer will never
+   be in the state that it indexed `A & C` but not `B`.
+2. An indexer will index an Ad chain, but will not make any gaurantees about
+   consistency in the presence of multiple ad chains for a given provider. For
+   example if a provider publishes two ad chains at the same time chain1 and
+   chain2 the indexer will apply whichever chain it learns about first first,
+   then apply the other chain.
+3. An indexer will not index the same Ad twice.
+4. An indexer will be resilient to restarts. If the indexer goes down and comes
```
masih, line 39: Should the following also be a constraint:
```
Given an entries chain of length lower than the maximum tolerable by the indexer, the indexer must guarantee that all its chunks are eventually ingested.
```

MarcoPolo: What is the maximum tolerable length? Is this globally or for a sync session?

doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+The indexer semantics are relatively simple, yet the ingester may fail in
+surprising and hard to reason about ways. The goal of this proposal is to
+outline a design for an alternate ingester that upholds the indexer semantics
+while being easy to reason about.
+
+## Overview of the network
+
+The indexer ingests data from publishers. There can be one publisher per
+provider or many publishers per provider. Additionally there can be faulty or
+malicious publishers that can point to other chains.
```
gammazero, line 20: Does this account for one publisher for many providers?

doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+   then apply the other chain.
+3. An indexer will not index the same Ad twice.
+4. An indexer will be resilient to restarts. If the indexer goes down and comes
+   back up it should not break constraint 1.
+
+## Proposal
+
+1. Introduce a new concept called "staging". An advertisement chain is put into
+   the staging area after it is synced but before the chain is indexed.
+2. Each *provider* has its own staging area.
```
gammazero, line 46: What happens if there are multiple publishers concurrently syncing the same chain for the same provider?

doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+3. The staging area is persisted.
+4. Advertisements are processed serially by a single worker per provider.
+5. A handler, which represents a single publisher, will write to an intermediate
+   in memory stack (via `StorageWriteOpener`[0]) as it discovers ads from a sync.
+   * If a chain contains multiple providers, each provider will have it's own
```
gammazero, line 51: I am not sure we want to allow different providers in the same chain of ads.  It would mean that different ads were signed by different keys and that some ads in the chain would not be allowed while others would be.

doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+     stack.[1]
+   * If the ad has already been processed (by checking the datastore), then we
+     don't add it to the stack and stop traversing to earlier ads (because of
+     constraint 1).
+6. When the handler's sync is finished it moves the stack(s)[2] into each
+   provider's staging area _before_ updating the latestSync state. It will also
+   trigger a worker to start processing ads for the provider(s).
```
gammazero, line 58: When the handler finishes processing the chain of ads into a stack of ads, could it then become the worker to process each of the ads in the stack?

doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+4. An indexer will be resilient to restarts. If the indexer goes down and comes
+   back up it should not break constraint 1.
+
+## Proposal
+
+1. Introduce a new concept called "staging". An advertisement chain is put into
+   the staging area after it is synced but before the chain is indexed.
+2. Each *provider* has its own staging area.
+3. The staging area is persisted.
+4. Advertisements are processed serially by a single worker per provider.
```
gammazero, line 48:
```suggestion
4. After the **handler** has finished processing a chain of advertisements into a per-provider stack, the advertisements are then processed serially by a single **worker**, one worker per provider/stack.
```
I think this should come after the step currently labeled as 5.

gammazero (commented):
I think the is conceptually a cleaner model than what we currently have.  It would be good to demonstrate a failure mode that the current indexer is susceptible to that this design avoids.  Especially if it helps us prioritize development. 

This means that we no longer need to use the block hook at all, is that correct?  If so, do we want to keep it in go-legs?

doc/ingest-proposal.md
```diff
@@ -0,0 +1,102 @@
+# Proposal for a new Ingest design
+
+## Current Problems
+
+1. Not resilient to process death.
+2. Half the logic is in storage writer the other half in a block hook.
+3. Undefined behavior if multiple publishers sync the same chain.
```
gammazero, line 7: It seems to me that would be very rare as it would mean multiple publishers are publishing for the same provider.

doc/ingest-proposal.md
MarcoPolo: A publisher only publishes a single chain. Are you saying that there may be multiple providers within a single chain?

MarcoPolo: Yes. If by this you mean that one publisher is managing one chain and that chain may have many providers.

A publisher (as it is designed right now) can not have multiple chains. There is a 1:1 mapping of publisher to chain. Of course a publisher may flip between chains but it would only advertise a single chain in its head.

doc/ingest-proposal.md
MarcoPolo: See the bottom half of the ingesterStaging.jpg image.
and:
> When the handler's sync is finished it moves the stack(s)[2] into each provider's staging area.

so it may put the ads that need to be processed in the staging area. And those ads may already be in the staging area, but that's okay it will just be a no-op when we process the ad.

doc/ingest-proposal.md
MarcoPolo: I agree we should not condone it. But I'm not sure how we would prohibit it.

doc/ingest-proposal.md
MarcoPolo: maybe? but I'm not sure the benefit would outweigh the complexity.

I imagined something like this:
StorageWriteOpener builds the in-memory stack.
when we get a onsyncfinished we move that in-memory stack to the staging area for the providers.
then trigger a worker to start for each provider.

You can imagine this handler is a single goroutine for everyone that watches the onsyncfinished chan and moves stacks to staging.

MarcoPolo:
Just a note, I don’t think we actually need to persist the staging area if we do something similar to what we’re currently doing where we don’t update the latestSync state until after we’ve processed an Ad.

MarcoPolo:
For some quicker changes.

Here are the issues I outlined earlier:
```
1. Not resilient to process death.
2. Half the logic is in storage writer the other half in a block hook.
3. Undefined behavior if multiple publishers sync the same chain.
4. Relies of a chain of blocking channels to reverse the order of graphsync.
5. not linear execution flow.
```
We can do some short term fixes for some of these:

We fixed, 1 and mostly 2.
For 3 we can take a lock for a provider so multiple publisher syncs don’t step on each other.
4 is not really changeable unless we adopt something like this.
5 we can do more here if we adopt https://github.com/filecoin-project/go-legs/pull/79

MarcoPolo:
Closing this since we merge #224 . I'll make a new PR with the latest semantics. Especially around resyncs, since that needs to be clarified.

PR #180, opened 2022-01-28
Process all ads in the presence of restarts

MarcoPolo:
Most of the diff is test related.

## Context

The indexer will currently drop advertisements if it restarts in certain critical parts of the flow. If you sync a chain A<-B<-C, you will set latestSync to C while still process A, B, and C. If the indexer restarts before finishing processing then it will never try to process those Ads again.

## Proposed change

The change here is relatively simple and relies on the fact that if the same advertisement is processed twice back to back there is no net effect.

The change is to not persist the latestSync of an advertisement until _after_ we know we've fully processed it. This way if the indexer restarts it sets the last fully processed ad as the `latestSync` and go-legs will call the storage hook on the ads we haven't processed yet.

As part of this we also move some logic (entryChunkCid -> adCid mapping) out of the `storageWriteOpener` since that only gets called when we first get the block and we're putting it somewhere.


MarcoPolo:
Tests failing because of https://github.com/filecoin-project/storetheindex/pull/181

gammazero:
Tests are failing due to import cycle

internal/ingest/linksystem.go
```diff
@@ -258,6 +253,7 @@ func (ing *Ingester) storageHook(pubID peer.ID, c cid.Cid) {
 			lnk, err := ad.FieldPreviousID().Must().AsLink()
 			if err != nil {
 				log.Errorw("Cannot read previous link from advertisement", "err", err)
+				return
```
gammazero, line 256: The return was intentionally absent so that if unable to read the previous link, we would still process this ad.  If you agree, then remove return.  If you want to skip processing this ad, then remove the following `else`.

internal/ingest/linksystem.go
```diff
@@ -451,6 +447,9 @@ func (ing *Ingester) indexContentBlock(adCid cid.Cid, pubID peer.ID, nentries ip
 	//
 	// Do not return here if error; try to index content in this chunk.
 	hasNextLink, linkErr := ing.setNextCidToAd(nchunk, adCid)
+	if linkErr != nil {
+		return linkErr
+	}
```
gammazero, line 452: See comment above:
```go
// Do not return here if error; try to index content in this chunk.
```

This was done to process the current entry, even if the link to the next entry could not be read.  Either remove the code that returns, or update the comment.

gammazero (changes requested):

internal/ingest/linksystem.go
MarcoPolo: I want to skip processing this ad. I think it's okay to not process an ad that is malformed and if this has a field for previous ID *and* it fails to parse as a link then it is malformed.

MarcoPolo:
Test failing because of: https://github.com/filecoin-project/go-legs/pull/75

internal/ingest/linksystem.go
```diff
@@ -449,8 +457,12 @@ func (ing *Ingester) indexContentBlock(adCid cid.Cid, pubID peer.ID, nentries ip
 	// entries CID to the ad CID so that the ad can be loaded when that chunk
 	// is received.
 	//
-	// Do not return here if error; try to index content in this chunk.
+	// If we error in reading the next link, then that means this entry chunk is
+	// malformed, so it is okay to skip it.
 	hasNextLink, linkErr := ing.setNextCidToAd(nchunk, adCid)
+	if linkErr != nil {
+		return linkErr
```
gammazero, line 464:
```suggestion
	hasNextLink, err := ing.setNextCidToAd(nchunk, adCid)
	if err != nil {
		return err
```
OK, then just use `err` instead of `linkErr` since it is not being handled specially.

gammazero (approved):
Approve, with minor comment.

gammazero (approved):

PR #187, opened 2022-01-28
Make FindLatency metric per multihash

gammazero:
Otherwise, we see big latency times for large batch finds.


MarcoPolo (approved):
We probably still want some measurements around how long a request took as well.

PR #190, opened 2022-01-29
Mark an advertisement as processed when the is has no entries

gammazero:
When processing an ad with no entries (Entries set to `NoEntries`), the ad was not being marked as processed.  This fixes that so that those ads are marked as processed. If the ad is not readable then it will also be marked as processed so that the indexer does not get stuck trying to reprocess the same unreadable ad.

This can be tested by removing a car file from the provider to create a delete advertisement, and syncing the indexer to the provider. Then restart the indexer and sync to the provider again, repeat.  Without this fix, the indexer fetches and reprocesses the ad each time.  With this fix, the indexer recognizes that the delete advertisement has already been processed.

- Update go-legs and libp2p


willscott (approved):

MarcoPolo:
Can we add a test too? I don’t see how if the entries is NoEntries it fails. 

Also I’m a bit worried the is too eager to skip processing an entry, but I guess in the other cases besides syncing sub entries you aren’t relying on flaky things like the network.

MarcoPolo:
>  I don’t see how if the entries is NoEntries it fails.

Oh we had an early return for rm ads. I missed that when I thought everything reached the .Sync line 😞

MarcoPolo:
dfcb98e675856a41ec6b434aae86346584ae6812 has a test, feel free to cherry pick

PR #192, opened 2022-01-30
Add compatibility with old signature format

gammazero:
There are still many advertisements that are using the deprecated signature format.  Since these ads are still valid, allow the indexer to process them.  Log when an ad with a deprecated signature format is seen.

This will allow the new indexer to be deployed without requiring an update to any data providers/publishers.

Other changes:
- Removed some unwanted/unneeded logging from finder and admin client
- Update version

### DIFF
--- a/api/v0/admin/client/http/client.go
+++ b/api/v0/admin/client/http/client.go
@@ -13,12 +13,9 @@ import (
 	"path"
 
 	"github.com/filecoin-project/storetheindex/api/v0/httpclient"
-	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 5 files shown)


willscott:
I'm i guess okay if you want to do this, or we can just only support the fixed one and ignore the old advertisements from the one provider that has created them with bad signatures

willscott (approved):

gammazero:
I would have just ignored the ads with old signature if sofia miner had not done its migration already.  We will be unable to sink some of the new content without this, and we may want to do a resync now that everything is fixed.  We can probably get rid of this in a few weeks.

willscott (approved):

PR #196, opened 2022-02-01
Cleanup entry to ad reverse mapping when skipping ads

gammazero:
When skipping ads, the initial reverse mapping of entry cid to ad cid needs to be deleted.  Also, when deleting by context ID the ad will have no entries, so a reverse mapping should not be created.

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -368,7 +368,7 @@ func (ing *Ingester) syncAdEntries(from peer.ID, ad schema.Advertisement, adCid,
 			dk := dsKey(admapPrefix + entCid.String())
 			err = ing.ds.Delete(context.Background(), dk)
 			if err != nil {
-				log.Errorw("cannot delete advertisement cid for entries cid from datastore: %w", err)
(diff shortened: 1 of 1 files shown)


internal/ingest/linksystem.go
```diff
@@ -361,6 +363,14 @@ func (ing *Ingester) syncAdEntries(from peer.ID, ad schema.Advertisement, adCid,
 	// This ad has bad data or has all of its content deleted by a subsequent
 	// ad, so skip any further processing.
 	if skip {
+		if elink != schema.NoEntries {
+			entCid := elink.(cidlink.Link).Cid
+			dk := dsKey(admapPrefix + entCid.String())
+			err = ing.ds.Delete(context.Background(), dk)
+			if err != nil {
+				log.Errorw("cannot delete advertisement cid for entries cid from datastore: %w", err)
```
willscott, line 371:
```suggestion
				log.Errorw("cannot delete advertisement cid for entries cid from datastore", "err", err)
```

willscott (approved):

PR #201, opened 2022-02-04
Add handler for announcing new heads on the ingest server

willscott:
see if this is roughly the shape of an announce handler you'd hope for.

### DIFF
--- a/server/ingest/handler/ingest_handler.go
+++ b/server/ingest/handler/ingest_handler.go
@@ -135,8 +135,8 @@ func (h *IngestHandler) Announce(ctx context.Context, data io.Reader) error {
 	if err != nil {
 		return err
 	}
-	allow, err := h.registry.Authorized(pid)
-	if err != nil {                                                                                                                                                      
+	allow, err := h.registry.Authorized(pid.ID)
+	if err != nil {
 		err = fmt.Errorf("error checking if peer allowed: %w", err)
 		return v0.NewError(err, http.StatusInternalServerError)
 	}
--- a/server/ingest/handler/ingest_handler.go
+++ b/server/ingest/handler/ingest_handler.go
@@ -135,7 +135,14 @@ func (h *IngestHandler) Announce(ctx context.Context, data io.Reader) error {
 	if err != nil {
 		return err
 	}
-
+	allow, err := h.registry.Authorized(pid)
+	if err != nil {                                                                                                                                                      
+		err = fmt.Errorf("error checking if peer allowed: %w", err)
+		return v0.NewError(err, http.StatusInternalServerError)
+	}
+	if !allow {
+		return v0.NewError(errors.New("not authorized to announce"), http.StatusForbidden)
+	}
 	cur, err := h.ingester.GetLatestSync(pid.ID)
 	if err == nil {
 		if cur.Equals(an.Cid) {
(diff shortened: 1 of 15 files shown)


server/ingest/handler/ingest_handler.go
```diff
@@ -107,3 +113,36 @@ func (h *IngestHandler) IndexContent(ctx context.Context, data []byte) error {
...
+	if len(an.Addrs) > 1 {
+		return fmt.Errorf("must specify 1 location to fetch on direct announcments")
+	}
+	// todo: require auth?
+
+	pid, err := peer.AddrInfoFromP2pAddr(an.Addrs[0])
+	if err != nil {
+		return err
+	}
+
```
gammazero, line 138:
```suggestion
	allow, err := h.registry.Authorized(pid)
	if err != nil {                                                                                                                                                      
		err = fmt.Errorf("error checking if peer allowed: %w", err)
		return v0.NewError(err, http.StatusInternalServerError)
	}
	if !allow {
		return v0.NewError(errors.New("not authorized to announce"), http.StatusForbidden)
	}
```
This includes the HTTP status with the error message.

server/ingest/http/handler.go
```diff
@@ -96,42 +97,16 @@ func (h *httpHandler) removeProvider(w http.ResponseWriter, r *http.Request) {
...
 
-// POST /ingest/content
-func (h *httpHandler) indexContent(w http.ResponseWriter, r *http.Request) {
-	body, err := io.ReadAll(r.Body)
+	defer r.Body.Close()
+	err := h.ingestHandler.Announce(r.Context(), r.Body)
 	if err != nil {
-		log.Errorw("failed reading body", "err", err)
+		log.Errorw("Advertise failed:", "err", err)
 		http.Error(w, "", http.StatusInternalServerError)
```
gammazero, line 108:
```suggestion
		httpserver.HandleError(w, err, "announce") 
```
This will handle the correct HTTP status for the error.

server/ingest/handler/ingest_handler.go
```diff
@@ -19,12 +23,14 @@ import (
 // that is common to all protocols
 type IngestHandler struct {
 	indexer  indexer.Interface
+	ingester *ingest.Ingester
 	registry *registry.Registry
 }
 
-func NewIngestHandler(indexer indexer.Interface, registry *registry.Registry) *IngestHandler {
+func NewIngestHandler(indexer indexer.Interface, ingester *ingest.Ingester, registry *registry.Registry) *IngestHandler {
```
gammazero, line 30: If you want to get rid of `IndexContent` then the `indexer` argument is not needed.

gammazero (approved):
After fixing the test failures (Lint check and `TestRegisterProvider` failing)...
Looks good, with a couple of suggestions.

server/ingest/http/protocol_test.go
```diff
@@ -70,3 +75,24 @@ func TestRegisterProvider(t *testing.T) {
...
+	}
+	test.RegisterProviderTest(t, httpClient, peerID, privKey, "/ip4/127.0.0.1/tcp/9999", reg)
+
+	ai, _ := peer.AddrInfoFromString("/ip4/127.0.0.1/tcp/9999/")
+	ai.ID = peerID
+
+	if err := httpClient.Announce(context.Background(), ai, cid.Undef); err != nil {
+		t.Fatal(err)
+	}
+}
```
gammazero, line 98: I think this should be moved into `/server/ingest/test/test.go` as `AnnounceTest(...)` and then called from both `/server/ingest/http/protocol_test.go` and `/server/ingest/libp2p/protocol_test.go`

server/ingest/handler/ingest_handler.go
willscott: doesn't hurt to keep it piped through to this server

willscott:
okay. test passes and I think i'm okay with this.

PR #206, opened 2022-02-08
A couple of metric tweaks

MarcoPolo:
Just a couple tweaks around the metrics to make them more useful or give them a better description.

- Set hasUpdate to be initially true so that we set the storeSize on restart
- Ad a new metric for count of successful ad syncs
- Add a new metric for total latency of ad ingest

I'm hesitant to add any more metrics until we have a need for them. Also there's probably a couple things we want to record inside go-legs (like httpsync vs. dtsync which should be transparent to the ingester).

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -190,7 +190,7 @@ func (ing *Ingester) signalMetricsUpdate() {
 // metricsUpdate periodically updates metrics.  This goroutine exits when the
 // sigUpdate channel is closed, when Close is called.
 func (ing *Ingester) metricsUpdater() {
-	var hasUpdate bool
+	hasUpdate := true
 	t := time.NewTimer(time.Minute)
 
 	for {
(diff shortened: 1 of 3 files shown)


internal/ingest/ingest.go
```diff
@@ -190,7 +190,7 @@ func (ing *Ingester) signalMetricsUpdate() {
 // metricsUpdate periodically updates metrics.  This goroutine exits when the
 // sigUpdate channel is closed, when Close is called.
 func (ing *Ingester) metricsUpdater() {
-	var hasUpdate bool
+	hasUpdate := true
```
willscott, line 193: shouldn't this default to false?

MarcoPolo: I purposely set it to true so that it exposes this metric (store size) on start. Note that:
1. Store size is a gauge, so we need to track what the current value is.
2. We don't expose a metric in `/metrics` until we first record it.

The correct behavior should be that the store size is known at start time, and updated periodically. Right now the latter is true, but not the former.

As it is now, we won't know the store size until we ingest something.

Does that make sense? If so I can add a comment explaining the gist of this.

willscott (approved):

internal/ingest/ingest.go
willscott: makes sense. i had confused this with a call to update metrics that was short lived.

gammazero (approved):
Very nice!

PR #207, opened 2022-02-08
Update to the newest version of go-legs

gammazero:
It is necessary to update to teh new version of go-legs in order to read pubsub messages from the most recent provider.

A broken unit test was also fixed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/filecoin-project/go-address v0.0.5
 	github.com/filecoin-project/go-dagaggregator-unixfs v0.2.0
 	github.com/filecoin-project/go-indexer-core v0.2.8
-	github.com/filecoin-project/go-legs v0.2.7
+	github.com/filecoin-project/go-legs v0.2.8
 	github.com/frankban/quicktest v1.14.0
 	github.com/gogo/protobuf v1.3.2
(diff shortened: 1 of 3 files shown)


willscott:
can we also pull in legs#82 from @MarcoPolo ?

server/ingest/handler/ingest_handler.go
```diff
@@ -117,63 +116,31 @@ func (h *IngestHandler) IndexContent(ctx context.Context, data []byte) error {
...
 
-	addrs, ok := top["Addrs"]
-	if !ok {
-		return fmt.Errorf("missing addrs")
-	}
-	addrList := make([]*json.RawMessage, 0)
-	if err := json.Unmarshal(*addrs, &addrList); err != nil {
+	// Decode CID and originator addresses from message.
+	an := dtsync.Message{}
+	if err = an.UnmarshalCBOR(bytes.NewBuffer(data)); err != nil {
```
MarcoPolo, line 127: nit, we can use bytes.NewReader since we don't actually need a buffer.

MarcoPolo (approved):

server/ingest/handler/ingest_handler.go
MarcoPolo: Also we don't need to call io.ReadAll above, we can just pass the limit reader in here.

PR #208, opened 2022-02-09
Keep a local cache of http providers we encounter

MarcoPolo:
## Context

When syncing ad entries for an http provider we would fail because we wouldn't know what the http addr was for the given peer. 

## Proposed solution

I haven't thought too much about what the correct thing should be here, but for now this adds a cache in the ingester that remembers that a given peer has an http addr. Then in syncAdEntries we check for that and use it if it's available.

This alone doesn't fix http providers in the indexer. For that we need to merge https://github.com/filecoin-project/go-legs/pull/82 as well.


internal/ingest/ingest.go
```diff
@@ -150,6 +154,24 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
...
+				if p.Code == multiaddr.P_HTTP || p.Code == multiaddr.P_HTTPS {
+					isHttp = true
+					break
+				}
+			}
+		}
+		if isHttp {
+			ing.httpProvidersMu.Lock()
+			if ing.httpProviders == nil {
+				ing.httpProviders = make(map[peer.ID]string)
```
willscott, line 169: can we make this in `NewIngester` instead?

internal/ingest/ingest.go
```diff
@@ -150,6 +154,24 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
...
+			}
+			ing.httpProviders[peerID] = peerAddr.String()
```
willscott, line 171: can there be multiple locations for an http provider?

willscott (commented):
i wonder if we should keep this cache generally, rather than just for HTTP peers? it seems like we'd want the same caching and eviction logic regardless of whether the peer is http vs graphsync?

internal/ingest/ingest.go
MarcoPolo: Not in this flow since we are only given a single address to sync to. In general yes though.

MarcoPolo:
> i wonder if we should keep this cache generally, rather than just for HTTP peers? it seems like we'd want the same caching and eviction logic regardless of whether the peer is http vs graphsync?

I thought go-libp2p manages its own address book for its peers. If that's the case we don't have to do anything here.

MarcoPolo:
Maybe it would be better to support this in go-legs? Since that's how it works with the libp2p backend you'd expect it to work similarly for the http backend.

willscott:
I'm happy with it either here or in legs

willscott:
Cool. is this good to merge as is from your perspective

MarcoPolo:
I think so. Just blocked on an approved review. The failing test is the flaky e2e test.

gammazero:
@MarcoPolo You may need to rebase to pick up dependencies that allow you to use the new message encoding that the current provider is using.

gammazero:
What about adding the HTTP address to the peerstore?  That should be able to serve as the address cache.  That can be done here:
https://github.com/filecoin-project/go-legs/blob/main/subscriber.go#L360-L361

It was not previously done, because the HTTP syncs were intended to be manual or scheduled only.  Now that we can receive announcements over HTTP, it makes sense to keep HTTP addresses in the peer store as well.  If we do that this PR should not be needed.

MarcoPolo:
Closing this in favor of https://github.com/filecoin-project/go-legs/pull/85

PR #211, opened 2022-02-09
Bring in C flags to e2e_test env

MarcoPolo:
The compile step within the test failed for me because it didn't bring in the C flags which meant the compiler couldn't find the CoreFoundation library.

I don't have Go or most libraries installed globally. Instead I load them into my environment when I enter a project. I do this with a combination of [direnv](https://direnv.net/) and [Nix Flakes](https://nixos.wiki/wiki/Flakes). For reference here's the dev environment for the PL stack: https://github.com/MarcoPolo/pl-stack-dev-env/blob/master/flake.nix. Also see https://marcopolo.io/code/declarative-dev-environments/ for more information on this setup.

### DIFF
--- a/e2e_test.go
+++ b/e2e_test.go
@@ -136,11 +136,23 @@ func TestEndToEndWithReferenceProvider(t *testing.T) {
 
 	// Use a clean environment, with the host's PATH, and a temporary HOME.
 	// We also tell "go install" to place binaries there.
-	e.env = []string{
+	hostEnv := os.Environ()
+	var filteredEnv []string
+	for _, env := range hostEnv {
+		if strings.Contains(env, "CC") || strings.Contains(env, "LDFLAGS") || strings.Contains(env, "CFLAGS") {
+			// Bring in the C compiler flags from the host. For example on a Nix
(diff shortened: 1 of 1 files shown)


willscott:
is there a way to make CI happy with this as well?

MarcoPolo:
> is there a way to make CI happy with this as well?
I'm currently looking into this. I just had to make this change first to even run the test locally :)

MarcoPolo:
It's actually a breaking change in index-provider. For example hhttps://github.com/filecoin-project/index-provider/commit/15ae8070a46c1aa2ff8feb1c4391bc7549653dc5 works but https://github.com/filecoin-project/index-provider/commit/75130f4cee2a97e53bed0a2643cc520e2d6aac2b does not. I'll keep looking.

MarcoPolo:
actually it fails because there's a mismatch of go-legs version on the indexer and the index-provider.

MarcoPolo:
One race test is still failing on ubuntu go 1.16, but I can continue to debug that if we want to merge this in since it has a couple real fixes.

willscott:
yeah, happy to merge this now and follow up with a fix for the race condition bug

PR #217, opened 2022-02-14
Export function that signs advertisements

masih:
Export the function that is used to sign advertisements so that it can
be reused for testing purposes in `index-provider`.

Note that its counterpart verify function is already exported.

Bump version in preparation for release.


willscott (approved):

masih:
CI test failures seem unrelated and down to intermittently failing tests it seems. 
merging.

PR #218, opened 2022-02-14
Stabilize tests and fix wait for previous advertisement

gammazero:
There was an error in waiting for the previous ad, which resulted from moving some code from the StorageWriteOpener into the StorageHook.  When waiting for the previous ad, if that ad was not yet wait-able, the code would first check to see if the ad was in the datastore to determine if it had already been processed.  This was unreliable because the ad would already be in the datastore whether or not is was already processed.  This was because it was put into the datastore in the Storage write opener, but this check was done in the StorageHook.

This is fixed now by watching for the previous block to be processed for a time, and if it does not get processed during that wait, then check for wait-ability once more and wither wait or move on because the block in not being synced.

The `Ingester.Sync()` method returns a read-only channel of cid.Cid that the caller can optionally wait on.  Nothing is put onto the channel until Sync has completed syncing up to the head advertisement.

- This PR includes an unrelated fix in the client to close hosts it creates.
- The testEnv type now has a Close method to clean up resources.


gammazero:
This is included in #224

PR #219, opened 2022-02-15
Metrics around errors in syncAdEntries

MarcoPolo:
First pass at adding some obs around these error cases. 

I think the `EntryChunkAlreadyPresent` case is especially worrisome and likely wrong. Our current flow has us delete an entry chunk after processing, so the only way we have an entry chunk stored and hit that codepath is if we started processing an entry chunk and then crashed before finishing. Or if there are concurrent syncs (because of multiple publishers). So that specific metric is to highlight if we're seeing this in the wild, since it's almost certainly a bug.

Also makes a test a bit stronger by testing a restart in the middle of processing an Ad, not just the beginning.

### DIFF
--- a/internal/ingest/ingest_test.go
+++ b/internal/ingest/ingest_test.go
@@ -150,7 +150,7 @@ func TestRestartDuringSync(t *testing.T) {
 	cCid := typehelpers.RandomAdBuilder{
 		EntryChunkBuilders: []typehelpers.RandomEntryChunkBuilder{
 			{ChunkCount: 10, EntriesPerChunk: 10, EntriesSeed: 1},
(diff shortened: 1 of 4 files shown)


gammazero:
Looks like many tests are failing.  Please clean that up first.

MarcoPolo:
Flakey tests / tests that were broken before (just rebased now).

I'll simplify the ingest proposal tomorrow morning and with https://github.com/filecoin-project/go-legs/pull/89 I think we can have a fairly simple robust ingest algorithm.

gammazero:
I think we should merge this after #224.

MarcoPolo:
Closing in favor of #239

Issue #222, opened 2022-02-17
support admin sync that allows full sync


willscott:
currently there's a depth limit that makes it difficult to walk back to the beginning of the chain.

PR #223, opened 2022-02-18
Add admin `sync` with custom recursion depth

masih:
Add the ability to specify the recursion depth when syncing from a
provider via `admin sync` command. The flag `depth` now takes a
recursion depth limit. If non-zero, the depth is used to construct a
custom selector that:
  1. Recurses to the specified depth
  2. Stops at the currently latest synced ad

Fixes: #222

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -17,8 +17,11 @@ import (
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/query"
 	logging "github.com/ipfs/go-log/v2"
+	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/datamodel"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/traversal/selector"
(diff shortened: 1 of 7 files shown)


internal/ingest/ingest.go
```diff
@@ -135,14 +136,23 @@ func (ing *Ingester) Close() error {
...
+//
 // The Context argument controls the lifetime of the sync.  Canceling it
 // cancels the sync and causes the multihash channel to close without any data.
 //
 // Note that the multihash entries corresponding to the advertisement are
 // synced in the background.  The completion of advertisement sync does not
 // necessarily mean that the entries corresponding to the advertisement are
 // synced.
-func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiaddr.Multiaddr) (<-chan multihash.Multihash, error) {
+func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiaddr.Multiaddr, sel ipld.Node) (<-chan multihash.Multihash, error) {
```
gammazero, line 155: Do we really want to expose the ability to give any selector to sync?  While it is a more flexible API, I am wondering if it invites trouble.  If we think there may be use cases that need customized selectors, then we should keep this.  Otherwise, consider making the parameter being a simple depth limit and constructing the selector inside this function, in the interest of making an easier to use API.

gammazero (approved):

PR #224, opened 2022-02-18
New Ingester System

MarcoPolo:
## Context

Full context: https://github.com/filecoin-project/storetheindex/pull/177/files.

Summary:
Our current ingest system is a patchwork of fixes including notoriously the cidwaiter née lockchain code. This simplifies that code while also adhering to the following constraints:

1. If a an Ad is applied all older ads this ad references have also been
   applied. For example, given some chain `A <- B <- C`, the indexer will never
   be in the state that it indexed `A & C` but not `B`.
2. An indexer will index an Ad chain, but will not make any gaurantees about
   consistency in the presence of multiple ad chains for a given provider. For
   example if a provider publishes two ad chains at the same time chain1 and
   chain2 the indexer will apply whichever chain it learns about first first,
   then apply the other chain.
3. An indexer will not index the same Ad twice.
4. An indexer will be resilient to restarts. If the indexer goes down and comes
   back up it should not break constraint 1.

## Test improvements

This change also includes many commits from @gammazero that massively helped speed up tests. We can now run all ingest tests in ~1s (compared to > 35s before).

It should also fix the flakiness issues we've been seeing.

## Proposed Solution

This is very similar to the proposal in #177 but makes many simplifications that are enabled by https://github.com/filecoin-project/go-legs/pull/89.

Look at `runIngestStep` and `ingestWorkerLogic` for an overview of the new system.


e2e_test.go
```diff
@@ -115,6 +118,7 @@ func (e *e2eTestRunner) stop(cmd *exec.Cmd, timeout time.Duration) {
 }
 
 func TestEndToEndWithReferenceProvider(t *testing.T) {
+	t.Skip("TODO this is our longest test by far.")
```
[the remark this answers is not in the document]
MarcoPolo: I was skipping it for a couple reasons:
1. This test takes a while and I wanted to get CI results quickly.
2. This test was currently broken on macOS in CI (but working locally). Haven't investigated why.
3. This test is not deterministic/reproducible. Really this test should not be pulling from master, but that is a bigger change. Since it pulls from master it often breaks, and we (or at least I) think it's a storetheindex issue when it's really a dependency issue/incompatibility.

I think this test is useful, but I'd like to make it better. Maybe in a separate PR. What do you think?

internal/ingest/ingest.go
```diff
@@ -121,6 +151,16 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr indexer.Interface, reg *re
 		return nil, err
 	}
 
+	ing.toStaging, ing.cancelOnSyncFinished = ing.sub.OnSyncFinished()
```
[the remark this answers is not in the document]
MarcoPolo: Ah yes, good catch! With the recent change that adds the custom depth, this would have been broken. I'm not sure if we want to change those go-legs semantics, we should think about it. But for the mean time I can use a scoped block hook to create my own SyncFinishedEvent and call `runIngestStep` manually.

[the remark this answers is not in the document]
MarcoPolo: It probably makes the most sense to be able to pass a flag that updates the latest even with a selector that's passed in.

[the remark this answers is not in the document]
MarcoPolo: Short term fix here: f069cd5

internal/ingest/ingest.go
```diff
@@ -202,19 +277,53 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
...
-		// Notification channel; buffered so as not to block if no reader.
-		out <- c.Hash()
-	}()
+		// If latest head had already finished syncing, then do not wait for
+		// syncDone since it will never happen.
+		if latest == c {
+			log.Infow("Latest advertisement already processed", "adCid", c)
+			out <- c
+			return
+		}
```
gammazero, line 306: This may be wrong if a custom depth is given.  Does the wait break? Can we wait on something else?

gammazero (approved):

MarcoPolo:
I think constraint 3 needs to be removed. Instead we should have the following constraints:
* Reindexing an Ad multiple times consecutively is idempotent (`indexed([A,B]) == indexed([A,B,B])`)
* Reindexing a continuous subchain of Ads is idempotent (`indexed([A,B]) == indexed([A,B,A,B])`). Given the repeated subchain immediately proceeds the original subchain. So  (`indexed([A,B]) != indexed([A,B,C,A,B])`)


This will let us re-sync ad chains like what https://github.com/filecoin-project/storetheindex/pull/228 wants to do.

gammazero:
> I think constraint 3 needs to be removed. Instead we should have the following constraints:

I think constraint 3 is good, as long as we can still _traverse_ the ad again.

internal/ingest/linksystem.go
```diff
@@ -225,24 +225,6 @@ func (ing *Ingester) ingestAd(publisher peer.ID, adCid cid.Cid) error {
...
-		if skip {
-			// An ad that is later in the chain has deleted all content for
-			// this ad, so skip this ad.
-			err = nil
-			log.Infow("Skipped advertisement that is removed later")
-		} else if isRm {
-			// This ad deletes all content for a contextID.  So, skip any
-			// previous (earlier in chain) ads that arrive later, that have
-			// the same contextID and provider.
-			ing.addSkip(skipKey)
```
gammazero, line 333: A skip key is only added if the ad is a removal.  Since we only do the removal by context ID, which removes all content for a context ID, it should be safe to skip any earlier ads that have the same context ID, regardless of what they do.

MarcoPolo:
Chatted with Andrew. Really the practically best way for resyncing to work is to reprocess the later ads so that we process them in the context of the earlier ads. 

I made changes that will resync by using the constraint of:
* Reindexing a continuous subchain of Ads is idempotent `(indexed([A,B]) == indexed([A,B,A,B]))`. Given the repeated subchain immediately proceeds the original subchain. So `(indexed([A,B]) != indexed([A,B,C,A,B]))`

see https://github.com/filecoin-project/storetheindex/pull/224#issuecomment-1048317042

internal/ingest/linksystem.go
gammazero: Can handle in separate PR.

internal/ingest/linksystem.go
```diff
@@ -277,25 +174,34 @@ func (ing *Ingester) storageHook(pubID peer.ID, c cid.Cid) {
 	// Remove the content block from the data store now that processing it
 	// has finished.  This prevents storing redundant information in
 	// several datastores.
-	err = ing.ds.Delete(context.Background(), dsKey(c.String()))
+	err = ing.ds.Delete(context.Background(), dsKey(entryChunkCid.String()))
```
gammazero, line 177: nit: could use key from above

internal/ingest/linksystem.go
```diff
@@ -172,101 +146,24 @@ func providerFromAd(ad schema.Advertisement) (peer.ID, error) {
...
+func (ing *Ingester) ingestEntryChunk(publisher peer.ID, adCid cid.Cid, entryChunkCid cid.Cid) error {
+	log := log.With("publisher", publisher, "adCid", adCid, "cid", entryChunkCid)
 
 	// Get data corresponding to the block.
-	val, err := ing.ds.Get(context.Background(), dsKey(c.String()))
+	val, err := ing.ds.Get(context.Background(), dsKey(entryChunkCid.String()))
 	if err != nil {
 		log.Errorw("Error while fetching the node from datastore", "err", err)
-		return
+		return err
```
gammazero, line 156: Please do not log the error, and then also return it (presumably to be logged again).  Prefer to log at the highest level so that the message can contain as much info as possible.
```suggestion
		return fmt.Errorf("cannot fetch node from datastore: %w", err)
```

internal/ingest/linksystem.go
```diff
@@ -172,101 +146,24 @@ func providerFromAd(ad schema.Advertisement) (peer.ID, error) {
...
-	}
-
-	// Get the advertisement CID corresponding to the link CID, from the
-	// reverse map.  Then load the advertisement to get the metadata for
-	// indexing all the content in the incoming block.
-	adCid, err := popCidToAdMapping(context.Background(), ing.ds, c)
-	if err != nil {
-		log.Errorw("Error getting advertisement CID for entry CID", "err", err)
-		return
+		return err
```
gammazero, line 163: Same comment as above.
```suggestion
		return fmt.Errorf("error decoding ipldNode: %w", err) 
```

internal/ingest/linksystem.go
gammazero: Should this be deleted in a `defer` function immediately after it is read (at top of function), so that it gets cleaned up in the case of error, or do we want it to remain in datastore if this function exits with error before getting here?

MarcoPolo: done in 32df0a3

gammazero (approved):
Looks good.  Please squash-merge to squash out all the commits.

PR #228, opened 2022-02-22
Allow explicit sync to proceed beyond the latest synced advertisement

masih:
Add an option that allows an explicit sync to continue beyond the CID of
an advertisement that is already synced.

This is useful for cases where the ad recursion limit is expanded and
already synced advertisements need to be re-synced beyond the previous
recursion limit.

Relates to: #222

### DIFF
--- a/command/flags.go
+++ b/command/flags.go
@@ -137,16 +137,19 @@ var adminSyncFlags = []cli.Flag{
 	peerFlag,
 	indexerHostFlag,
 	&cli.StringFlag{
-		Name:     "addr",
-		Usage:    "Multiaddr address of peer to sync with",
-		Required: false,
+		Name:  "addr",
+		Usage: "Multiaddr address of peer to sync with",
 	},
 	&cli.Int64Flag{
 		Name: "depth",
 		Usage: "The recursion depth limit while syncing advertisements. The value -1 indicates no limit. " +
 			"If unspecified, the configured indexer advertisement recursion limit is used.",
-		Required: false,
-		Value:    0,
+		Value: 0,
+	},
+	&cli.BoolFlag{
+		Name:  "nolatest",
+		Usage: "Whether to continue traversal beyond the latest synced advertisement.",
+		Value: false,
 	},
 }
 
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -59,6 +59,8 @@ type Ingester struct {
 
 	skips      map[string]struct{}
 	skipsMutex sync.Mutex
+
+	cfg config.Ingest
 }
 
 // NewIngester creates a new Ingester that uses a go-legs Subscriber to handle
@@ -101,6 +103,7 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr indexer.Interface, reg *re
 		adWaiter:    adWaiter,
 		entriesSel:  entSel,
 		reg:         reg,
+		cfg:         cfg,
 	}
 
 	// Create and start pubsub subscriber.  This also registers the storage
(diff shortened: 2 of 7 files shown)


willscott (approved):

command/flags.go
```diff
@@ -137,16 +137,20 @@ var adminSyncFlags = []cli.Flag{
...
 	&cli.Int64Flag{
 		Name: "depth",
 		Usage: "The recursion depth limit while syncing advertisements. The value -1 indicates no limit. " +
 			"If unspecified, the configured indexer advertisement recursion limit is used.",
-		Required: false,
-		Value:    0,
+		Value: 0,
+	},
+	&cli.BoolFlag{
+		Name:    "ignoreLatestSync",
```
gammazero, line 150: Let's change this to "nolatest".  That is more consistent with other negation flags.

internal/ingest/ingest.go
```diff
@@ -163,32 +182,37 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
...
-		// If depth is non-zero, construct a selector for sync accordingly.
-		if depth != 0 {
-			log.With("depth", depth)
-			if depth < -1 {
-				log.Error("Recursion depth limit must not be less than -1")
-				return
-			}
+		// If depth is non-zero or traversal should not stop at the latest synced, then construct a
+		// selector to behave accordingly.
+		if depth != 0 || ignoreLatestSync {
```
gammazero, line 187: Can you put the contents of this code block in a separate function?
```go
if depth != 0 || ignoreLatestSync {
	sel, err = ing.makeLimitedDepthSelector(peerID, depth, ignoreLatest)
	if err != nil {
		log.Errorw("Cannot create limited depth selector", "err", err)
		return
	}
}
```

command/flags.go
```diff
@@ -137,16 +137,20 @@ var adminSyncFlags = []cli.Flag{
...
+		Aliases: []string{"ils"},
```
gammazero, line 151: If you use "nolatest", then I think the alias is not needed.

gammazero (approved):
Looks good, See comments before merge.

internal/ingest/ingest.go
```diff
@@ -225,6 +218,45 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
...
+			return nil, err
+		}
+
+		if latest != cid.Undef {
+			stopAt = cidlink.Link{Cid: latest}
+		}
+	}
+	// The stop link may be nil, in which case it is treated as no stop link.
+	// Log it regardless for debugging purposes.
+	log = log.With("stopAt", stopAt)
```
gammazero, line 248: I think this should be moved into `if latest != cid.Undef {`, so that it is not logged if not present.

internal/ingest/ingest.go
```diff
@@ -154,50 +165,32 @@ func (ing *Ingester) Close() error {
 // synced in the background.  The completion of advertisement sync does not
 // necessarily mean that the entries corresponding to the advertisement are
 // synced.
-func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiaddr.Multiaddr, depth int64) (<-chan multihash.Multihash, error) {
-	log := log.With("peerID", peerID)
+func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiaddr.Multiaddr, depth int64, nolatest bool) (<-chan multihash.Multihash, error) {
```
gammazero, line 168: I had only meant to use `nolatest` for the CLI flag and the API query parameter string.  Here, standard golang conventions apply, so `ignoreLatest` is OK.

Anyway, just a minor clarification, code is good either way you prefer.

gammazero (approved):
Good.  See if you want to change anything based on the very minor comments, and merge.  No additional review is needed.

internal/ingest/ingest.go
masih: I think we want to log it regardless for debugging purposes. Now that this function is refactored, the log including this is only made in debug level, and if debug level is set we probably want to know whether stop node was nil or not.

internal/ingest/ingest.go
masih: Sure thanks Andrew

PR #229, opened 2022-02-22
Encode query params into URL

gammazero:
The depth and nolatest query params were not getting set in the URL. This fixes the the issue.


willscott (approved):

masih:
Thank you for fixing this 🚀

PR #234, opened 2022-02-23
Explicit Sync always updates latest synced

gammazero:
When resyncing or syncing to a new head, the latest synced ad should be recorded.

When not doing a resync, and only setting a custom depth, make sure that go-legs still sends a sync done event since a non-nil selector is used for the custom depth.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -281,9 +281,6 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
 				seenAdCids = append(seenAdCids, c)
 			}),
 		}
-		if sel != nil && !resync && latest != cid.Undef {
-			opts = append(opts, legs.CheckAlreadySynced())
-		}
 		c, err := ing.sub.Sync(ctx, peerID, cid.Undef, sel, peerAddr, opts...)
 
 		// If this is a resync, we need to mark the adChain as unprocessed so that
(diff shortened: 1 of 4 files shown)


internal/ingest/ingest_test.go
```diff
@@ -199,7 +199,9 @@ func TestFailDuringResync(t *testing.T) {
 			{ChunkCount: 1, EntriesPerChunk: 1, EntriesSeed: 2},
 		}}.Build(t, te.publisherLinkSys, te.publisherPriv)
 
-	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
MarcoPolo, line 202: I personally feel like the test should just timeout here and we shouldn't worry about setting timeouts everywhere.

internal/ingest/ingest.go
```diff
@@ -276,11 +276,12 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
...
 		var seenAdCids []cid.Cid
 		opts := []legs.SyncOption{
+			legs.AlwaysUpdateLatest(),
 			legs.ScopedBlockHook(func(i peer.ID, c cid.Cid) {
 				seenAdCids = append(seenAdCids, c)
 			}),
 		}
-		if sel != nil && !resync {
+		if sel != nil && !resync && latest != cid.Undef {
 			opts = append(opts, legs.CheckAlreadySynced())
```
MarcoPolo, line 285: with https://github.com/filecoin-project/go-legs/pull/97 I think we could get rid of this block

MarcoPolo (approved):

internal/ingest/ingest_test.go
gammazero: Previously it was after 10m (before fix was made), which was an annoying time to wait for failure, so I added the timeout.

internal/ingest/ingest.go
gammazero: exterminated

PR #236, opened 2022-02-24
Fix nil ptr crash on websocket conns

MarcoPolo:
We were hitting this bug: https://github.com/libp2p/go-yamux/issues/79

This updates the go-libp2p dep to fix that bug


gammazero (approved):
Do we want to use the resource manager by default now?

MarcoPolo:
I think so

PR #237, opened 2022-02-24
Return more information on ad ingest errors

MarcoPolo:
## Context

There was no way to exit early if we encountered an error while processing ads. It would just skip the ad that errored, and keep ingesting the ad chain. This is a problem in the case where we temporarily can't get an ad (network is temporarily down). Previously we would skip the ad and try ingesting the next one. That would lead to many wrongfully skipped ads.

Instead, when encountering a network flake, we should stop processing the whole ad chain. Then we can try again later.

Of course there are some errors we encounter while processing an ad that aren't temporary. For example, the ad could be malformed. In that case we do want to skip the ad.

## Proposed Change

Create a new adIngestError type. This type includes the general kind of error this was (was it a permanent error or a temporary sync error). The ingest worker can then decide if it wants to skip this ad and continue or stop processing the ad chain altogether.

This also fixes a bug in the logic of skipping processing earlier ads if a later ad has already been processed.

This change also paves the way to better observability around errors. In a future PR I'll make that change that emits a metric on error, and we can bucket them by kind.

Aside: This is the kind of logic that would have been very very tricky to implement in our previous ingester system, so it's nice to see this refactor already paying dividends!

## Testing

I added a new test that tests that we are actually skipping earlier ads if a later ad was processed.

Until the caller gets the information on if an ad was skipped or not, it's kind of hard to test the flow that an adchain bailed early (or skipped some ads). Although I would like the caller to get that information so we can test that. I can do that in this PR or another (depending on how soon we want to get these fixes in).

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -13,6 +13,7 @@ import (
 	"github.com/filecoin-project/go-legs"
 	"github.com/filecoin-project/storetheindex/api/v0/ingest/schema"
 	"github.com/filecoin-project/storetheindex/config"
+	"github.com/filecoin-project/storetheindex/internal/metrics"
 	"github.com/filecoin-project/storetheindex/internal/registry"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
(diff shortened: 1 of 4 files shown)


internal/ingest/ingest_test.go
```diff
@@ -488,22 +488,55 @@ func TestReSyncWithDepth(t *testing.T) {
 	wait, err := te.ingester.Sync(context.Background(), te.pubHost.ID(), te.pubHost.Addrs()[0], 1, false)
 	require.NoError(t, err)
 	<-wait
-	allAds := typehelpers.AllMultihashesFromAdLink(t, adHead, te.publisherLinkSys)
-	require.NoError(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allAds[2:]))
-	require.Error(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allAds[0:1]))
+	allMHs := typehelpers.AllMultihashesFromAdLink(t, adHead, te.publisherLinkSys)
+	require.NoError(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allMHs[2:]))
+	require.Error(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allMHs[0:1]))
```
gammazero, line 493:
```suggestion
	require.Error(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allMHs[0:2]))
```

The test is actually incorrect because `allMHs[2:]` is a zero-length slice. Add
```go
{ChunkCount: 1, EntriesPerChunk: 1, EntriesSeed: 3},
```
to `EntryChunkBuilders`

internal/ingest/ingest_test.go
```diff
@@ -488,22 +488,55 @@ func TestReSyncWithDepth(t *testing.T) {
...
 	wait, err = te.ingester.Sync(context.Background(), te.pubHost.ID(), te.pubHost.Addrs()[0], 2, false)
 	require.NoError(t, err)
 	<-wait
-	require.NoError(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allAds[2:]))
-	require.Error(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allAds[0:1]))
+	require.NoError(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allMHs[2:]))
+	require.Error(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allMHs[0:1]))
 
 	// When resync with greater depth, check that everything in synced.
 	wait, err = te.ingester.Sync(context.Background(), te.pubHost.ID(), te.pubHost.Addrs()[0], 2, true)
```
gammazero, line 503:
```suggestion
	wait, err = te.ingester.Sync(context.Background(), te.pubHost.ID(), te.pubHost.Addrs()[0], -1, true)
```

internal/ingest/ingest_test.go
```diff
@@ -488,22 +488,55 @@ func TestReSyncWithDepth(t *testing.T) {
...
 	<-wait
-	allAds := typehelpers.AllMultihashesFromAdLink(t, adHead, te.publisherLinkSys)
-	require.NoError(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allAds[2:]))
-	require.Error(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allAds[0:1]))
+	allMHs := typehelpers.AllMultihashesFromAdLink(t, adHead, te.publisherLinkSys)
+	require.NoError(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allMHs[2:]))
+	require.Error(t, checkAllIndexed(te.ingester.indexer, te.pubHost.ID(), allMHs[0:1]))
 
 	// When not resync, check that nothing beyond the latest is synced.
 	wait, err = te.ingester.Sync(context.Background(), te.pubHost.ID(), te.pubHost.Addrs()[0], 2, false)
```
gammazero, line 496:
```suggestion
	wait, err = te.ingester.Sync(context.Background(), te.pubHost.ID(), te.pubHost.Addrs()[0], -1, false)
```

internal/ingest/linksystem.go
```diff
@@ -182,82 +182,76 @@ func (ing *Ingester) ingestEntryChunk(publisher peer.ID, adCid cid.Cid, entryChu
...
+	adIngestEntryChunkErr adIngestState = "ingestEntryChunkErr"
+)
+
+func (e adIngestError) Error() string {
+	return fmt.Sprintf("%s: %s\n", e.state, e.err)
+}
+
+// ingestAd fetches all the entries for a single advertisement and processes
+// them.
+func (ing *Ingester) ingestAd(publisher peer.ID, adCid cid.Cid) (err error) {
```
gammazero, line 209:
```suggestion
func (ing *Ingester) ingestAd(publisher peer.ID, adCid cid.Cid) error {
```

internal/ingest/linksystem.go
```diff
@@ -182,82 +182,76 @@ func (ing *Ingester) ingestEntryChunk(publisher peer.ID, adCid cid.Cid, entryChu
 	return nil
 }
 
-// ingestAd fetches all the entries for a single advertisement
-func (ing *Ingester) ingestAd(publisher peer.ID, adCid cid.Cid) error {
+type adIngestError struct {
+	state adIngestState
```
gammazero, line 186: Instead of having different error states, would it be simpler to use a bool `permanent bool` instead?

internal/ingest/linksystem.go
```diff
@@ -182,82 +182,76 @@ func (ing *Ingester) ingestEntryChunk(publisher peer.ID, adCid cid.Cid, entryChu
...
+const (
+	adIngestDatastoreErr        adIngestState = "datastoreErr"
+	adIngestIndexerErr          adIngestState = "indexerErr"
+	adIngestDecodingErr         adIngestState = "decodingErr"
+	adIngestMalformedErr        adIngestState = "malformedErr"
+	adIngestRegisterProviderErr adIngestState = "registerErr"
+	adIngestSyncEntriesErr      adIngestState = "syncEntriesErr"
+	// Happens if there is an error during ingest of an entry chunk (rather than fetching it).
+	adIngestEntryChunkErr adIngestState = "ingestEntryChunkErr"
+)
```
gammazero, line 201: Does the code really care about these individual states, or only whether or not they are permanent type errors?  If the code does not care, then the description should be part of the error message, and use a `permanent bool` for the code to handle permanent and non-permanent errors differently.

gammazero (commented):
Please see comments and suggestions for some tweaks and simplification.  Otherwise, looks good.

internal/ingest/ingest.go
```diff
@@ -678,30 +679,63 @@ func (ing *Ingester) ingestWorker() {
...
+		}
+
+		if markErr := ing.markAdProcessed(msg.publisher, adCid); markErr != nil {
+			log.Errorw("Failed to mark ad as processed", "err", markErr)
+		} else {
+			stats.Record(context.Background(), metrics.AdSyncedCount.M(1))
+		}
+		// Distribute the legs.SyncFinished notices to waiting Sync calls.
+		ing.inEvents <- legs.SyncFinished{Cid: adCid, PeerID: msg.publisher}
+	}
```
gammazero, line 740:
```suggestion
		// Cleanup add from datastore.
		err = ing.ds.Delete(context.Background(), dsKey(adCid.String()))
		if err != nil {
			log.Errorw("Failed to clean up ad from datastore", "err", err)
		}
	}
```
I think this can be cleaned up now that it is completely ingested.  Maybe it is not even necessary to store the ad in the datastore at all, and use only the adCache instead.

internal/ingest/ingest_test.go
MarcoPolo: I'll just fix it to just need the two ads.

internal/ingest/linksystem.go
MarcoPolo: This is on purpose for a couple reasons:

1. It makes adding metrics to these error states more useful. We could have an ingest ad error counter metric and we add a label for each kind of error. Then we know something is wrong if we suddenly see a spike in say decoding errors. And, more helpfully, we know the general kind of thing that is wrong.
2. The caller should decide what to do in the different error states. While we could add a `permanent` bool as well, the caller may want to do something else depending on the error state. For example, in a datastore err we may want to treat that as a temporary error if our datastore is actually a network filesystem rather than a local filesystem. ingestAd shouldn't try to predict what the caller would want to do in it's different errors. It should just report what failed.

Thoughts?

internal/ingest/linksystem.go
MarcoPolo: Let's continue this discussion here: https://github.com/filecoin-project/storetheindex/pull/237#discussion_r814924082

internal/ingest/linksystem.go
MarcoPolo: see https://github.com/filecoin-project/storetheindex/pull/239/files#diff-50d2db719b550ec2183b831d2727d9184635a5647247ea8872ff82d91b6e31bdR730 for 1.

gammazero: OK, then there is a code-oriented reason for the states, so keep it.

internal/ingest/ingest.go
```diff
@@ -678,30 +679,63 @@ func (ing *Ingester) ingestWorker() {
...
+			splitAtIndex = i
+			break
+		}
 	}
+	yetToBeProcessed := msg.cids[:splitAtIndex]
+
+	log.Infow("Running worker on ad stack", "headAdCid", msg.cids[0], "publisher", msg.publisher, "numAdsToProcess", len(yetToBeProcessed))
+	for i := len(yetToBeProcessed) - 1; i >= 0; i-- {
+		// Note that we are iterating backwards here. Earliest to newest.
+		adCid := yetToBeProcessed[i]
```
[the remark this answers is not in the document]
gammazero: You are correct in what you said here, so maybe a matter of preference.  I thought readability was improved by having one less variable in this case; maybe just my opinion.

gammazero (approved):

PR #240, opened 2022-02-25
Handle additional errors as permanent

gammazero:


MarcoPolo (commented):
Do we know if this is permanent vs temporary? Like is this error that we reached the endpoint and they said no vs. we didn't find the content because the endpoint couldn't be reached?

Also we'd want something similar in the http backend. Do this likely requires a change in legs

PR #242, opened 2022-02-26
Gitops readiness

coryschwartz:
Preparedness for a future move to weave works.

I switched you to bare yaml files rather than helm, since as @masih indicated, the helm chart was probably a little difficult to tailor anyway. 

This branch (gitops-readiness) is currently being deployed by our proof-of-concept weave/flux CD, and it is running in the us-east-2 cluster.


coryschwartz:
As with the similar PR I submmitted to the dealbot -- the intention is not to force a move to gitops, unless you want to try it out.

The fluxcd installation we have currently running will go away once we have a contract with WeaveWorks. However, it should be a drop-in replacement -- if it works here, it will work with Weave . That's the idea.

The existing deployment system (triggered by CircleCI) is of course goign to continue to run until Weave is able to take over for us. This is just to be ready for when that happens.

willscott:
is there a grafana / visibility into this deployment instance we should be using?

deploy/us-east-2-dev/config.yaml
```diff
@@ -0,0 +1,66 @@
...
+        "ValueStoreType": "sth"
+      },
+      "Ingest": {
+        "PubSubTopic": "/indexer/ingest/mainnet",
+        "StoreBatchSize": 256,
+        "SyncTimeout": "2h0m0s",
+        "EntriesDepthLimit": 400,
+        "AdvertisementDepthLimit": 400
+      }
+    }
```
masih, line 66: It may be clearer to:
1. Define env vars in code and use configmap `envFrom` in deployment instead of mounting config.
    This will have the added advantage of making overrides easier when we have a base overlay and a concrete overlay.
3. Remove any default values.

deploy/us-east-2-dev/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+  selector:
+    matchLabels:
+      app: storetheindex
+      release: storetheindex-0
+  volumeClaimTemplates:
+  template:
+    metadata:
+      labels:
+        app: storetheindex
+        chart: lotus-bundle-0.0.13
```
masih, line 20: We might want to drop help labels and use `commonLabels` in Kustomization YAML.

deploy/us-east-2-dev/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+        - name: storetheindex-config-volume
+          configMap:
+            name: storetheindex-0-storetheindex-config
+        - name: shared-volume
+          persistentVolumeClaim:
+            claimName: storetheindex-pvc
+      initContainers:
+        - args:
+          - -c
+          - envsubst </config/config >/data/storetheindex/config
```
masih, line 39: I think using `envFrom` and `configmap`of environment variables will eliminate the need for this and the init container alltogher.

deploy/us-east-2-dev/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+          image: cmd.cat/gettext
+          name: setupconfig
+          volumeMounts:
+          - mountPath: /data/storetheindex
+            name: shared-volume
+            subPath: ntwk-mainnet-storetheindex/storetheindex-0
+          - mountPath: /config
+            name: storetheindex-config-volume
+      containers:
+        - name: application
```
masih, line 62: Consider a name more reminiscent of the application running.

deploy/us-east-2-dev/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+          image: filecoin/storetheindex:20e2265419aded6f5ae4bd6c3d7d173e270a94c2
+          imagePullPolicy: Always
+          command:
+            []
```
masih, line 66: Omit redundant lines?

deploy/us-east-2-dev/ingress.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: storetheindex-0-storetheindex
+  annotations:
+    kubernetes.io/ingress.class: nginx
+spec:
+  rules:
+  - host: gitops.cid.contact
```
masih, line 10: `stg-2` instead of `gitops`? To clients of this host, GitOps means very little.

deploy/us-east-2-dev/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+---
+kind: Kustomization
+metadata:
```
masih, line 3: * Add version
* Specify namespace

deploy/us-east-2-dev/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+resources:
+  - config.yaml
+  - deployment.yaml
+  - ingress.yaml
+  - kustomization.yaml
+  - pv.yaml
+  - pvc.yaml
+  - secrets-sealed.yaml
+  - servicemonitor.yaml
+  - services.yaml
```
masih, line 14: Re comments earlier on using env vars etc. consider using `configMapGenerator` here.

deploy/us-east-2-dev/secrets-sealed.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: bitnami.com/v1alpha1
+kind: SealedSecret
```
masih, line 2: A totally different subject for discussion: Have you considered using [`sops`](https://github.com/mozilla/sops) with AWS KMS instead of sealed secrets? That will offer a simpler set up while allowing fine control over encryption keys. It also has built-in support in fluxCD I believe.

deploy/us-east-2-dev/servicemonitor.yaml
```diff
@@ -0,0 +1,19 @@
...
+# Source: lotus-bundle/templates/service-monitor.yaml
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: storetheindex-0-storetheindex
+  namespace: default
+spec:
+  namespaceSelector:
+    matchNames:
+      - default
```
masih, line 11: Should the namespace be `default`?

deploy/us-east-2-dev/services.yaml
```diff
@@ -0,0 +1,61 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: storetheindex-0-storetheindex
+  namespace: default
```
masih, line 6: Ditto re namespace (and other places).

deploy/us-east-2-dev/services.yaml
```diff
@@ -0,0 +1,61 @@
...
+    targetPort: 3002
+  selector:
+    app: storetheindex
+    release: storetheindex-0
+  type: ClusterIP
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: storetheindex-0-storetheindex-1
```
masih, line 43: Can we give services here a more descriptive name, say `finder`? Ditto re other services (and objects in storetheindex namespace).

P.S. I am not sure why pretty much ever object name is qualified by `storetheindex`? I imagine the separation of namespace is enough, no? 

My vote: Simpler, more descriptive names would make working with kubectl and debugging easier.

masih (commented):
@coryschwartz thank you so much for putting this PR together. Cannot wait to see it running 🍻

deploy/us-east-2-dev/deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+          image: cmd.cat/gettext
+          name: setupconfig
+          volumeMounts:
+          - mountPath: /data/storetheindex
+            name: shared-volume
+            subPath: ntwk-mainnet-storetheindex/storetheindex-0
+          - mountPath: /config
+            name: storetheindex-config-volume
+      containers:
+        - name: application
```
MarcoPolo, line 55: What's the point of this container? You now have two containers that are almost identical in the compiled yaml file

masih: This is the main application container, i.e. the container that runs the daemon. The previous one is used to basically run `sit init` as an `initContainer`. I think we can remove the init container entirely and use mounted pre-initialised config pointed out [here](https://github.com/filecoin-project/storetheindex/pull/242/files/b9b880b4fb019405b83ecf4fefaa81db0fe28fd2#r815721055).

MarcoPolo: This is the compiled yaml:
https://gist.github.com/MarcoPolo/cf861f5fd6d46bb694ffcaa4ea2daa92

Notice you have two identical containers here ([named application](https://gist.github.com/MarcoPolo/cf861f5fd6d46bb694ffcaa4ea2daa92#file-full-yaml-L183)) and here ([named storetheindex](https://gist.github.com/MarcoPolo/cf861f5fd6d46bb694ffcaa4ea2daa92#file-full-yaml-L203)). This is completely independent of the initContainer defined [here](https://gist.github.com/MarcoPolo/cf861f5fd6d46bb694ffcaa4ea2daa92#file-full-yaml-L214).

My question is why is there both a container named application and another one named storetheindex.

masih: I totally misunderstood your question; You are right, that's not clear for me either :)

masih:
I think this PR now might redundant considering: https://github.com/filecoin-project/storetheindex/pull/297

If there are no objections I'll close this

PR #243, opened 2022-02-28
Remove readBeforeErr hack now that cbor-gen bug is fixed

MarcoPolo:
Now that cbor-gen correctly reads from an io.Reader we can drop this workaround.


willscott (approved):

gammazero:
Is there a dependency update that needs to be part of this PR?

MarcoPolo:
https://github.com/filecoin-project/go-legs/pull/101

Already in the current version of go-legs this is using. No dep update needed, we already did it.

PR #246, opened 2022-03-01
Reduce log volume and log progress

gammazero:
- Reduce logging by logging some facilities at info or debug
- Log progress when processing chain of ads

Log volume is reduced when logging at info or debug level, by setting the log level to "warn" level.  This can be canceled by using the "log-all" flag.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -720,9 +720,17 @@ func (ing *Ingester) ingestWorkerLogic(msg toWorkerMsg) {
 	}
 
 	log.Infow("Running worker on ad stack", "headAdCid", msg.adInfos[0].cid, "publisher", msg.publisher, "numAdsToProcess", splitAtIndex)
+	var count int
 	for i := splitAtIndex - 1; i >= 0; i-- {
 		// Note that we are iterating backwards here. Earliest to newest.
 		ai := msg.adInfos[i]
+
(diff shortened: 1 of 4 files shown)


internal/ingest/ingest.go
```diff
@@ -698,9 +698,17 @@ func (ing *Ingester) ingestWorkerLogic(msg toWorkerMsg) {
...
+	var count int
 	for i := splitAtIndex - 1; i >= 0; i-- {
 		// Note that we are iterating backwards here. Earliest to newest.
 		ai := msg.adInfos[i]
+
+		count++
+		log.Infow("Processing advertisement",
+			"adCid", ai.cid,
+			"publisher", msg.publisher,
+			"progress", fmt.Sprintf("%d of %d", count, splitAtIndex))
```
willscott, line 710: this will be a lot of log lines. are we worried this will be too spamy? seems maybe `Debug` level?

gammazero: It was adding fields to the existing "Processing advertisement" log entry, of which there is one per advertisement.  I think this should be at info level as it denotes the start ad processing.  I thing there are log entries for individual entries that I can move to debug.

gammazero:
Bump.  This makes the logs much more readable IMO.

MarcoPolo (approved):

PR #247, opened 2022-03-03
Indexer auto-syncs after period of no contact with publisher

gammazero:
If the indexer has had no contact with any publisher for a provider, the indexer attempts to sync with the provider's last known publisher.  As long as there is no contact with a publisher for the provider, the indexer continues retrying to sync periodically until contact is made or until the indexer gives up after a configurable abount of time.

### DIFF
--- a/config/discovery.go
+++ b/config/discovery.go
@@ -14,9 +14,18 @@ type Discovery struct {
 	// Policy configures which providers are allowed and blocked
 	Policy Policy
 	// PollInterval is the amount of time to wait without getting any updates
-	// from a provider, before sending a request for the latest advertisement.
+	// for a provider, before sending a request for the latest advertisement.
 	// Values are a number ending in "s", "m", "h" for seconds. minutes, hours.
 	PollInterval Duration
+	// PollRetruAfter is the amount of time that must elapse from one poll
+	// attempt, without a response, to the next poll attempt.  This must be
+	// smaller than PollStopAfter for there to be more than one poll attempt.
+	// Time resolution is in hours.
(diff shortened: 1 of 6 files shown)


config/discovery.go
```diff
@@ -32,6 +41,8 @@ func NewDiscovery() Discovery {
 		LotusGateway:   "https://api.chain.love",
 		Policy:         NewPolicy(),
 		PollInterval:   Duration(24 * time.Hour),
+		PollRetryAfter: Duration(5 * time.Hour),
+		PollStopAfter:  Duration(48 * time.Hour),
```
willscott, line 45: I'd default probably to 1 week rather than 2 days

internal/registry/registry.go
```diff
@@ -25,8 +25,10 @@ import (
 )
 
 const (
-	// providerKeyPath is where provider info is stored in to indexer repo
+	// providerKeyPath is where provider info is stored in to indexer repo.
 	providerKeyPath = "/registry/pinfo"
+	// pollCheckInterval is how often to check for idle providers to poll.
+	pollCheckInterval = time.Hour
```
willscott, line 31: why is this a constant? can we make it a config param as well?

willscott (approved):

internal/registry/registry.go
gammazero: I changed this to use `cfg.PollRetryAfter` as the check interval, becuase that will be the lowest frequency the check can be run and still be able to retry to sync in the configured time.

willscott (approved):

PR #250, opened 2022-03-07
Support multiple metadata protocols

willscott:
* metadata protocol wire format extended to support multiple uvarints in increasing order

### DIFF
--- a/api/v0/metadata.go
+++ b/api/v0/metadata.go
@@ -22,7 +22,9 @@ func (e ErrInvalidMetadata) Error() string {
 
 // Metadata is data that provides information about retrieving
 // data for an index, from a particular provider.
-type Metadata struct {
+type Metadata []byte
+
+type ParsedMetadata struct {
 	Protocols []ProtocolMetadata
 }
 
@@ -38,12 +40,12 @@ type ProtocolMetadata interface {
 }
 
 var (
-	_ encoding.BinaryMarshaler   = (*Metadata)(nil)
-	_ encoding.BinaryUnmarshaler = (*Metadata)(nil)
+	_ encoding.BinaryMarshaler   = (*ParsedMetadata)(nil)
+	_ encoding.BinaryUnmarshaler = (*ParsedMetadata)(nil)
 )
 
 // Equal determines if two Metadata values are equal.
-func (m Metadata) Equal(other Metadata) bool {
+func (m ParsedMetadata) Equal(other ParsedMetadata) bool {
 	if len(m.Protocols) != len(other.Protocols) {
 		return false
 	}
@@ -62,7 +64,7 @@ func (m Metadata) Equal(other Metadata) bool {
 }
 
 // Protocols returns the parsed protocols
-func (m *Metadata) Codes() []multicodec.Code {
+func (m *ParsedMetadata) Codes() []multicodec.Code {
 	protocols := make([]multicodec.Code, len(m.Protocols))
 	for i, p := range m.Protocols {
 		protocols[i] = p.Protocol()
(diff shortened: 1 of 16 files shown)


willscott:
> Only problem is that e2e test fails

I believe this is because the indexer in de-serializing wire format is also attempting to parse the metadata in this structure. since it won't have any protocols registered, it will fail. I need to switch the code paths in the indexer to just treat metadata as unparsed bytes.

api/v0/metadata.go
```diff
@@ -19,55 +22,183 @@ func (e ErrInvalidMetadata) Error() string {
 
 // Metadata is data that provides information about retrieving
 // data for an index, from a particular provider.
-type Metadata struct {
-	// ProtocolID defines the protocol used for data retrieval.
-	ProtocolID multicodec.Code
-	// Data is specific to the identified protocol, and provides data, or a
-	// link to data, necessary for retrieval.
-	Data []byte
+type Metadata []byte
```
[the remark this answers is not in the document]
willscott: that's true. and i'm not totally sold on this, but we should have an answer we're happy with for how to how to handle providers that speak multiple protocols.

* do we stick to our initial plan of different records for each protocol? what does the ingest interface look like to make it easier for providers to advertise multiple protocols?
* do we change the ingest format as well, putting a payload length after each payload, so we can parse the list even when there are protocol ids that aren't understood?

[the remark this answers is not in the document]
gammazero: As long as the outer Metadata is:
```go
type Metadata struct {
	ProtocolID multicodec.Code
	Data []byte
}
```
Then anything can go inside the Data, and we can find an efficient way to pack it and delimit it.  It could be a list of protocol IDs with a single metadata, or a list of separate complete Metadata values, and the outer encoding tells what to expect inside.

masih (approved):

api/v0/metadata.go
```diff
@@ -19,55 +22,183 @@ func (e ErrInvalidMetadata) Error() string {
...
 }
 
-// UnmarshalBinary implements encoding.BinaryUnmarshaler.
-func (m *Metadata) UnmarshalBinary(data []byte) error {
-	protocol, protoLen, err := varint.FromUvarint(data)
-	if err != nil {
-		return err
+func MetadataFromBytes(data []byte) (*ParsedMetadata, error) {
+	m := ParsedMetadata{
+		Protocols: make([]ProtocolMetadata, 0),
```
gammazero, line 149:
```suggestion
```

api/v0/metadata.go
```diff
@@ -19,55 +22,183 @@ func (e ErrInvalidMetadata) Error() string {
...
 	}
```
gammazero, line 150:
```suggestion
	m := new(ParsedMetadata)
```

api/v0/metadata.go
```diff
@@ -19,55 +22,183 @@ func (e ErrInvalidMetadata) Error() string {
...
-	if protocol == 0 {
-		return &ErrInvalidMetadata{Message: "encountered protocol ID 0 on decode"}
+	if err := m.UnmarshalBinary(data); err != nil {
+		return nil, err
 	}
+	return &m, nil
```
gammazero, line 154:
```suggestion
	return m, nil
```

api/v0/registry.go
```diff
@@ -0,0 +1,29 @@
+package v0
+
+import (
+	"fmt"
```
gammazero, line 4:
```suggestion
	"errors"
	"fmt"
```

masih:
Please note, the following two PRs will make this PR redundant:
* https://github.com/filecoin-project/storetheindex/pull/260
* https://github.com/filecoin-project/index-provider/pull/198

The former removes all `Metadata` parsing from storetheindex, treating metadata as `[]byte` only.
The latter moves all of the functionality, including the stuff implemented here, to `index-provider` library.

masih (changes requested):
See: https://github.com/filecoin-project/storetheindex/pull/250#issuecomment-1068507634

gammazero:
Replaced by:
- https://github.com/filecoin-project/storetheindex/pull/260
- https://github.com/filecoin-project/index-provider/pull/198

PR #252, opened 2022-03-09
add cors headers to find interface

willscott:
as you can see at https://cid.contact currently, query requests fail to load the json due to lack of cors cross-domain.


willscott:
i think it's fine to leave that to an nginx layer rather than trying to have it be the indexer manage that complexity. i think the default is that anyone who can access the endpoint should be able to consume find requests browser or otherwise

PR #256, opened 2022-03-10
Set latestsynchandler on go legs

MarcoPolo:
## Context

This hooks up the changes from https://github.com/filecoin-project/go-legs/pull/106. See that PR for more context.

## Proposed Change

The latestSync state of go-legs now asks the indexer what the last ad processed was. The indexer ignores updates on latestSync from legs, since the indexer is the source of truth here.

## Tests

Added a test that repros that bug.


gammazero:
Please `go mod tidy`

gammazero (approved):

PR #258, opened 2022-03-11
Don't optimistically delete an Ad

MarcoPolo:
## Context

This fixes a bug where an ingest loop loads an ad (and thereby deleting it) and for whatever reason fails to finish ingesting the ad. The next time the loop tries to ingest the ad it will fail to load the ad and skip it.

Note this happens if we are ingesting the same ad chain back to back.

## Proposed change

Don't delete the ad when we load it. Instead only delete the ad when we've finished processing it, since that's really the only time we'll know that we'll never need this ad again.

Most of this diff is the test. The actual fix is just deleting the delete call in `loadAd`.

## Tests

This also adds a test to make sure we correctly ingest all ads in the case that we are given the same adChain back to back. Which can happen in at least the following cases:
1. A publisher broadcast the same head twice in gossipsub
2. We call sync twice back to back.
3. Combination of 1 & 2.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -40,6 +40,16 @@ const (
 	adProcessedPrefix = "/adProcessed/"
 )
 
+type adProcessedEvent struct {
+	publisher peer.ID
+	// The head of the chain we're proecessing.
+	headAdCid cid.Cid
+	// The actual adCid we're processing.
+	adCid cid.Cid
+	// if not nil we failed to process the ad for adCid
+	err error
+}
+
 // Ingester is a type that uses go-legs for the ingestion protocol.
 type Ingester struct {
 	host    host.Host
@@ -58,13 +68,13 @@ type Ingester struct {
 
 	cfg config.Ingest
 
-	// inEvents is used to send a legs.SyncFinished to the distributeEvents
-	// goroutine, when an advertisement in marked complete.
-	inEvents chan legs.SyncFinished
+	// inEvents is used to send a adProcessedEvent to the distributeEvents
+	// goroutine, when an advertisement in marked complete or err'd.
+	inEvents chan adProcessedEvent
 
 	// outEventsChans is a slice of channels, where each channel delivers a
 	// copy of a legs.SyncFinished to an onAdProcessed reader.
-	outEventsChans map[peer.ID][]chan cid.Cid
+	outEventsChans map[peer.ID][]chan adProcessedEvent
 	outEventsMutex sync.Mutex
 
 	waitForPendingSyncs sync.WaitGroup
@@ -114,7 +124,7 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr indexer.Interface, reg *re
 		entriesSel:  entSel,
 		reg:         reg,
 		cfg:         cfg,
-		inEvents:    make(chan legs.SyncFinished, 1),
+		inEvents:    make(chan adProcessedEvent, 1),
 
 		closePendingSyncs: make(chan struct{}),
 
@@ -293,9 +303,9 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
 		log.Debugw("Syncing advertisements up to latest", "adCid", c)
 		for {
 			select {
-			case adCid := <-syncDone:
-				log.Debugw("Synced advertisement", "adCid", adCid)
-				if adCid == c {
+			case adProcessedEvent := <-syncDone:
+				log.Debugw("Synced advertisement", "adCid", adProcessedEvent.adCid)
+				if adProcessedEvent.adCid == c || adProcessedEvent.err != nil && adProcessedEvent.headAdCid == c {
 					out <- c
 					ing.signalMetricsUpdate()
 					return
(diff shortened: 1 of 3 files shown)


internal/ingest/ingest.go
```diff
@@ -58,13 +68,13 @@ type Ingester struct {
...
 
-	// inEvents is used to send a legs.SyncFinished to the distributeEvents
-	// goroutine, when an advertisement in marked complete.
-	inEvents chan legs.SyncFinished
+	// inEvents is used to send a adProcessedEvent to the distributeEvents
+	// goroutine, when an advertisement in marked complete or err'd.
+	inEvents chan adProcessedEvent
 
 	// outEventsChans is a slice of channels, where each channel delivers a
 	// copy of a legs.SyncFinished to an onAdProcessed reader.
```
gammazero, line 76:
```suggestion
	// copy of an adProcessedEvent to an onAdProcessed reader.
```

internal/ingest/ingest.go
```diff
@@ -365,6 +375,11 @@ func (ing *Ingester) markAdUnprocessed(adCid cid.Cid) error {
 	return ing.ds.Put(context.Background(), datastore.NewKey(adProcessedPrefix+adCid.String()), []byte{0})
 }
 
+func (ing *Ingester) adHasBeenprocessed(adCid cid.Cid) bool {
```
gammazero, line 378:
```suggestion
func (ing *Ingester) adAlreadyProcessed(adCid cid.Cid) bool {
```

internal/ingest/ingest.go
```diff
@@ -365,6 +375,11 @@ func (ing *Ingester) markAdUnprocessed(adCid cid.Cid) error {
...
+	v, err := ing.ds.Get(context.Background(), datastore.NewKey(adProcessedPrefix+adCid.String()))
+	return err == nil && v[0] == byte(1)
```
gammazero, line 380:
```suggestion
	v, err := ing.ds.Get(context.Background(), datastore.NewKey(adProcessedPrefix+adCid.String()))
	if err != nil {
		if err != datastore.ErrNotFound {
			log.Errorw("Failed to read advertisement processed state from datastore", "err", err)
		}
		return false
	} 
	return v[0] == byte(1)
```

internal/ingest/ingest.go
```diff
@@ -552,10 +567,17 @@ func (ing *Ingester) runIngestStep(syncFinishedEvent legs.SyncFinished) {
...
+			// This ad has been processed so all earlier ads already have been
+			// processed.
+			break
+		}
+
 		ad, err := ing.loadAd(c)
 		if err != nil {
 			stats.Record(context.Background(), metrics.AdLoadError.M(1))
-			log.Errorf("Failed to load ad CID: %s skipping", err)
+			log.Errorf("Failed to load ad CID: %s. skipping. err:%s", c, err)
```
gammazero, line 580:
```suggestion
			log.Errorw("Failed to load advertisement CID, skipping", "cid", c, "err", err)
```

internal/ingest/ingest.go
```diff
@@ -293,9 +303,9 @@ func (ing *Ingester) Sync(ctx context.Context, peerID peer.ID, peerAddr multiadd
 		log.Debugw("Syncing advertisements up to latest", "adCid", c)
 		for {
 			select {
-			case adCid := <-syncDone:
-				log.Debugw("Synced advertisement", "adCid", adCid)
-				if adCid == c {
+			case adProcessedEvent := <-syncDone:
+				log.Debugw("Synced advertisement", "adCid", adProcessedEvent.adCid)
+				if adProcessedEvent.adCid == c || adProcessedEvent.err != nil && adProcessedEvent.headAdCid == c {
```
gammazero, line 308: Please add comment to explain why `adProcessedEvent.adCid` will not equal `c` on error.

gammazero (changes requested):
Mostly minor changes

internal/ingest/ingest.go
```diff
@@ -552,10 +567,17 @@ func (ing *Ingester) runIngestStep(syncFinishedEvent legs.SyncFinished) {
 	for _, c := range syncFinishedEvent.SyncedCids {
 		// Group the CIDs by the provider. Most of the time a publisher will only
 		// publish Ads for one provider, but it's possible that an ad chain can include multiple providers.
+
+		if ing.adHasBeenprocessed(c) {
```
gammazero, line 571:
```suggestion
		if ing.adAlreadyProcessed(c) {
```

internal/ingest/ingest.go
```diff
@@ -606,8 +628,7 @@ func (ing *Ingester) ingestWorkerLogic(msg toWorkerMsg) {
 	splitAtIndex := len(msg.adInfos)
 	for i, ai := range msg.adInfos {
 		// iterate latest to earliest
-		v, err := ing.ds.Get(context.Background(), datastore.NewKey(adProcessedPrefix+ai.cid.String()))
-		if err == nil && v[0] == byte(1) {
+		if ing.adHasBeenprocessed(ai.cid) {
```
gammazero, line 631:
```suggestion
		if ing.adAlreadyProcessed(ai.cid) {
```

gammazero (approved):

PR #259, opened 2022-03-12
Fair share ingest workers

MarcoPolo:
## Context 
Currently it's possible for a provider to claim all workers in the ingestion worker pool. Workers will not ingest ads from a provider concurrently (since they hold a lock for a provider). This means that all but one worker is stalled waiting for a lock to be released. This seriously hampers performance since it can prevent any parallel processing from happening.

## Proposed change

The change here is to introduce a new map of type `map[provider Peer.Id]atomic.Value`. atomic.Value is an atomic wrapped adChain. That represents the next queued up adChain that needs to be processed for a provider.

When a new adChain to be processed comes in, the ingest loop places it in that map. If there was no previous value, then it also schedules a worker to ingest that adChain. If there was a previous value, it is dropped (okay since it should be a subset of the new value) and we do not schedule a worker to ingest that adChain (since a previous run already scheduled a worker and we simply updated the data it should ingest).

## Tests

No new tests are added, but, if the reviewer would like, I'm happy to add a test case here.


MarcoPolo:
`Value.Swap` isn't defined in go 1.16. I guess I could change this to swap pointer :/

MarcoPolo:
Go 1.18 is expected [this month](https://tip.golang.org/doc/go1.18). Can we skip the go 1.16 tests until then? Since we won't have to maintain 1.16 after 1.18 is out.

gammazero:
Please rebase on main

MarcoPolo:
Rebased.

Will change to unsafe pointer tomorrow

internal/ingest/ingest.go
```diff
@@ -609,11 +617,21 @@ func (ing *Ingester) runIngestStep(syncFinishedEvent legs.SyncFinished) {
 		if _, ok := ing.providersBeingProcessed[p]; !ok {
 			ing.providersBeingProcessed[p] = &sync.Mutex{}
 		}
+		if _, ok := ing.providerAdChainStaging[p]; !ok {
+			ing.providerAdChainStaging[p] = &atomic.Value{}
+		}
 		ing.providersBeingProcessedMu.Unlock()
-		ing.toWorkers <- toWorkerMsg{
+
+		oldAssignment := ing.providerAdChainStaging[p].Swap(workerAssignment{
```
gammazero, line 625: Does this also need to be done while the `ing.providersBeingProcessedMu` lock is held?  Otherwise, there could be a read/write race with `ing.providerAdChainStaging[p] = &atomic.Value{}` on line 621.

internal/ingest/ingest.go
```diff
@@ -623,20 +641,29 @@ func (ing *Ingester) ingestWorker() {
...
 	// It's assumed that the runIngestStep puts a mutex in this map.
-	ing.providersBeingProcessed[msg.provider].Lock()
-	defer ing.providersBeingProcessed[msg.provider].Unlock()
+	ing.providersBeingProcessed[provider].Lock()
+	defer ing.providersBeingProcessed[provider].Unlock()
+
+	// Pull out the assignment for this provider. Note that runIngestStep populates this atomic.Value.
+	assignmentInterface := ing.providerAdChainStaging[provider].Swap(workerAssignment{none: true})
+	if assignmentInterface == nil || assignmentInterface.(workerAssignment).none {
+		// Nothing to do. Someone took our assignment.
```
gammazero, line 659: Please also explain when this could happen.  I am guessing that it can only happen with concurrent sync requests for the same provider and both requests get sent to `ing.toWorkers` before one of them is marked assigned.

gammazero (approved):
See comments.

This change should also include some documentation about the behavior, that a single publisher cannot starve other providers from getting their ad chains processed.

gammazero (changes requested):
Need go1.17 compatibility first.

internal/ingest/ingest.go
MarcoPolo: 621 only inserts an atomic value if there was no value set in the map. Once you have the atomic value you don't need to lock the map anymore since you are only dealing with references to the atomic value.

In other words we only set `ing.providerAdChainStaging[p] = &atomic.Value{}` once in the lock. Outside the lock we only deal with `*atomic.Value` so no need for a lock.

internal/ingest/ingest.go
MarcoPolo: I don't think this would happen in practice. It's here for completeness.

because of `atomic.Value` even concurrent ingest requests should be fine since we'll know either a worker picked up an assignment and the pending assignment is empty (in which case we properly schedule another one), _or_ we have a pending assignment that we swapped for a more recent assignment and a pending worker hasn't picked it up.

It's not possible to get into a state where two workers are queued on the same assignment, and one takes it and leaves the other with nothing to do.

Unless of course someone manually calls `ingestWorkerLogic` like in a test or something.

gammazero (approved):

gammazero:
Code looks good.  Can merge when tests pass (go1.16 compatability)

gammazero:
Since go1.18 is officially out, we can just skip the 1.16 tests in CI and wait until https://github.com/protocol/.github catches up.  This means that we should change `go.mod` to use `go 1.17`

PR #268, opened 2022-03-16
Use IPLD bindnode and reduce boilerplate code in ingest schema

masih:
Use IPLD bindnode to directly wrap and unwrap go structs in order to
reduce boilerplate code when encoding or decoding ingest types.

The codegen used before to generate IPLD schema types was also out of
date. Using bindnode has the added advantage that there is no need to
keep generated code up-to-date.


gammazero (commented):
This all looks good, but I do not think I am qualified to review this.  For example, I cannot say this change [here](https://github.com/filecoin-project/storetheindex/pull/268/files#diff-a441ac27969d6a7f7816d7e60c7ab93a54c0ce9c42c2f290db9a6c6996b87e0cR992-R996) is absolutely correct without learning more about IPLD schema generation, bindnode, etc.

masih:
Thanks for looking @gammazero. I have tagged Daniel as a review to check IPLD stuff.

api/v0/ingest/schema/schema.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// TypeAdvertisement represents the IPLD schema type for an advertisement.
+	//
+	// See: Advertisement.
+	TypeAdvertisement schema.Type
+
+	// AdvertisementPrototype represents the IPLD node prototype of Advertisement.
+	//
+	// See: bindnode.Prototype.
+	AdvertisementPrototype ipld.NodePrototype
```
mvdan, line 49: FYI these could be schema.TypedPrototype, which also lets you e.g. decode at the representation level. I actually imagine that's what you want, as otherwise you're decoding at the type level, ignoring e.g. representation strategies.

api/v0/ingest/schema/schema.go
```diff
@@ -0,0 +1,83 @@
...
+			Codec:    uint64(multicodec.DagJson),
+			MhType:   uint64(multicodec.Sha2_256),
+			MhLength: 16,
+		},
+	}
+
+	// TypeAdvertisement represents the IPLD schema type for an advertisement.
+	//
+	// See: Advertisement.
+	TypeAdvertisement schema.Type
```
mvdan, line 44: If you take the TypedPrototype suggestion above, then the type is reachable via the Type method - no need to export both.

api/v0/ingest/schema/schema.go
```diff
@@ -0,0 +1,83 @@
...
+	TypeEntryChunk schema.Type
+
+	// EntryChunkPrototype represents the IPLD node prototype of EntryChunk.
+	//
+	// See: bindnode.Prototype.
+	EntryChunkPrototype ipld.NodePrototype
+
+	//go:embed schema.ipldsch
+	schemaBytes []byte
+	typeSystem  *schema.TypeSystem
```
mvdan, line 64: nit: I would move this inside `func init`, as otherwise you keep it in scope forever unnecessarily

api/v0/ingest/schema/types.go
```diff
@@ -0,0 +1,178 @@
...
+			switch x := r.(type) {
+			case string:
+				err = errors.New(x)
+			case error:
+				err = x
+			default:
+				err = fmt.Errorf("unknown panic: %v", r)
+			}
+		}
+	}()
```
mvdan, line 54: I would hope that you can remove this recover at some point, once your code and bindnode is more stable :) perhaps add a TODO for that?

it should only panic if there's a bug in bindnode, or if you put values inside `a` which are incompatible with the schema, e.g. an union with no value set (which is disallowed).

api/v0/ingest/schema/types.go
```diff
@@ -0,0 +1,178 @@
...
+	n = bindnode.Wrap(&a, TypeAdvertisement).Representation()
+	return
+}
+
+// FromNode populates the fields of this advertisement from the given node.
+func (a *Advertisement) FromNode(node ipld.Node) error {
+	if node.Prototype() != AdvertisementPrototype {
```
mvdan, line 61: When do you expect this to kick in? I would document it.

If it's kicking in often, then I imagine something's not right - perhaps your link loading stuff is using `basicnode.Prototype.Any` when following links, and then you're doing the extra work of converting the untyped node to Advertisement here. At the very least this needs a TODO to avoid the duplicate work at some point in the future.

api/v0/ingest/schema/types.go
```diff
@@ -0,0 +1,178 @@
...
+		return fmt.Errorf("unwrapped node does not match schema.Advertisement")
+	}
+	a.PreviousID = other.PreviousID
+	a.Provider = other.Provider
+	a.Addresses = other.Addresses
+	a.Signature = other.Signature
+	a.Entries = other.Entries
+	a.ContextID = other.ContextID
+	a.Metadata = other.Metadata
+	a.IsRm = other.IsRm
```
mvdan, line 81: is there a particular reason you want to copy the fields rather than returning or using `other` directly?

api/v0/ingest/schema/types.go
```diff
@@ -0,0 +1,178 @@
...
+			switch x := r.(type) {
+			case string:
+				err = errors.New(x)
+			case error:
+				err = x
+			default:
+				err = fmt.Errorf("unknown panic: %v", r)
+			}
+		}
+	}()
```
mvdan, line 134: it's the second time I see this, so probably move it to a `func safeWrap`

api/v0/ingest/schema/types.go
```diff
@@ -0,0 +1,178 @@
...
+// FromNode populates the fields of this entry chunk from the given node.
+func (e *EntryChunk) FromNode(node ipld.Node) error {
+	if node.Prototype() != EntryChunkPrototype {
+		adBuilder := EntryChunkPrototype.NewBuilder()
+		err := adBuilder.AssignNode(node)
+		if err != nil {
+			return fmt.Errorf("faild to convert node prototype: %w", err)
+		}
+		node = adBuilder.Build()
+	}
```
mvdan, line 148: ditto for the TODO

api/v0/ingest/schema/types_test.go
```diff
@@ -0,0 +1,156 @@
...
+	Addresses List_String
+	Signature Bytes
+	Entries Link
+	ContextID Bytes
+	Metadata Bytes
+	IsRm Bool
+}
+
+type Link_Advertisement &Advertisement
+`))
```
mvdan, line 42: nit: I would add two levels of indentation to the schema string, which makes it easier to read the code, and still works as the schema DSL doesn't care about whitespace like that.

api/v0/ingest/schema/types_test.go
```diff
@@ -0,0 +1,156 @@
...
+	require.NoError(t, err)
+
+	rng := rand.New(rand.NewSource(1413))
+	ad := generateAdvertisement(rng)
+	oldAdType := oldSchema.TypeByName("Advertisement")
+	nodeFromOldAdType := bindnode.Wrap(&ad, oldAdType).Representation()
+
+	var newAd stischema.Advertisement
+	err = newAd.FromNode(nodeFromOldAdType)
```
mvdan, line 51: following my previous "why does FromNode copy the fields" comment - an alternative would be easier to use as well, e.g.:

    newAd, err := UnwrapAdvertisement(nodeFromOldAdType)

api/v0/ingest/schema/types.go
```diff
@@ -0,0 +1,178 @@
...
+		Addresses  []string
+		Signature  []byte
+		Entries    ipld.Link
+		ContextID  []byte
+		Metadata   []byte
+		IsRm       bool
+	}
+	EntryChunk struct {
+		Entries []multihash.Multihash
+		Next    *ipld.Link
```
mvdan, line 25: I know this pointer to interface feels weird, because an interface can already be nil, which should be enough for optional/nullable. See https://github.com/ipld/go-ipld-prime/issues/378 FYI

mvdan (approved):
SGTM, I'd just recommend avoiding these repeated methods on types if possible, as I think using bindnode directly should be pretty straightforward.

api/v0/ingest/schema/types.go
masih: The linksystem is passed around to other libraries like go-legs which might load things using `Any` and I don't have control over that unfortunately. All code in this repo, however uses appropriate prototype when loading.

mvdan: That's fine, I would just document it then. I think @willscott, you, and I could briefly chat about this at some point, so that we can properly plumb a NodePrototypeChooser through libraries like go-legs.

api/v0/ingest/schema/types.go
masih: The idea is to have a type-agnostic interface: i.e. `FromNode(ipld.Node) error`.

mvdan: What does that actually gain you, though? Constructors in Go generally don't fit into interfaces, and that's fine. You reach for `bytes.NewBuffer`, not `var buf bytes.Buffer; buf.FromBytes(...)`.

If you want to deduplicate code in the downstream packages, I would suggest having a helper function that works at a higher level, and doing away with the interface entirely. For example:

```
func UnwrapNode(proto schema.TypedPrototype, input ipld.Node) (interface{}, error)
```

This function could do essentially the same as you do now: AssignNode if the prototype doesn't match, otherwise do the Unwrap directly. You do have the `interface{}` return, but you have to deal with bindnode's `interface{}` at some point, given that we can't use generics just yet.

masih: > What does that actually gain you, though?

Here it's a total YAGNI no doubt about it :) 

I wanted to avoid casting by keeping the function signature agnostic of the schema types. 
You are right. I'll have a play around to see how much I can remove here.

api/v0/ingest/schema/types.go
masih: Great to see that issue captured 👍 In the meantime, I have to use pointer to interfaces right?

mvdan: Yup.

api/v0/ingest/schema/types_test.go
masih: Do you think keeping `LoadFrom` and `StoreTo` functions as receivers of schema types makes sense? That's also a YAGNI here, and the only reason I wrote it this way is the TODO I left about potentially proposing the interfaces to be part of IPLD.

api/v0/ingest/schema/types.go
masih: I am torn on this. Here we don't construct, we populate. And that's similar to what a lot of the interfaces in go SDK do, like `ReaderFrom` or `UnmarshalBinary` and other similar encoding interfaces. 

Changing this would also mean I need to get rid of the `LoadFrom` and `StoreTo` receivers on types and have package level helper functions instead. 

Either works. From your perspective, do you still think helper functions for all of it makes more sense?

masih: Having looked at the code again I think those interfaces are an overkill here. Ripped them all out, to avoid further bike shedding by yours truly 😊 

Thank you 🍻

mvdan: I'm late to respond here, but methods vs constructors and whether to not use an interface is a bit of a "it depends on the situation".

For example, ReadFrom or UnmarshalBinary do directly modify the receiver in-place. They also need to be an interface, so that custom user types satisfy it the way they want.

The reason I don't think a method fits you is:

1) You are unwrapping into a new value, then copying into the receiver value. So you're doing extra work.
2) You're satisfying an interface that doesn't _really_ need to be there, other than to scratch an itch for abstraction layers :)

mvdan: I should clarify that perhaps an interface would be nice if each type implemented the methods in a different way, but as far as I can see, they would all implement them the same. Which is why I suggest a top-level helper func that does the same for each type, as opposed to each type reimplementing the same logic.

masih: Agreed; thank you for your patience with me sir 🍻

masih:
@gammazero  that's all Daniel's review comments addressed, and branch rebased from `main`. 

This is ready for another look whenever you get the chance.

MarcoPolo (approved):
over 10:1 delete to add ratio. Nice!

PR #295, opened 2022-03-29
Install `nginx` ingress controller on dev K8S cluster

masih:
## Context
Install `nginx` ingress controller onto dev K8S cluster using an overlay
on top of base `ingress-nginx` kustomization.

Note that this should get automatically installed once merged to main now that we have a CD pipeline running for cluster level changes.

## Proposed Changes
See context.

## Tests
N/A

## Revert Strategy
`git revert`, commit and merge to `main`. CD would then remove it from dev cluster.


masih:
Cc @LexLuthr

Issue #287, opened 2022-03-24
Move service configuration to gitops


davidd8:
This is based on FluxCD and will help simplify production configuration management.

masih:
Done

PR #297, opened 2022-03-29
Deploy storetheindex to dev cluster

masih:
## Context
Define `kustomization` base for storetheindex as a statefulset along with overlay that deploys it to the dev cluster. This work also adds encrypted secret management via [sops](https://github.com/mozilla/sops) and KMS to securely check in encrypted secrets into code, and have the CD pipeline decrypt them on the fly by assuming roles that permit it to do so.

## Proposed Changes
* Define base storetheindex manifest overlay
* Define dev concrete overlay to deploy storetheindex
* Set up secret management and generate a random libp2p identity for the dev instance  

## Tests
N/A

## Revert Strategy
`git revert` then `terraform apply` for infrastructure part.

Relates to: #287

### DIFF
--- a/deploy/infrastructure/dev/us-east-2/versions.tf
+++ b/deploy/infrastructure/dev/us-east-2/versions.tf
@@ -1,3 +1,3 @@
 terraform {
-  required_version = ">= 1.13.7"
+  required_version = ">= 1.1.7"
 }
--- a/deploy/manifests/base/storetheindex/config
+++ b/deploy/manifests/base/storetheindex/config
@@ -1,0 +1,61 @@
+{
+  "Identity": {
+    "PeerID": "",
+    "PrivKey": ""
+  },
+  "Addresses": {
+    "Admin": "/ip4/0.0.0.0/tcp/3002",
+    "Finder": "/ip4/0.0.0.0/tcp/3000",
+    "Ingest": "/ip4/0.0.0.0/tcp/3001",
+    "P2PAddr": "/ip4/0.0.0.0/tcp/3003",
+    "NoResourceManager": true
+  },
+  "Bootstrap": {
+    "Peers": [
+      "/dns4/bootstrap-0.ipfsmain.cn/tcp/34721/p2p/12D3KooWQnwEGNqcM2nAcPtRR9rAX8Hrg4k9kJLCHoTR5chJfz6d",
+      "/dns4/bootstrap-3.mainnet.filops.net/tcp/1347/p2p/12D3KooWKhgq8c7NQ9iGjbyK7v7phXvG6492HQfiDaGHLHLQjk7R",
+      "/dns4/bootstrap-4.mainnet.filops.net/tcp/1347/p2p/12D3KooWL6PsFNPhYftrJzGgF5U18hFoaVhfGk7xwzD8yVrHJ3Uc",
+      "/dns4/bootstrap-6.mainnet.filops.net/tcp/1347/p2p/12D3KooWP5MwCiqdMETF9ub1P3MbCvQCcfconnYHbWg6sUJcDRQQ",
+      "/dns4/bootstrap-7.mainnet.filops.net/tcp/1347/p2p/12D3KooWRs3aY1p3juFjPy8gPN95PEQChm2QKGUCAdcDCC4EBMKf",
+      "/dns4/bootstrap-8.mainnet.filops.net/tcp/1347/p2p/12D3KooWScFR7385LTyR4zU1bYdzSiiAb5rnNABfVahPvVSzyTkR",
+      "/dns4/node.glif.io/tcp/1235/p2p/12D3KooWBF8cpp65hp2u9LK5mh19x67ftAam84z9LsfaquTDSBpt",
+      "/dns4/bootstrap-2.mainnet.filops.net/tcp/1347/p2p/12D3KooWEWVwHGn2yR36gKLozmb4YjDJGerotAPGxmdWZx2nxMC4",
+      "/dns4/bootstrap-5.mainnet.filops.net/tcp/1347/p2p/12D3KooWLFynvDQiUpXoHroV1YxKHhPJgysQGH2k3ZGwtWzR4dFH",
+      "/dns4/bootstrap-1.starpool.in/tcp/12757/p2p/12D3KooWQZrGH1PxSNZPum99M1zNvjNFM33d1AAu5DcvdHptuU7u",
+      "/dns4/bootstrap-0.mainnet.filops.net/tcp/1347/p2p/12D3KooWCVe8MmsEMes2FzgTpt9fXtmCY7wrq91GRiaC8PHSCCBj",
+      "/dns4/bootstrap-1.mainnet.filops.net/tcp/1347/p2p/12D3KooWCwevHg1yLCvktf2nvLu7L9894mcrJR4MsBCcm4syShVc",
+      "/dns4/bootstrap-0.starpool.in/tcp/12757/p2p/12D3KooWGHpBMeZbestVEWkfdnC9u7p6uFHXL1n7m1ZBqsEmiUzz",
+      "/dns4/lotus-bootstrap.ipfsforce.com/tcp/41778/p2p/12D3KooWGhufNmZHF3sv48aQeS13ng5XVJZ9E6qy2Ms4VzqeUsHk",
+      "/dns4/bootstrap-1.ipfsmain.cn/tcp/34723/p2p/12D3KooWMKxMkD5DMpSWsW7dBddKxKT7L2GgbNuckz9otxvkvByP"
+    ],
+    "MinimumPeers": 15
+  },
+  "Datastore": {
+    "Type": "levelds",
+    "Dir": "/data/datastore"
+  },
+  "Discovery": {
+    "LotusGateway": "wss://api.chain.love",
+    "Policy": {
+      "Allow": true,
+      "Except": null,
+      "Trust": true,
+      "TrustExcept": null
+    },
+    "PollInterval": "24h0m0s",
+    "PollRetryAfter": "5h0m0s",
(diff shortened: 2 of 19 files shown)


deploy/manifests/base/storetheindex/config
```diff
@@ -0,0 +1,61 @@
...
+      "/dns4/bootstrap-2.mainnet.filops.net/tcp/1347/p2p/12D3KooWEWVwHGn2yR36gKLozmb4YjDJGerotAPGxmdWZx2nxMC4",
+      "/dns4/bootstrap-5.mainnet.filops.net/tcp/1347/p2p/12D3KooWLFynvDQiUpXoHroV1YxKHhPJgysQGH2k3ZGwtWzR4dFH",
+      "/dns4/bootstrap-1.starpool.in/tcp/12757/p2p/12D3KooWQZrGH1PxSNZPum99M1zNvjNFM33d1AAu5DcvdHptuU7u",
+      "/dns4/bootstrap-0.mainnet.filops.net/tcp/1347/p2p/12D3KooWCVe8MmsEMes2FzgTpt9fXtmCY7wrq91GRiaC8PHSCCBj",
+      "/dns4/bootstrap-1.mainnet.filops.net/tcp/1347/p2p/12D3KooWCwevHg1yLCvktf2nvLu7L9894mcrJR4MsBCcm4syShVc",
+      "/dns4/bootstrap-0.starpool.in/tcp/12757/p2p/12D3KooWGHpBMeZbestVEWkfdnC9u7p6uFHXL1n7m1ZBqsEmiUzz",
+      "/dns4/lotus-bootstrap.ipfsforce.com/tcp/41778/p2p/12D3KooWGhufNmZHF3sv48aQeS13ng5XVJZ9E6qy2Ms4VzqeUsHk",
+      "/dns4/bootstrap-1.ipfsmain.cn/tcp/34723/p2p/12D3KooWMKxMkD5DMpSWsW7dBddKxKT7L2GgbNuckz9otxvkvByP"
+    ],
+    "MinimumPeers": 4
```
gammazero, line 31: Should this be 15 so that the indexer tries to stay connected to them all?

deploy/manifests/base/storetheindex/indexer-statefulset.yaml
```diff
@@ -0,0 +1,50 @@
...
+            name: indexer-config
+  volumeClaimTemplates:
+    - metadata:
+        name: data
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 2Ti
```
gammazero, line 50: Seems like this will be a problem given that we already have 2.5Ti of index data?  Seems like this should be at least 4Ti, probably more.

deploy/manifests/base/storetheindex/config
```diff
@@ -0,0 +1,61 @@
+{
+  "Identity": {
+    "PeerID": "",
+    "PrivKey": ""
+  },
+  "Addresses": {
+    "Admin": "/ip4/0.0.0.0/tcp/3002",
```
gammazero, line 7: Do we want to expose this externally, or keep it on localhost and require exec into pod?

deploy/manifests/base/storetheindex/ingest-service.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    app: ingest
+spec:
+  ports:
+    - name: ingest
+      port: 3001
+      targetPort: ingest
+  selector:
+    app: indexer
+  type: LoadBalancer
```
gammazero, line 14: If we do not expose the admin port, then this is certainly not needed.  Even if we do expose the admin port, do we want that going through the load balancer since it is only used to handle a few specific commands?

deploy/manifests/dev/us-east-2/tenant/storetheindex/kustomization.yaml
```diff
@@ -3,3 +3,12 @@ kind: Kustomization
...
   - ../../../../base/storetheindex
+
+patchesStrategicMerge:
+  - patch-indexer.yaml
+
+configMapGenerator:
+  - name: indexer-env-vars
+    behavior: merge
+    literals:
+      - STORETHEINDEX_PRIV_KEY_PATH=/identity/identity.key
```
gammazero, line 14: What creates that key file in the first place?

gammazero (commented):
Have questions and comments.

deploy/manifests/base/storetheindex/config
masih: This is copied from prod. Happy to change it to whatever you see fit.

Is 15 what it should be?

deploy/manifests/base/storetheindex/config
masih: Two things here: 
1. binding to `0.0.0.0` does not mean it is exposed externally. This is only accessible within k8s.
2.This is how it's configured in prod.

deploy/manifests/dev/us-east-2/tenant/storetheindex/kustomization.yaml
masih: I have generated it manually, encrypted it and checked it into code as a secret. 
That secret is then mounted as a volume on `/identity` path.

deploy/manifests/base/storetheindex/config
masih: Updated to 15 here. I suggest we check the value of this config in prod since it is indeed set to 4.

deploy/manifests/base/storetheindex/ingest-service.yaml
gammazero: Comment was referring to admin port above.  NM if not exposed.

gammazero (approved):

deploy/infrastructure/dev/us-east-2/versions.tf
```diff
@@ -1,3 +1,3 @@
 terraform {
-  required_version = ">= 1.13.7"
```
MarcoPolo, line 2: Typo?

masih: yup; I must have fat fingered it in previous PRs and not noticed :)

deploy/manifests/base/storetheindex/kustomization.yaml
```diff
@@ -3,11 +3,25 @@ kind: Kustomization
...
 configMapGenerator:
-  - name: storetheindex-config
+  - name: indexer-env-vars
     behavior: create
     literals:
       - GOLOG_LOG_LEVEL=INFO
+      - GOLOG_LOG_FMT=json
+      - STORETHEINDEX_LOTUS_GATEWAY=wss://api.chain.love
+      - STORETHEINDEX_PATH=/config
+      - STORETHEINDEX_LISTEN_ADMIN=/ip4/0.0.0.0/tcp/3002
```
MarcoPolo, line 22: Is this used? I thought this was set in the config?

deploy/infrastructure/dev/us-east-2/versions.tf
masih: That's defining the minimum version accepted to interact with terraform files here. It is set to the [lates version](https://www.terraform.io/downloads). What is the issue with using the latest version? 

Re change in this PR, I accidentally must have added a `3` in there, which I removed in this PR, i.e. the fat fingering.

MarcoPolo: Oh! my mistake sorry I missed that terraform is already at 1.x. I was thinking 0.15 was the latest. Ignore me :)

deploy/manifests/base/storetheindex/indexer-statefulset.yaml
```diff
@@ -0,0 +1,50 @@
...
+            - name: config
+              mountPath: /config
+      volumes:
+        - name: config
+          configMap:
+            name: indexer-config
+  volumeClaimTemplates:
+    - metadata:
+        name: data
+      spec:
```
MarcoPolo, line 45: just a note, for the production instances we probably want to define a new storage class that uses io2

deploy/manifests/base/storetheindex/kustomization.yaml
masih: Added it here since it overrides the default. It also gets overridden in docker file for some reason. and yes it is also set in config.  I can't remember which gets picked up in the end. so I set them. happy to remove redundant ones.

deploy/manifests/base/storetheindex/indexer-statefulset.yaml
masih: `gp2` is what we are using by EKS default

deploy/manifests/base/storetheindex/kustomization.yaml
MarcoPolo: I'd say yes to remove the redundant ones, since it's not clear which ones get set (env should always have priority fwiw).

> since it overrides the default.

Isn't 3002 the default anyways?

masih: 3002 yes but not `0.0.0.0`.

deploy/manifests/base/storetheindex/indexer-statefulset.yaml
MarcoPolo: yep. I'm saying `io2` will give us better perf in prod. Maybe even worth doing it in dev so we get the same performance characteristics.

deploy/manifests/base/storetheindex/kustomization.yaml
MarcoPolo: ah I see. Thank you

deploy/manifests/base/storetheindex/indexer-statefulset.yaml
masih: I see. I need to look into what it takes to get that working on k8s

Issue #257, opened 2022-03-11
Checking ad provider ID matches signer peer ID is never false if signature verification passes


masih:
A [check](https://github.com/filecoin-project/storetheindex/blob/9d30817b79093fc0b86f7c9b851f473b9e7c8c5e/internal/ingest/linksystem.go#L124) is performed to assert that the signer of an envelop matches the provider of an advertisement.

But, the signature verification logic [will fail ](https://github.com/filecoin-project/storetheindex/blob/main/api/v0/ingest/schema/envelope.go#L154) otherwise anyway because it requires `advID` across envelops to match, and the ad [provider ID is included](https://github.com/filecoin-project/storetheindex/blob/main/api/v0/ingest/schema/envelope.go#L76) in the `advID` calculation. This makes the check referenced above to be always false, since a verified signature by definition must mean provider IDs must have matched.

Consider, either removing that check completely or relax the need for  `provider ID`s to match across envelops during signature verification.

gammazero:
The [`VerifySignature`](https://github.com/filecoin-project/storetheindex/blob/9d30817b79093fc0b86f7c9b851f473b9e7c8c5e/api/v0/ingest/schema/envelope.go#L125) checks that the advertisement was signed by the private key that matches the public key that is included in the envelope.  However, an advertisement could be constructed where this public key is different than the provider ID in the advertisement.  Therefore it is still necessary to check that the identity (public key) used to sign the advertisement is the ID that we wanted to sign the advertisement, which currently is the provider ID.

masih:
That makes sense thanks @gammazero  🍻

PR #299, opened 2022-03-31
Add comments and tests for ad signature

gammazero:
Added comments to document that the advertisement does not have to be signed by the provider, and a test to show this.

This addresses #257

## Context
Demonstrates that it is still necessary to check that the signer of an advertisement is allowed to sign the advertisement.

## Proposed Changes
- Added comments
- Added test to show valid signature on advertisement, when not signed by provider.

## Tests
- Show valid signature on advertisement, when not signed by provider

## Revert Strategy
Change is safe to revert.


masih (approved):
Thank you for this; that clears up my confusion 🍻

gammazero:
> We are already checking that the signerID is equal to the provider ID, correct? So this is just documentation then.

Yes, just doc.

PR #300, opened 2022-03-31
Set up `dev.cid.contact` with all things manged by K8S

masih:
## Context
Configure `dev.cid.contact` as the HTTP endpoint for the dev instance.

Install controllers to manage everythign to make that possible doable
via GitOps in K8S; that is: cert-manager to generate certificates,
external-dns to handle nameserver changes and ingress controller to
manage load balancing HTTP routing.

## Proposed Changes
Install:
* cert-manager to issue valid certificates for public dev endpoint
* external-dns to handle name server changes via k8s 
* nginx-ingress to manage K8S ingress, load balancing and HTTP content routing
* ...and finally define indexer ingress with TLS enabled accessible via `dev.cid.contact`
* profit 💸 

While at it, simplify the service structure in base `storetheindex` by removing, loadbalancers now that ingress handles it all, combining service objects and change to ip type none to delegate load balancing choice entirely to ingress.

## Tests
Tested by manually deploying to dev. Currently [running](https://dev.cid.contact/providers)

## Revert Strategy
`git revert` `terraform apply`

Relates to #287


MarcoPolo:
I don't think you want to run `terraform destroy` when you revert this.

masih:
> I don't think you want to run `terraform destroy` when you revert this.

Fixed

I am not sure the new PR template entirely makes sense. It seems superfluous at times. Having said that, I am too lazy to think of an alternative 🤐

MarcoPolo:
Maybe! happy to change this to a better one. The key thing is that when there's an issue it should be straightforward to understand how to undo a change without having to re-review the PR. 

Maybe this field should be called "Safe to revert?" and you can simply say yes if it's a simple git revert and no if there are some subtleties when reverting.

deploy/manifests/dev/us-east-2/cluster/external-dns/deploy-patch.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - name: external-dns
+          args:
+            - --source=service
+            - --source=ingress
+            - --domain-filter=dev.cid.contact
+            - --provider=aws
+            - --policy=upsert-only
+            - --aws-zone-type=public
+            - --registry=txt
+            - --txt-owner-id=Z0346653G546POP574CS
```
MarcoPolo, line 27: Where does this come from?

masih: [Here](https://github.com/filecoin-project/storetheindex/blob/2bb47c38c3ca0200e4da9605a1c7c3f8b03399cc/deploy/infrastructure/dev/us-east-2/route53.tf).

MarcoPolo: From a terraform output?
Maybe a comment that says something to that effect? Something like
> Value is from `terraform output <name>`

Issue #286, opened 2022-03-24
Set up a cloudfront caching layer for web traffic


davidd8:
Adding a caching layer will help the Network Indexer scale to over 30K qps based on load tests and should help handle its current load volume from IPFS.

masih:
I believe this is done in fil-managed prod.

This is also done for both `dev` and `prod` instances deployed on K8S clusters managed by us.

PR #301, opened 2022-03-31
Set up caching layer for dev endpoint using CloudFront

masih:
## Context
Hitting `dev.cid.contact` makes a call to indexer every time. Considering the nature of data being served, this PR sets up caching using CloudFront accessible via `cdn.dev.cid.contact`

## Proposed Changes
Define:
* CloudFront distribution for dev
* New viewer certificate in `us-east-1`; existing cert used by ingress only lives in K8S.
* `A` records for `cdn.dev.cid.contact`

## Tests
```sh
$ curl -kIs https://cdn.dev.cid.contact | grep x-cache
x-cache: Hit from cloudfront
```

## Revert Strategy
`git revert` then `terraform apply`

Relates to: #286

### DIFF
--- a/deploy/infrastructure/dev/us-east-2/couldfront.tf
+++ b/deploy/infrastructure/dev/us-east-2/couldfront.tf
@@ -1,0 +1,99 @@
+locals {
+  cdn_origin_id = "${local.environment_name}_sti_cdn"
+  cdn_subdomain = "cdn"
+}
+
+resource "aws_cloudfront_distribution" "cdn" {
+  enabled = true
+
+  aliases     = ["${local.cdn_subdomain}.${aws_route53_zone.dev_external.name}"]
+  price_class = "PriceClass_All"
+
+  origin {
+    domain_name = aws_route53_zone.dev_external.name
+    origin_id   = local.cdn_origin_id
+    custom_origin_config {
+      http_port              = 80
+      https_port             = 443
+      origin_protocol_policy = "https-only"
+      origin_ssl_protocols   = ["SSLv3", "TLSv1", "TLSv1.1", "TLSv1.2"]
+    }
+    origin_shield {
+      enabled              = true
+      origin_shield_region = data.aws_region.current.name
+    }
+  }
+
+  default_cache_behavior {
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD"]
+    target_origin_id = local.cdn_origin_id
+
+    forwarded_values {
+      query_string = false
+
+      cookies {
+        forward = "none"
+      }
+    }
+    compress               = true
+    viewer_protocol_policy = "redirect-to-https"
+    min_ttl                = 0
+    default_ttl            = 3600
+    max_ttl                = 86400
+  }
+
+  restrictions {
+    geo_restriction {
+      restriction_type = "none"
+    }
+  }
+  viewer_certificate {
+    acm_certificate_arn = module.cdn_cert.acm_certificate_arn
+    ssl_support_method  = "sni-only"
+
+  }
+
+  tags = local.tags
+}
+
+provider "aws" {
+  alias  = "use1"
+  region = "us-east-1"
+}
+
+module "cdn_cert" {
+  source  = "registry.terraform.io/terraform-aws-modules/acm/aws"
+  version = "3.4.0"
+
+  #  Certificate must be in us-east-1 as dictated by CloudFront
+  providers = {
+    aws = aws.use1
+  }
+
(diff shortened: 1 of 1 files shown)


MarcoPolo:
> For CloudFront to work, what it points to must be public I believe.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/restrict-access-to-load-balancer.html

masih:
Thanks for the link @MarcoPolo yeah that should do it. ELB will still remain exposed from network routing perspective.

Moving the cache into K8S would allow us to tighten this up and have more control over it. But hey, it's probably a YAGNI for now

masih:
Another approach might be to maybe whitelist only [cloudfront](https://ip-ranges.amazonaws.com/ip-ranges.json) IP ranges in the ingress if we want to stick with cloudfront.

deploy/infrastructure/dev/us-east-2/couldfront.tf
```diff
@@ -0,0 +1,99 @@
...
+
+  domain_name               = aws_route53_zone.dev_external.name
+  zone_id                   = aws_route53_zone.dev_external.zone_id
+  subject_alternative_names = ["*.${aws_route53_zone.dev_external.name}"]
+
+  tags = local.tags
+}
+
+module "records" {
+  source  = "registry.terraform.io/terraform-aws-modules/route53/aws//modules/records"
```
gammazero, line 82: Extra `/` at `aws//modules`

deploy/infrastructure/dev/us-east-2/couldfront.tf
```diff
@@ -0,0 +1,99 @@
...
+  version = "2.6.0"
+
+  zone_id = aws_route53_zone.dev_external.zone_id
+
+  records = [
+    {
+      name = local.cdn_subdomain
+      type = "A"
```
MarcoPolo, line 90: Should this be an alias record?
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-to-cloudfront-distribution.html
https://www.bluematador.com/docs/troubleshooting/route53-hosted-zone

MarcoPolo: also make me wonder if the record for the LB should also be an alias

deploy/infrastructure/dev/us-east-2/couldfront.tf
```diff
@@ -0,0 +1,99 @@
...
+
+module "records" {
+  source  = "registry.terraform.io/terraform-aws-modules/route53/aws//modules/records"
+  version = "2.6.0"
+
+  zone_id = aws_route53_zone.dev_external.zone_id
+
+  records = [
+    {
+      name = local.cdn_subdomain
```
gammazero, line 89: I do not understand what the purpose of this A-record is, so I am not sure if "local.x" is required or if it can be something more descriptive.

MarcoPolo (changes requested):

gammazero (commented):
Has a couple of comments, but looks OK.  This PR should get approval from someone who is more familiar with CloudFront and can determine if anything is missing.

deploy/infrastructure/dev/us-east-2/couldfront.tf
masih: This A records specifies `cdn.dev.cid.contact` should go to the host name created by cloudfront. The `local` directive defines a variable for `cdn` portion to make sure the subdomain we set here is the same as the prefix of alias defined in cloudfron distribution [here](https://github.com/filecoin-project/storetheindex/pull/301/files#diff-9b40fc774548f654a662e1eaae3eb8b063d22c1034c8efd68a58da382d9fa685R9). The name of that variable is `cdn_subdomain` which to me is quite descriptive :)

Do you have alternative suggestions for the name of that variable?

deploy/infrastructure/dev/us-east-2/couldfront.tf
masih: The LB records are created automatically using `external-dns` controller. I'll take a look in settings to see if we can change that or indeed understand why it's using `A` records.

masih: OK; looks like `A` and `AAAA` records are indeed alias records in route53. The [guide](url) you kindly linked above it also instructs to create `A` record:

> Record type
>   Choose A – IPv4 address.
>   If IPv6 is enabled for the distribution and you're creating a second record, choose AAAA – IPv6 address.

Resolving; please shout if I have missed anything.

MarcoPolo (approved):

PR #308, opened 2022-04-01
Set up cluster-autoscaler in `dev`

masih:
## Context
Use [autoscaler](https://github.com/kubernetes/autoscaler) to automatically manage EC2 instances in `dev` EKS cluster node group depending on the workload that's running.

**Note: to avoid destroying resources already applied in #301, that PR is merged into this PR.**

## Proposed Changes
* Add base layer for `cluster-autoscaler`
* Configure roles necessary for the autoscaler to do its job.
* Define overlay in dev cluster for autoscaler, assign the role to its service account and make configuration changes specific to dev for the auto-scaler deployment object

## Tests
N/A

## Revert Strategy
`git revert` then `terraform apply`

### DIFF
--- a/deploy/manifests/base/cluster-autoscaler/kustomization.yaml
+++ b/deploy/manifests/base/cluster-autoscaler/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - https://raw.githubusercontent.com/kubernetes/autoscaler/cluster-autoscaler-chart-9.16.2/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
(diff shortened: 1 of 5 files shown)


masih:
Please review after #301 is merged.

MarcoPolo:
we should rebase this PR I believe

deploy/manifests/base/cluster-autoscaler/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - https://raw.githubusercontent.com/kubernetes/autoscaler/master/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
```
MarcoPolo, line 5: Should this be using a specific release rather than the head of master?

MarcoPolo (approved):
I'm unfamiliar with this setup, but the changes seem fine. Just one question about using a specific version of the dependency rather than the head of master.

PR #312, opened 2022-04-01
Add cache control header to http Finder

MarcoPolo:
## Context
We aren't caching 404s with CloudFront yet. I think it's okay to cache 404s since if we don't have it now we also likely still won't have it if another caller asks us immediately afterwards.

## Proposed Changes
This adds the [cache control header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) to the finder HTTP endpoint. This should allow Cloudfront to cache our 404s to the same CID. See [How CloudFront processes and caches HTTP 4xx and 5xx status codes from your origin](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/HTTPStatusCodes.html)

Note that this sets the default cache age to be 30 minutes. I think this okay since we don't have an expectation of data being available immediately after ingest. We could of course tweak this later if we notice this becoming a pain point.

## Tests

```
❯ curl 'http://127.0.0.1:3000/cid/bafybeigvgzoolc3drupxhlevdp2ugqcrbcsqfmcek2zxiw5wctk3xjpjwy' -sI | grep Cache-Control
Cache-Control: max-age=1800
```

## Revert Strategy

Change is safe to revert.

### DIFF
--- a/server/finder/http/options.go
+++ b/server/finder/http/options.go
@@ -8,6 +8,7 @@ import (
 const (
 	apiWriteTimeout = 30 * time.Second
 	apiReadTimeout  = 30 * time.Second
+	cacheMaxAge     = 30 * time.Minute
 	maxConns        = 8_000
 )
 
@@ -15,6 +16,7 @@ const (
 type serverConfig struct {
 	apiWriteTimeout time.Duration
 	apiReadTimeout  time.Duration
+	cacheMaxAge     time.Duration
 	maxConns        int
 }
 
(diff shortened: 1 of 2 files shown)


willscott (approved):
we probably want negative caches to be smaller at least initially - if someone's racing between trying to find the file they just stored and loading from the indexer and queries before we get the announcement it would be a shame to force them to wait 30 min to refresh and find it.

MarcoPolo:
hmm actually I think cloudfront should do this by just setting minimum TTL. Investigating...

MarcoPolo:
Yep we can do this with this setting in cloudfront (terraform representation):
```
  custom_error_response {
    error_code            = 404
    error_caching_min_ttl = 300
  }

```

Issue #318, opened 2022-04-04
Build and push images to ECR tagged with server


masih:
To utilise the `image-automation` and `image-reflector` in flux CD we need to tag images with semver. 

The current pushed images are tagged with commit ID of a tagged version. 

- [ ] Publish images tagged with somatic version. 
- [ ] Publish images to ECR.

masih:
Fixed by https://github.com/filecoin-project/storetheindex/pull/346

PR #319, opened 2022-04-04
Push additional docker images with git tag as docker tag

masih:
## Context
In order to have semantically versioned docker containers, tag images
with git tag as well as commit ID and push to docker hub as normal.

Relates to: #318

## Proposed Changes
Tag docker images with `git tag` and push in addition to existing commit-id tagged images.

## Tests
N/A

## Revert Strategy
`git revert`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,8 +20,9 @@ jobs:
             DOCKER_TAG=$CIRCLE_SHA1
             echo "export DOCKER_REPO=$DOCKER_REPO" >> $BASH_ENV
             echo "export DOCKER_TAG=$DOCKER_TAG" >> $BASH_ENV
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -72,6 +73,9 @@ workflows:
           context:
             - filecoin-dockerhub-publish
           filters:
+            # Do not run on branch builds
+            branches:
+              ignore: /.*/
```
masih, line 78: Looks like this job was running for every branch which I believe was not intended @coryschwartz ?

The change here will only run the job on semver tagged commits.

masih:
No longer needed in light of https://github.com/filecoin-project/storetheindex/pull/346

PR #320, opened 2022-04-04
Fix out of bounds panic in registery

gammazero:
When registering or updating the registry with provider information that does not have provider addresses or publisher addresses, an out-of-bounds panic could happen because it was assumed that the provider info would at least have provider addresses.

This is not assumed now.  If the situation is detected it is logged.

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -391,7 +391,12 @@ func (r *Registry) RegisterOrUpdate(ctx context.Context, providerID peer.ID, add
 	}
 
 	if info.Publisher.Validate() == nil && info.PublisherAddr == nil && info.Publisher == info.AddrInfo.ID {
-		info.PublisherAddr = info.AddrInfo.Addrs[0]
+		if len(info.AddrInfo.Addrs) == 0 {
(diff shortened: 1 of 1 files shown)


internal/registry/registry.go
```diff
@@ -391,7 +391,12 @@ func (r *Registry) RegisterOrUpdate(ctx context.Context, providerID peer.ID, add
 	}
 
 	if info.Publisher.Validate() == nil && info.PublisherAddr == nil && info.Publisher == info.AddrInfo.ID {
-		info.PublisherAddr = info.AddrInfo.Addrs[0]
+		if len(info.AddrInfo.Addrs) == 0 {
+			log.Warnw("Register provider with no provider or publisher addresses",
+				"provider", info.AddrInfo.ID, "publisher", info.Publisher)
+		} else {
+			info.PublisherAddr = info.AddrInfo.Addrs[0]
```
willscott, line 398: file a low importance bug to be able to support multiple addresses round-robin :)

willscott (approved):

internal/registry/registry.go
gammazero: @willscott #321

PR #327, opened 2022-04-05
Restructure identity settings to allow per instance key

masih:
## Context
We want each indexer instance to have a different peer ID but a stable peer ID. This change introduces a framework to specify different but stable peer identity for each instance of the indexer named after its statefulset pod name. So, all we need to do is to pre generate peer IDs then construct a secrets that map pod name as key to the identity. 

That secret is then mounted as a volume on indexer where each indexer instance is configured to read a specific file: the identity named after its pod name. 

## Proposed Changes
* inject `POD_NAME` as env var.
* construct path to indexer identity based on the expanded value of that env var.

## Revert Strategy
`git revert`


deploy/manifests/dev/us-east-2/tenant/storetheindex/kustomization.yaml
```diff
@@ -10,13 +10,7 @@ resources:
...
-  - name: indexer-env-vars
-    behavior: merge
-    literals:
-      - STORETHEINDEX_PRIV_KEY_PATH=/identity/identity.key
-
 secretGenerator:
   - name: indexer-identity
     files:
-      - identity.key=identity.encrypted
+      - indexer-0.key=indexer-0-identity.encrypted
```
willscott, line 16: will we need one of these for each instance?

willscott (approved):

deploy/manifests/dev/us-east-2/tenant/storetheindex/kustomization.yaml
masih: yes

masih:
> I wonder if we can start with a single secret and generate peerids as a function of that secret and replica number.

Captured: https://github.com/filecoin-project/storetheindex/issues/328

MarcoPolo (approved):

PR #333, opened 2022-04-06
Update version.json

MarcoPolo:
Bump again since release process was not clear.

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,3 +1,3 @@
 {
-  "version": "v0.4.2"
+  "version": "v0.4.3"
 }


MarcoPolo:
(Runbook has been updated)

masih:
If a tag exists, CI triggered by retrospectively updating the version [does check the existence](https://github.com/filecoin-project/storetheindex/runs/5858009319?check_suite_focus=true) and will not re-tag if it exists. So I think this PR is not necessary if created due to prior manual tagging.

masih (approved):

PR #337, opened 2022-04-08
Define production infrastructure in `us-east-2`

masih:
## Context
production environment infrastructure set up in storetheindex aws account region `us-east-2` while we wait for the PL managed clusters to be created.

## Proposed Changes
Define terraforms that instantiate production infrastructure pretty much
identical to dev environment with following changes:
- larger instance types in EKS nodegroup
- a dedicated r5b instance type node group to take advantage of high
  IOPS of io2 type EBS volumes supported by this type.

The terraform also removes duplicate tags by setting them all once in
the terraform `aws` provider. It also changes the naming convention for
IAM resources by using `path` that includes both region and environment
name to avoid naming conflicts across multiple environments created in
the same region of the same AWS account.

## Tests
Applied.

## Revert Strategy
`terraform destroy` `git revert`

### DIFF
--- a/deploy/infrastructure/prod/us-east-2/.terraform.lock.hcl
+++ b/deploy/infrastructure/prod/us-east-2/.terraform.lock.hcl
@@ -1,0 +1,61 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/aws" {
+  version     = "4.8.0"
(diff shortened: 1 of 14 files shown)


willscott (approved):

masih:
There is an issue with EKS worker nodes joining the cluster which I am investigating. Marking as draft until resolved.

MarcoPolo (commented):
Looks good, but I can review again once we figure out the worker node issue

masih:
> There is an issue with EKS worker nodes joining the cluster which I am investigating. Marking as draft until resolved.

...and the issue was [using role path for node group role](https://github.com/aws/containers-roadmap/issues/926).

PR #338, opened 2022-04-08
Use storethehash that has index GC

gammazero:
## Context
Use a new version of go-indexer-core that includes the latest go-storethehash with index garbage collection.

## Proposed Changes
- Update dependency on go-storethehash
- Make storethehash GC interval configurable

## Tests
Included in go-storethehash

## Revert Strategy
This is not safe to revert since it automatically upgrades the storethehash index format.

To revert, it would be necessary to write a converter that converts multiple index files back into a single index file.  This would be relatively simple as it would mean writing the old header to a file and appending all the index files to that file.  However, this is not available immediately and would need to be written.

### DIFF
--- a/config/indexer.go
+++ b/config/indexer.go
@@ -1,5 +1,9 @@
 package config
 
+import (
+	"time"
+)
+
 // Indexer holds configuration for the indexer core.
 type Indexer struct {
 	// Maximum number of CIDs that cache can hold. Setting to 0 disables cache.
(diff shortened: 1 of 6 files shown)


masih (approved):

config/indexer.go
```diff
@@ -17,6 +24,7 @@ func NewIndexer() Indexer {
 		CacheSize:      300000,
 		ValueStoreDir:  "valuestore",
 		ValueStoreType: "sth",
+		GCInterval:     Duration(30 * time.Minute),
```
masih, line 27: Would be good to make this configurable in future work.

gammazero: That's is just the default value.  It is configurable as a value in the config file.  Am I missing something?

masih: Nope it's me overlooking that; thanks

PR #340, opened 2022-04-08
Define prod kustomization overlay

masih:
## Context
K8S manifests to deploy cluster level and sit to self-Managed prod cluster.

## Proposed Changes
Define the prod kustomization overlay, which is pretty much the same as
dev overlay, with the following differences:
 - storetheindex demands specific taints to run on r5b instance type
 - indexer replica count is increased to 2

## Tests
https://prod.cid.contact 
Currently deployed with CD pipelines pointing to this PR for testing.

## Revert Strategy
`git revert`

### DIFF
--- a/deploy/manifests/prod/us-east-2/cluster/cert-manager/issuer.yaml
+++ b/deploy/manifests/prod/us-east-2/cluster/cert-manager/issuer.yaml
@@ -1,0 +1,17 @@
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt
+spec:
+  acme:
+    server: https://acme-v02.api.letsencrypt.org/directory
+    email: bedrock@protocol.ai
+    privateKeySecretRef:
+      name: letsencrypt
+    solvers:
+      - dns01:
+          route53:
+            region: us-east-2
+        selector:
+          dnsZones:
+            - prod.cid.contact
--- a/deploy/manifests/prod/us-east-2/cluster/kube-system/aws-auth.yaml
+++ b/deploy/manifests/prod/us-east-2/cluster/kube-system/aws-auth.yaml
@@ -1,0 +1,46 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: aws-auth
+  namespace: kube-system
+data:
+  mapRoles: |
+    - groups:
+      - system:bootstrappers
+      - system:nodes
+      rolearn: arn:aws:iam::407967248065:role/prod-ue2-r5b-xl-eks-node-group
+      username: system:node:{{EC2PrivateDNSName}}
+    - groups:
+      - system:bootstrappers
+      - system:nodes
+      rolearn: arn:aws:iam::407967248065:role/prod-ue2-m4-xl-eks-node-group
+      username: system:node:{{EC2PrivateDNSName}}
+  mapUsers: |
+    - userarn: arn:aws:iam::407967248065:user/masih
+      username: masih
+      groups:
+        - system:masters
+    - userarn: arn:aws:iam::407967248065:user/marco
+      username: marco
+      groups:
+        - system:masters
+    - userarn: arn:aws:iam::407967248065:user/gammazero
+      username: gammazero
+      groups:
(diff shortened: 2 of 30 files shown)


deploy/manifests/prod/us-east-2/cluster/cert-manager/issuer.yaml
```diff
@@ -0,0 +1,17 @@
...
+    email: bedrock@protocol.ai
+    privateKeySecretRef:
+      name: letsencrypt
+    solvers:
+      - dns01:
+          route53:
+            region: us-east-2
+        selector:
+          dnsZones:
+            - prod.cid.contact
```
willscott, line 17: if another entity wanted to deploy their own deployment of store the index, they'd need to change these variables. is there a way to template these specific names/email addresses out?

deploy/manifests/prod/us-east-2/cluster/kube-system/aws-auth.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - system:bootstrappers
+      - system:nodes
+      rolearn: arn:aws:iam::407967248065:role/prod-ue2-r5b-xl-eks-node-group
+      username: system:node:{{EC2PrivateDNSName}}
+    - groups:
+      - system:bootstrappers
+      - system:nodes
+      rolearn: arn:aws:iam::407967248065:role/prod-ue2-m4-xl-eks-node-group
+      username: system:node:{{EC2PrivateDNSName}}
+  mapUsers: |
```
willscott, line 18: can we have access templated to be shared between deployments, rather than duplicated in each instance / zone?

deploy/manifests/prod/us-east-2/tenant/storetheindex/indexer-1-identity.encrypted
```diff
@@ -0,0 +1,22 @@
+{
+	"data": "ENC[AES256_GCM,data:gMTrwBeT8M1FpliLFp3NBd7paAS+8JoemI/G+TxOMW8x2ERZi8wUALYk6p3Qv4yRRt6iuB93+evRrvuMZhfg+wGNUdw=,iv:49tzBmjrf8Nta6pgdceVRgv07aKIPqhKrFKKiKn7PVM=,tag:uL/CSXdJ0HsWRPbj4mjmFQ==,type:str]",
```
willscott, line 2: a readme / comment somewhere on how to regenerate these secrets would be great

willscott (approved):

deploy/manifests/prod/us-east-2/cluster/cert-manager/issuer.yaml
masih: Context: Manifests built using `Kustomization` overlays patch some base layer to add/edit the things they need to change. This means there is no templating as such. One would simply define a patch and use strategic merge to override values.

In the case of storetheindex, the base kustomization does not define `Ingress` object at all. Because we don't know how other people may choose to expose their indexer. This is why the ingress object is defined in `prod` and `dev` overlays explicitly. The `ingress` object is also somewhat Cloud-provider specific (you can see route53 conifig in there).

Another reason for not including it in base is that in kustomization overlays one can't remove manifests. So things like this are left out so that whoever is using the kustomization can configure the right abstraction for themselves.

deploy/manifests/prod/us-east-2/cluster/kube-system/aws-auth.yaml
masih: I _think_ we should be able to for users 👍 I will address this in a separate PR.

willscott (approved):

deploy/manifests/prod/us-east-2/tenant/storetheindex/indexer-1-identity.encrypted
masih: A quick guide added [here](https://github.com/filecoin-project/storetheindex/pull/340/commits/639cd3f29eaae20eade4f528b02eb2cbf975ebf8). I will flesh the docs out once the deployment manifests stabilise.

PR #341, opened 2022-04-08
Add script to generate random peer identity

masih:
## Context
Deployment of indexer nodes works with explicitly generated peer identity that is then encrypted and checked into code. Add a script that makes the generation of random identity easier. 

## Proposed Changes
Add a script to generate a random peer identity, used to instantiate
new indexer nodes. The script also prints out the peer ID from the
generated private key.

## Tests
Used to generate identities locally.

## Revert Strategy
`git revert`

### DIFF
--- a/scripts/generate_peer_id.go
+++ b/scripts/generate_peer_id.go
@@ -1,0 +1,45 @@
+package main
+
+import (
+	"crypto/rand"
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+func main() {
+
+	dest := flag.String("out", "", "Path at which to store the marshalled peer identity. Required.")
+	flag.Parse()
+
+	if dest == nil || *dest == "" {
+		panic("destination path must be specified via --out flag.")
+	}
+
+	out := filepath.Clean(*dest)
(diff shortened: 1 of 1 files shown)


MarcoPolo (changes requested):
Should this hook into SOPS to encrypt the contents? I feel like writing the file out unencrypted (especially in `0666`) is a foot gun.

masih:
> Should this hook into SOPS to encrypt the contents? I feel like writing the file out unencrypted (especially in `0666`) is a foot gun.

Are you suggesting calling SOPS directly from go? Sure we can do that. 
My thinking was to do the glueing in bash as a `make` target once manifests are merged to master.

MarcoPolo:
> Are you suggesting calling SOPS directly from go? Sure we can do that.
yep that's what I was suggesting.

We can use make instead, but I'd like to avoid the footgun if someone ran this manually.

What if the secret was output to stdout and the file was used to write the public peer id? That way you have to handle the secret material within the process and you can't accidentally commit a file with secrets. In the make script it can use a temp file for the peer id and delete it once it finishes.

masih:
Yup thank you for pointing this out; I'll have another go to see how I can make this safe. 

Side note: Re Peer ID printing, that's just so we know what the peer ID of the node would be. I made changes to the config [a while back](https://github.com/filecoin-project/storetheindex/pull/296) which makes peer ID optional if private key is specified, since it can be generated from private key, and assert that if it is specified it's consistent with private key.

PR #343, opened 2022-04-08
Add loadgen and loadgen-verify commands

MarcoPolo:
## Context

As part of load testing, I've maintained a set of tools out of repo for creating write load. Those tools lived at https://github.com/marcopolo/storetheindex-deployment. Now it's time to move the write load generator to this repo so that it stays alive and can verify consistency automatically via CI.

Note that the read load generator is a bit trickier since it involves coordinating lambdas to query an endpoint. The the other repo should be reproducible at any point in the future, and the things that may change are the easy bits of which url to hit for a find query, while the hard stuff (managing builds that get deployed to lambdas and having a nice interface to start a load test) should be stable.

## Proposed Changes
Adds a loadgen and loadgen-verify command

These are commands that can be used to generate write load on a storage provider, and verify the Ads have been correctly ingested.

Useful for both stress testing an indexer as well as verifying consistency.

To run this locally for example you would (in the root of this repo):

1. Start the indexer `go run . daemon`
2. Start the load generator `go run . loadgen --topic="indexer/ingest"` (note you may need to change the topic depending on how your local indexer is configure. See ~/.storetheindex/config for your specific settings)
3. Verify that all the load generated was correctly ingested `go run . loadgen-verify`

## Tests
Adds a couple new test that serve as a both a check that the loadgen command works as well as a way to verify the consistency of the whole storetheindex system.

## Revert Strategy

Change is safe to revert.


command/loadgen_test.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func TestSmallLoadOverHTTP(t *testing.T) {
+	ctx, cncl := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cncl()
+	testLoadHelper(ctx, t, 1, 1000, true)
+}
+
+func TestLargeLoad(t *testing.T) {
+	ctx, cncl := context.WithTimeout(context.Background(), 180*time.Second)
```
MarcoPolo, line 35: I may skip this in our current CI, but at least having a large load test in CI is pretty straightforward.

MarcoPolo: ah this finishes in 15s on the mac runner and 7s on the ubuntu runner. I'll keep it since it's our only concurrent integration test.

MarcoPolo:
`TestEndToEndWithReferenceProvider` is broken again. It works locally for me.

willscott:
I see:
```
 --- FAIL: TestLargeLoad (43.22s)
=== CONT  
    testing.go:1152: race detected during execution of test
FAIL
FAIL	github.com/filecoin-project/storetheindex/command	66.429s
```

should we not run that one with race detector?

MarcoPolo:
@gammazero I tagged you for a review. Mind taking a look at this? I'll rebase this soon. Thanks!

command/loadgen_test.go
```diff
@@ -0,0 +1,137 @@
...
+			LoadGenCmd,
+			LoadGenVerifyCmd,
+		},
+	}
+
+	err := app.RunContext(ctx, []string{"storetheindex", "init"})
+	require.NoError(t, err)
+
+	go func() {
+		app.RunContext(ctx, []string{"storetheindex", "daemon"})
```
gammazero, line 79: Should wait until this goroutine stops at end of test, so that tempDir can be removed on Windows.

e2e_test.go
```diff
@@ -118,6 +118,7 @@ func (e *e2eTestRunner) stop(cmd *exec.Cmd, timeout time.Duration) {
 }
 
 func TestEndToEndWithReferenceProvider(t *testing.T) {
+	t.Skip("Skipping because it's failing on CI but works locally. So it is unactionable. Before re-enabling this test, please try to make it reproducible. Ideally in a clean environment that is completely independent of the host machine.")
```
gammazero, line 121: This is working fine in CI now.

gammazero (commented):
Need to update to not use go-legs.

MarcoPolo:
@gammazero can you please take this? You have the most context at this point. 

Also if you want to run the load tests again, I would recommend against testground. I would suggest either the existing AWS setup or docker compose

gammazero:
@MarcoPolo Yes, I got this. Thanks for the suggestion re. testground.

There is a truly strange problem where the e2e_test fails when running all tests after clearing the cache. I have narrowed it down to this: `go clean -cache; go test -v ./ ./command/...`

MarcoPolo:
Some notes on testground: https://github.com/libp2p/test-plans/issues/103

gammazero (approved):

PR #357, opened 2022-04-14
Fix wrong publisher address when using HTTP

gammazero:
## Context
When retrieving an advertisement over HTTP, the indexer gets the wrong publisher address.  The indexer should get the address that the advertisement was retrieved from as the publisher address.

## Proposed Changes
When an advertisement is retrieved via HTTP, the publisher address is not in the peerstore.  The indexer then uses the provided address as the publisher address, since it cannot find an address for the publisher.  To get the HTTP address it is necessary to get the http peerstore from go-legs and look there first.

Also, revised command flags to correctly specify publisher or provider peer ID.

## Tests
Manual testing

## Revert Strategy
Safe to revert.

### DIFF
--- a/command/flags.go
+++ b/command/flags.go
@@ -14,13 +14,6 @@ var indexerHostFlag = altsrc.NewStringFlag(&cli.StringFlag{
 	Value:    "localhost",
 })
 
-var indexerIDFlag = altsrc.NewStringFlag(&cli.StringFlag{
-	Name:     "peerid",
-	Usage:    "Peer ID of indexer to use, when using libp2p",
-	EnvVars:  []string{"INDEXER_ID"},
-	Required: false,
-})
-
 var cacheSizeFlag = &cli.Int64Flag{
 	Name:     "cachesize",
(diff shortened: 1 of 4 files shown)


command/flags.go
```diff
@@ -142,8 +140,13 @@ var adminReloadPolicyFlags = []cli.Flag{
 }
 
 var adminSyncFlags = []cli.Flag{
-	peerFlag,
 	indexerHostFlag,
+	&cli.StringFlag{
+		Name:     "peer",
+		Usage:    "Publisher peer ID",
+		Aliases:  []string{"p"},
```
willscott, line 147: can we have these peer IDs differentiated in their name, e.g. 'pubpeer' vs 'propeer' or some way where we've made it clear if the ID for varying commands is referring to the publisher vs the provider?

willscott (approved):

command/flags.go
gammazero: Done

PR #358, opened 2022-04-14
Configure readiness probe and pod disruption budget

masih:
## Context
We want to maintain availability during deployment, and during other admin operations like worker node drainage.

## Proposed Changes
* Specify readiness probe so that pods are not used to serve incoming requests unless ready.
* Specify pod disruption budget so that we tell the cluster the maximum unavailable pods allowed.

## Tests
N/A

## Revert Strategy
`git revert`

### DIFF
--- a/deploy/manifests/base/storetheindex/indexer-statefulset.yaml
+++ b/deploy/manifests/base/storetheindex/indexer-statefulset.yaml
@@ -34,6 +34,15 @@ spec:
               mountPath: /data
             - name: config
               mountPath: /config
+          readinessProbe:
+            httpGet:
+              port: finder
+              path: /health
+            initialDelaySeconds: 10
+            failureThreshold: 3
+            successThreshold: 1
+            timeoutSeconds: 5
+            periodSeconds: 10
       volumes:
         - name: config
           configMap:
--- a/deploy/manifests/base/storetheindex/kustomization.yaml
+++ b/deploy/manifests/base/storetheindex/kustomization.yaml
@@ -6,6 +6,7 @@ namespace: storetheindex
 resources:
   - indexer-statefulset.yaml
   - indexer-service.yaml
+  - pdb.yaml
 
 transformers:
   - labels.yaml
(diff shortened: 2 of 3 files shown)


deploy/manifests/base/storetheindex/indexer-statefulset.yaml
```diff
@@ -34,6 +34,15 @@ spec:
               mountPath: /data
             - name: config
               mountPath: /config
+          readinessProbe:
+            httpGet:
+              port: finder
+              path: /health
+            initialDelaySeconds: 10
+            failureThreshold: 3
```
willscott, line 42: is this that it'll try 3 times each 10 seconds?

masih: Not quite; `failureThreshold` means how many times to try before giving up and marking as unready. 
3 is the default. I have specifically added the defaults to be explicit.

See [K8S reference](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#Probe) for more info.

willscott: if it gets marked unready what happens? just noting that we well may end up with startup times of more than 30 seconds, and that hopefully shouldn't result in needing some sort of manual intervention

masih: Nope; it just keeps trying until it's ready. 

Here is a summary of how the readiness probe behaves as configured above:
* Wait for 10 seconds before attempting to check for readiness (`initialDelaySeconds`)
* Then call HTTP `GET` on port `finder` at path `/health` every 10 seconds (`periodSeconds`) with timeout of 5 seconds (`timeoutSeconds`)
* Mark it as ready if we get `OK 200` back at least once (`successThreshold`)
* Mark it as unready if we get non `OK` three times in a row (`failureThreshold`) i.e. over 30 seconds.

The check runs for the lifetime of the pod and marks it as unready whenever the readiness probe fails.

willscott (approved):
sounds good

PR #429, opened 2022-04-22
Log and config update

gammazero:
## Context

* Put depth limit settings back into config as means to protect against endless chains.
It may be necessary to have depth limits to protect against unreasonably long chains of advertisements and entries.  Might as well make this configurable, so added them back to config.

* Log publisher of advertisements that fail to parse.
If publishers create advertisements that fail to decode, logs should indicate who that publisher is.

* Upgrading the config works even if config file is at current version.
If items from a config are missing or have been removed so that default values are used, it can be useful to recreate those items in the config file to see what values are available.  This can now be done by running `./storetheindex init --upgrade`  at anytime.

* Config upgrade renames previous config file with version suffix.
* Order config items alphabetically to make them easier to find.
* Remove unused constants and perform additional error checks

### DIFF
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -7,12 +7,14 @@ import (
 // Ingest tracks the configuration related to the ingestion protocol.
 type Ingest struct {
 	// The recursion depth limit when syncing advertisements. The value -1
-	// means no limit and zero means use the default value. This prupose is to
-	// prevent overload from endless advertisement chains.
+	// means no limit and zero means use the default value. Limiting the depth
+	// of advertisements can be done if there is a need to prevent an indexer
+	// from ingesting long chains of advertisements.
 	AdvertisementDepthLimit int
 	// The recursion depth limit when syncing advertisement entries. The value
-	// -1 means no limit and zero means use the default value. This prupose is
-	// to prevent overload from endless entries chains.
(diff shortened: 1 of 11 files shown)


config/ingest.go
```diff
@@ -6,37 +6,56 @@ import (
...
 }
 
 // NewIngest returns Ingest with values set to their defaults.
 func NewIngest() Ingest {
 	return Ingest{
-		PubSubTopic:       "/indexer/ingest/mainnet",
-		StoreBatchSize:    4096,
-		SyncTimeout:       Duration(2 * time.Hour),
-		IngestWorkerCount: 10,
+		AdvertisementDepthLimit: 65536,
```
willscott, line 37: some providers will pretty quickly exceed this, i think

willscott (approved):

config/ingest.go
gammazero: If a publisher publishes an advertisement every second, then that is 86400 a day, 604800 a week, 31.5M a year. So maybe something > 32M?

I can use a large value above or I could make it unlimited by default (-1), but still have it configurable if someone needs it. WDYT?

willscott: 32M seems pretty safe

PR #450, opened 2022-05-06
use default cidlink length

willscott:
## Context
cids are truncated to the first half of the sha256 digest

## Proposed Changes
Use the default mhlength on the cid prototype

## Tests
go test continues to pass

## Revert Strategy
revert

### DIFF
--- a/api/v0/ingest/schema/schema.go
+++ b/api/v0/ingest/schema/schema.go
@@ -35,7 +35,7 @@ var (
 			Version:  1,
 			Codec:    uint64(multicodec.DagJson),
 			MhType:   uint64(multicodec.Sha2_256),
-			MhLength: 16,
+			MhLength: -1,
 		},
 	}
 


masih (approved):
Looking at [previous versions](https://github.com/filecoin-project/storetheindex/blob/v0.3.8/api/v0/ingest/schema/utils.go#L31) The truncated multihash length predates the bind node changes which moved the code to `schema.go` hence the git history.

The changes look good; though I don't know why it was truncated in the first place. Maybe @gammazero  remembers?

adlrocha:
> The changes look good; though I don't know why it was truncated in the first place. Maybe @gammazero remembers?

This is from when we were tinkering with the underlying storage trying to figure out if using truncated hashes could lead to better performance. We didn't find any performance benefits and I am afraid we didn't change it back. Indeed, there's no reason for truncating the cid. 

(I hope this helps)

masih:
It does thanks @adlrocha 🍻

PR #459, opened 2022-05-09
Define `index-observer` ECR repo and CI to build and push images

masih:
Define a dedicated ECR repository for `index-observer` in `us-east-2`
and allow GitHub Actions to assume the roles necessary to build and push
container images to it.

While at it reformat terraform files in `common` via `terraform fmt`.


willscott:
is there a place to pass arguments to the container?

willscott:
in particular, it should be passed `--indexer https://cid.contact` as an argument

willscott (approved):

PR #464, opened 2022-05-09
add reframe api server

petar:
Add a Reframe API HTTP server to STI.

Added end-to-end test with a Reframe Client calling FindProviders on the STI HTTP server's Reframe endpoint.

Change is safe to revert.

### DIFF
--- a/server/reframe/reframe.go
+++ b/server/reframe/reframe.go
@@ -35,7 +35,7 @@ func (x *ReframeService) FindProviders(key cid.Cid) (<-chan client.FindProviders
 	if err != nil {
 		return nil, err
 	}
-	ch := make(chan client.FindProvidersAsyncResult)
+	ch := make(chan client.FindProvidersAsyncResult, 1)
 	peerAddrs := []peer.AddrInfo{}
 	for _, mhr := range fr.MultihashResults {
 		if !bytes.Equal(mhr.Multihash, mh) {


server/reframe/reframe.go
```diff
@@ -0,0 +1,70 @@
...
+	mh := key.Hash()
+	fr, err := x.finderHandler.Find([]multihash.Multihash{mh})
+	if err != nil {
+		return nil, err
+	}
+	ch := make(chan client.FindProvidersAsyncResult)
+	peerAddrs := []peer.AddrInfo{}
+	for _, mhr := range fr.MultihashResults {
+		if !bytes.Equal(mhr.Multihash, mh) {
+			continue
```
gammazero, line 42: Should an error be logged here, since there should never be a result for a multihash that was not asked for?

server/reframe/reframe.go
```diff
@@ -0,0 +1,70 @@
...
+	finderHandler *handler.FinderHandler
+}
+
+func (x *ReframeService) FindProviders(key cid.Cid) (<-chan client.FindProvidersAsyncResult, error) {
+	mh := key.Hash()
+	fr, err := x.finderHandler.Find([]multihash.Multihash{mh})
+	if err != nil {
+		return nil, err
+	}
+	ch := make(chan client.FindProvidersAsyncResult)
```
gammazero, line 38: Will there always be a reader for this channel?  If not, then make it buffered so that the writer goroutine does not hang/leak when there is no reader.

Also, if there are cases when there may not be a reader, then consider having the caller pass in a channel if they want to reveive an async result.

server/reframe/reframe.go
petar: The type ReframeService is just a pass-through adapter on top of FinderHandler. Therefore it should not be used-for/relied-on for discovering bugs in the underlying FinderHandler. I am happy to add a log, but if you want to catch bugs in the FinderHandler, the logging should be there. Let me know what you think.

petar:
@gammazero @willscott Ping to take a look at my responses.

gammazero (approved):
LGTM

PR #470, opened 2022-05-11
Onboard `autoretrieve` as a temporary tenant of sti `dev` cluster

masih:
While `autoretrieve` is being developed, provide a temporary place it
can be deployed in such that metrics could be forwarded to PL grafana
along with logs etc. All of this is already set up in sti EKS clusters.

Create a dedicated ECR repository for automating the build and push of
`autoretrieve` container images. It looks like the exising one is
manually created and pushed to. Here we create a separate one to avoid
any conflicts.

### DIFF
--- a/deploy/infrastructure/common/ecr.tf
+++ b/deploy/infrastructure/common/ecr.tf
@@ -4,6 +4,7 @@ module "ecr_ue2" {
   repositories = [
     "storetheindex/storetheindex",
     "index-observer/index-observer",
+    "autoretrieve/autoretrieve",
   ]
   tags = local.tags
 }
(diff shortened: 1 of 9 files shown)


willscott (approved):
I'm okay with this, but it's a pretty non-obvious place and eventually it would be great to copy the deploy structure in sti to autoretrieve directly

masih:
> I'm okay with this, but it's a pretty non-obvious place and eventually it would be great to copy the deploy structure in sti to autoretrieve directly

There are two levels of administration in the current setup: cluster level and application level. 


The cluster level controls cluster-wide controllers, namespace and tenancy creation. While the application level defines what should run in a tenant's namespace. 

For storetheindex, _it so happens that_ we manage both from the same repository. These are separated in `/cluster` and `/tenant` directories under each deployment manifest. 

In the case of `autoretrieve` ( or `index-observer`) They are tenants of storetheindex cluster but they define their own application level manifests. _Because_, this repo manages the cluster-level manifests, we need to onboard tenants here. But each of those tenants will then define their own application level manifests in their own repo. I have left READMEs in the root of cluster level tenancy manifests to point this out. 

Existing example for `index-observer`: https://github.com/filecoin-shipyard/index-observer/tree/main/deploy/manifests
As you can see this directory only contains manifests to run `index-observer` itself and not any other controllers etc. Because, it is a tenant of another cluster.

Does that make sense?

PR #480, opened 2022-05-13
Use larger instance types and override config in dev to limit providers

masih:
The dev instances are still running on 10GiB memory because the maximum
available on the worker node type is 16. It also uses a configuration
identical to prod, which means it would ingest everything prod does.

Here we whitelist only a selected number of providers to reduce load on
the `dev` instance since it is meant to be for development purposes.
This should shield it from failures that say prod instance goes through
due to excessive ingest load.

The memory given to the instances as well as the instance types are
upgraded to match production. Taint is set on the dev instance to use
r5b instance types just like prod. This will allow us to change the dev
config as we see fit to match the prod config or selectively ingest from
specific providers for debuging purposes without having to worry about
mismatching hardware requirements.

### DIFF
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/config
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/config
@@ -1,0 +1,71 @@
+{
+  "Version": 1,
+  "Identity": {
+    "PeerID": "",
+    "PrivKey": ""
+  },
+  "Addresses": {
+    "Admin": "/ip4/0.0.0.0/tcp/3002",
+    "Finder": "/ip4/0.0.0.0/tcp/3000",
+    "Ingest": "/ip4/0.0.0.0/tcp/3001",
+    "P2PAddr": "/ip4/0.0.0.0/tcp/3003",
+    "NoResourceManager": true
+  },
+  "Bootstrap": {
+    "Peers": [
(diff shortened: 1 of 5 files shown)


deploy/manifests/dev/us-east-2/tenant/storetheindex/config
```diff
@@ -0,0 +1,71 @@
...
+    "Dir": "/data/datastore"
+  },
+  "Discovery": {
+    "LotusGateway": "wss://api.chain.love",
+    "Policy": {
+      "Allow": false,
+      "Except": [
+        "12D3KooWDtiA9w5c37MnFpGWr2M12m8neoUqDroHEMsJVuf3ELi7",
+        "12D3KooWBwUERBhJPtZ7hg5N3q1DesvJ67xx9RLdSaStBz9Y6Ny8",
+        "12D3KooWLDf6KCzeMv16qPRaJsTLKJ5fR523h65iaYSRNfrQy7eU"
```
masih, line 45: @gammazero please double check this is the right syntax for whiltelisting only a selected set of providers.

gammazero: That is correct.

gammazero (approved):

PR #489, opened 2022-05-14
Sync ads and entries in configurable segments

### LINKED ISSUE #487: HTTP sync of large advertisement chains causes out of memory
When fetched over HTTP, very long advertisement chains cause out of memory very quickly compared to over graphsync.

Comment from @masih :
> we run out of memory very quickly for long chains synced via HTTP.  In order to implement the selector mechanism over HTTP, we are using traversal.Progress [here](https://github.com/filecoin-project/go-legs/blob/main/httpsync/sync.go#L186).  That type keeps updating an internal field called Path with whatever path it comes across. Eventually that path grows massive when the recursion depth is unlimited. That path is stored in memory.
> 
> Breaking up calls to `progress.Walk` won’t help since even if we resume we need the path to tell it where to resume from.
So i think we need to break up sync in the storetheindex level since the application knows what the node types are  and it can then instantiate a new sync say from an ad’s previous ID having synced all its entry after some configurable depth

masih:
Syncing excessively long advertisement or entry chunk chains from an
HTTP publisher results in out of memory error. This is caused by an
internal state maintained by IPLD during `traversal.Progress`, where the
traversal path is updated in memory until the traversal completes.

To avoid memory issues as a result of syncing long chains in one go
use the segmented sync mechanism introduced in go-legs, where a chain is
synced in segments with a configurable segment depth limit. The sync
process is then repeated until the desired total depth limit is
traversed or we reach the end of traversal. At each segmented sync
cycle, depending on the kind of node being synced, the latest synced
node is decoded, and based on its type (either `Advertisement` or
`EntryChunk`) the next segment link is determined.

By breaking up syncs across multiple smaller segments we will avoid
memory problems that seem to surface when syncing very long chains while
at the same time we retain the ability to fully sync such chains.

Note that storetheindex default selector only selects advertisements.
The syncing of the ads is either triggerred via an explicit sync call on
the ingester or in background via announce messages. Therefore, the
default segment block hook that determines the next link in segmented
sync is set to only decode cids as advertisements. The advertisement
entries are synced explicitly as part of processing logic and use a
scoped segment block hook to only decode entry chunks.

The default segment depth limit is set to 2K even though an indexer
instance can sync much larger DAGs in one go without hitting memory
issues. This is because the indexer could be syncing from many index
providers at the same time and the total sync time is not dramatically
impacted by the number of segments.

Fixes #487

### DIFF
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -6,20 +6,26 @@ import (
 
 // Ingest tracks the configuration related to the ingestion protocol.
 type Ingest struct {
-	// The recursion depth limit when syncing advertisements. The value -1
-	// means no limit and zero means use the default value. Limiting the depth
-	// of advertisements can be done if there is a need to prevent an indexer
-	// from ingesting long chains of advertisements.
+	// AdvertisementDepthLimit is the total maximum recursion depth limit when
+	// syncing advertisements. The value -1 means no limit and zero means use
+	// the default value. Limiting the depth of advertisements can be done if
+	// there is a need to prevent an indexer from ingesting long chains of
+	// advertisements.
+	// Note that sync is divided across multiple individual calls to a provider.
+	// See SyncSegmentDepthLimit.
 	AdvertisementDepthLimit int
-	// The recursion depth limit when syncing advertisement entries. The value
-	// -1 means no limit and zero means use the default value. The purpose is
-	// to prevent overload from extremely long entry chains resulting from
-	// publisher misconfiguration.
+	// EntriesDepthLimit is the total maximum recursion depth limit when syncing
+	// advertisement entries. The value -1 means no limit and zero means use
+	// the default value. The purpose is to prevent overload from extremely
+	// long entry chains resulting from publisher misconfiguration.
+	// Note that sync is divided across multiple individual calls to a provider.
+	// See SyncSegmentDepthLimit.
 	EntriesDepthLimit int
-	// How many ingest worker goroutines to spawn. This controls how many
-	// concurrent ingest from different providers we can handle.
+	// IngestWorkerCount sets how many ingest worker goroutines to spawn. This
+	// controls how many concurrent ingest from different providers we can handle.
 	IngestWorkerCount int
-	// PubSubTopic used to advertise ingestion announcements.
+	// PubSubTopic sets the topic name to which to subscribe for ingestion
+	// announcements.
 	PubSubTopic string
 	// RateLimit contains rate-limiting configuration.
 	RateLimit RateLimit
@@ -33,6 +39,11 @@ type Ingest struct {
 	// or a chain of advertisement entries. The value is an integer string
 	// ending in "s", "m", "h" for seconds. minutes, hours.
 	SyncTimeout Duration
+	// SyncSegmentDepthLimit is the depth limit of a single sync in a series of
+	// calls that collectively sync advertisements or their entries. The value
+	// -1 disables the segmentation where the sync will be done in a single call
+	// and zero means use the default value.
+	SyncSegmentDepthLimit int
 }
 
 // NewIngest returns Ingest with values set to their defaults.
@@ -45,6 +56,7 @@ func NewIngest() Ingest {
 		RateLimit:               NewRateLimit(),
 		StoreBatchSize:          4096,
 		SyncTimeout:             Duration(2 * time.Hour),
+		SyncSegmentDepthLimit:   2_000,
 	}
 }
 
@@ -70,5 +82,8 @@ func (c *Ingest) populateUnset() {
 	}
 	if c.SyncTimeout == 0 {
 		c.SyncTimeout = def.SyncTimeout
+	}
+	if c.SyncSegmentDepthLimit == 0 {
+		c.SyncSegmentDepthLimit = def.SyncSegmentDepthLimit
 	}
 }
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/filecoin-project/go-address v0.0.5
 	github.com/filecoin-project/go-dagaggregator-unixfs v0.2.0
 	github.com/filecoin-project/go-indexer-core v0.2.15
-	github.com/filecoin-project/go-legs v0.3.14
+	github.com/filecoin-project/go-legs v0.3.15-0.20220523114716-031a27fc689d
 	github.com/frankban/quicktest v1.14.2
 	github.com/gogo/protobuf v1.3.2
 	github.com/gorilla/handlers v1.5.1
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -45,7 +45,7 @@ const (
 
 type adProcessedEvent struct {
 	publisher peer.ID
-	// Head of the chain being proecessed.
+	// Head of the chain being processed.
 	headAdCid cid.Cid
 	// Actual adCid being processing.
 	adCid cid.Cid
@@ -78,7 +78,7 @@ type Ingester struct {
 	inEvents chan adProcessedEvent
 
 	// outEventsChans is a slice of channels, where each channel delivers a
-	// copy of an adProccedEvent to an onAdProcessed reader.
+	// copy of an adProcessedEvent to an onAdProcessed reader.
 	outEventsChans map[peer.ID][]chan adProcessedEvent
 	outEventsMutex sync.Mutex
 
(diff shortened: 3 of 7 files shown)


config/ingest.go
```diff
@@ -42,6 +53,7 @@ func NewIngest() Ingest {
 		PubSubTopic:             "/indexer/ingest/mainnet",
 		StoreBatchSize:          4096,
 		SyncTimeout:             Duration(2 * time.Hour),
+		SyncSegmentDepthLimit:   1000,
```
willscott, line 56: did we have problems that caused us to set this so low? lets see if we can get away with 10k if we haven't

masih, line 56: The value is entirely made up; setting it would need some measurements. Please shout if you have a more reasonable initial value in mind 🙇

internal/ingest/segmented_sync.go
```diff
@@ -0,0 +1,194 @@
+package ingest
```
willscott, line 1: this filename doesn't describe the methods in it super well. not sure if there's a clearer naming convention between functions and filenames that might help

internal/ingest/ingest.go
```diff
@@ -140,9 +120,19 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr indexer.Interface, reg *re
...
 
+	// Construct a selector that recursively looks for nodes with field
+	// "PreviousID" as per Advertisement schema. Note that the entries within
+	// an advertisement are synced separately, triggered by storage hook. This
+	// allows checking if a chain of chunks already exists before syncing it.
+	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
+	adSequence := ssb.ExploreFields(
+		func(efsb builder.ExploreFieldsSpecBuilder) {
+			efsb.Insert("PreviousID", ssb.ExploreRecursiveEdge())
+		}).Node()
```
willscott, line 131: these could be top level variables, not re-defined each time the function runs. They are are also available in the segmented_sync.go file, so should try not to shadow.

internal/ingest/segmented_sync.go
```diff
@@ -0,0 +1,194 @@
...
+		func(efsb builder.ExploreFieldsSpecBuilder) {
+			efsb.Insert("PreviousID", ssb.ExploreRecursiveEdge())
+		})
+	entriesSequence = ssb.ExploreFields(func(efsb builder.ExploreFieldsSpecBuilder) {
+		efsb.Insert("Next", ssb.ExploreRecursiveEdge()) // Next field in EntryChunk
+	})
+)
+
+func (ing *Ingester) syncEntries(ctx context.Context, pid peer.ID, c cid.Cid, peerAddr multiaddr.Multiaddr, hookFunc legs.BlockHookFunc) error {
+
```
willscott, line 30: remove empty line

internal/ingest/segmented_sync.go
```diff
@@ -0,0 +1,194 @@
...
+		sel := ssb.ExploreRecursive(selector.RecursionLimitDepth(int64(nextDepthLimit)), entriesSequence).Node()
+		_, err := ing.sub.Sync(ctx, pid, nextSegmentCid, sel, peerAddr, hookOpt)
+		if err != nil {
+			return err
+		}
+		depthSoFar += nextDepthLimit
+
+		// If the latest chunk load failed in block hook then return it here.
+		if latestLoadErr != nil {
+			log.Error("failed to load chunk during segmented sync", "err", err)
```
willscott, line 77: does this mean on failure we aren't caching the subset we have already and start over from the beginning of the sync? is there a way to have failures still cache the subset that was gotten and resume from there?

internal/ingest/segmented_sync.go
```diff
@@ -0,0 +1,194 @@
...
+			return latestLoadErr
+		}
+
+		// If there is no next chunk link we are at the end of entries chain; nothing else to do.
+		if nextChunkLink == nil {
+			return nil
+		}
+		nextSegmentCid = (*(nextChunkLink)).(cidlink.Link).Cid
```
willscott, line 85: catch panics?

internal/ingest/segmented_sync.go
```diff
@@ -0,0 +1,194 @@
...
+		if latestLoadErr != nil {
+			log.Error("failed to load ad during segmented sync", "err", err)
+			return cid.Undef, latestLoadErr
+		}
+
+		// If there is no previous ad link we are at the end of the ad chain; nothing else to do.
+		if prevAdLink == nil {
+			return headCid, nil
+		}
+		nextSegmentCid = (*(prevAdLink)).(cidlink.Link).Cid
```
willscott, line 176: catch panics

masih: Sorry could you helm me understand what could cause a panic here?

willscott: if the ipld.link was not a cidlink.Link. I think we can probably ignore that edge case though.

masih: There are a lot of places in the codebase where we cast `ipld.link` to `codlink.Link`. If this is not a blocker just now, I'll defer safety check in another PR and apply it across the entire repo.

willscott: yeah, low priority. can be safely ignored for now, i think

config/ingest.go
gammazero: Our EC2 dev node was able to get to around 15000 advertisements before it died.  So, I think 10000 for prod is reasonable.

internal/ingest/segmented_sync.go
masih: Going with `synced.go` please shout if you can think of clearer alternatives

internal/ingest/segmented_sync.go
masih: See: https://github.com/filecoin-project/storetheindex/pull/489#discussion_r873449249

internal/ingest/segmented_sync.go
masih: The blocks are written into datastore right up to the point where a potential failure happens. We can implement some resumption strategy here where:
* given a CID, we recursively walk its chain (either `Next` for `EntryChunk` or `PreviousID` for `Advertisement`)
* for each link we check if it is present in the datastore
* start segmented sync from the first link that is not found in the datastore.

A potential issues I see is that background gorutines might have already processed and removed those ads from the datastore.

Thoughts?

willscott: happy with this being in a subsequent follow-up PR.

eventually with storing blocks in e.g. a local IPFS data store i'd hope the blocks wouldn't get removed once they're there.

masih: That'd be great; that'd also make the sync more efficient in that we don't need to decode every cid during sync to learn its next node in the chain since processing might remove it if we defer the load

internal/ingest/segmented_sync.go
masih: https://github.com/filecoin-project/storetheindex/issues/491

internal/ingest/segmented_sync.go
masih: https://github.com/filecoin-project/storetheindex/issues/492

willscott (approved):

gammazero:
Looks like it deadlocks.  That should be addressed first.

masih:
> Looks like it deadlocks. That should be addressed first.

Yep working on it.

internal/ingest/linksystem.go
```diff
@@ -242,29 +241,23 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
...
 	if err != nil {
-		return fmt.Errorf("cannot fetch the node from datastore: %w", err)
+		return fmt.Errorf("failed to load chunk from datastore ipldNode: %w", err)
 	}
+
 	defer func() {
 		// Remove the content block from the data store now that processing it
 		// has finished. This prevents storing redundant information in several
 		// datastores.
+		entryChunkKey := dsKey(entryChunkCid.String())
```
gammazero, line 253: Since `chunkKey` has to be computed twice, here and in `loadEntryChunk`, it would be better to compute it once and pass it into `loadEntryChunk`.  That will avoid an additional `dsKey.Clean`

internal/ingest/ingest_test.go
```diff
@@ -744,28 +752,27 @@ func TestSkipEarlierAdsIfAlreadyProcessedLaterAd(t *testing.T) {
...
+	const entriesDepth = 10
+	totalChunkCount := entriesDepth * 2
+
 	srcStore := dssync.MutexWrap(datastore.NewMapDatastore())
 	h := mkTestHost()
 	pubHost := mkTestHost()
-	ing, core, _ := mkIngest(t, h)
+	ing, core, _ := mkIngestWithConfig(t, h, config.Ingest{
+		AdvertisementDepthLimit: 100,
+		EntriesDepthLimit:       10,
```
gammazero, line 742: Should this be set using the const `entriesDepth` or is it just a coincidence these have the same value?

gammazero (approved):
Code looks good.

gammazero (changes requested):
After indexer reads the first segment of ads, it then starts to immediately process the entries for those ads, even though there are many more segments on the ad chain to process.

All segments need to be read in first so that all ads can be processed in order. Ad processing is triggered by SyncFinished event.  So we need to have the sync finished only when all segments have been read in, or need to somehow know that it is necessary to wait for additional segments.

masih:
Depends on https://github.com/filecoin-project/go-legs/pull/126

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/filecoin-project/go-address v0.0.5
 	github.com/filecoin-project/go-dagaggregator-unixfs v0.2.0
 	github.com/filecoin-project/go-indexer-core v0.2.15
-	github.com/filecoin-project/go-legs v0.3.14
+	github.com/filecoin-project/go-legs v0.3.15-0.20220519133545-bcb09a8350f0
```
masih, line 10: To be replaced with a concrete version once https://github.com/filecoin-project/go-legs/pull/126 is merged.

gammazero (approved):
This can be merged now and go-legs API changes handled in another PR, or you can continue on to handle those API changes in this PR.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/filecoin-project/go-address v0.0.5
 	github.com/filecoin-project/go-dagaggregator-unixfs v0.2.0
 	github.com/filecoin-project/go-indexer-core v0.2.15
-	github.com/filecoin-project/go-legs v0.3.14
+	github.com/filecoin-project/go-legs v0.3.15-0.20220523114716-031a27fc689d
```
masih, line 10: This is a [commit from `main`](https://github.com/filecoin-project/go-legs/commit/031a27fc689d1e1339638ad225118f62efd3bacd) branch in`go-legs`. The concrete version is pending the review and merge of https://github.com/filecoin-project/go-legs/pull/127.

In the meantime, I am going ahead with the merge of this PR to get the memory fix deployed on `dev` cluster and start the full sync from NFT storage provider.

Cc @willscott @gammazero

PR #498, opened 2022-05-19
Fix intermittent test failures

masih:
In `TestEndToEndWithReferenceProvider` reset `Stdin` and `Stderr` for
the command before running it in a combined output. Otherwise,
`CombinedOutput` will return an error. Increase total E2E test timeout
to 2 minutes, since it seems to intermittently fail on MacOS.

In `TestMultiplePublishers`, assert indices from each test publisher are
eventually indexed since the background ad processing might not have
finished when the sync returns. Assert Synced CID is equal to the
expected head before checking reference to the latest synced.

In `TestPollProvider` registry tests close the registry to avoid failure
in Windows when tests attempt to remove temporary files that otherwise
will be in use and cannot be deleted.

### DIFF
--- a/e2e_test.go
+++ b/e2e_test.go
@@ -46,8 +46,12 @@ func (e *e2eTestRunner) run(name string, args ...string) []byte {
 
 	cmd := exec.CommandContext(e.ctx, name, args...)
 	cmd.Env = e.env
+	// Behaviour seems to be platform specific.
+	// Assure std in/err is not set otherwise combined output fails.
+	cmd.Stdin = nil
+	cmd.Stderr = nil
 	out, err := cmd.CombinedOutput()
-	qt.Assert(e.t, err, qt.IsNil, qt.Commentf("output: %s", out))
+	qt.Assert(e.t, err, qt.IsNil, qt.Commentf("err: %v, output: %s", err, out))
 	return out
 }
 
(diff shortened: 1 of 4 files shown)


masih:
@gammazero I pulled the changes here from #489; I thought we might want these in `main` sooner.

willscott (approved):
🙇

e2e_test.go
```diff
@@ -46,8 +46,12 @@ func (e *e2eTestRunner) run(name string, args ...string) []byte {
 
 	cmd := exec.CommandContext(e.ctx, name, args...)
 	cmd.Env = e.env
+	// Behaviour seems to be platform specific.
+	// Assure std in/err is not set otherwise combined output fails.
+	cmd.Stdin = nil
+	cmd.Stderr = nil
```
gammazero, line 52: This should not be possible for `cmd.Stdin` or `cmd.Stderr` to be non-nil, because this is the go code [here](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/os/exec/exec.go;l=169) which is not platform-specific. Also `cmd.CombinedOutput` would always fail for whatever platform is not setting these to `nil`, because of [this](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/os/exec/exec.go;l=558-563).

Please let me know how you detected this condition because that indicates a significant golang bug.

gammazero (approved):
See comment.

e2e_test.go
masih: Thank you for pointing this out; you are right, it certainly looks that way. 
Going to address this in a separate PR

masih: https://github.com/filecoin-project/storetheindex/pull/500

PR #512, opened 2022-05-24
Retry HTTP sync requests

masih:
Syncing very long ad chains from HTTP publishers can fail with retryable
error in `500` range. If the HTTP requests during sync in legs
`httpsync` fail the entire sync call made to legs subscriber will be
marked as failure. This means for a sync call on subscriber to succeed
the entire ad chain sync must be successful.

This issue becomes more visible when syncing long chains. If such
failures occur consistently, say due to hitting rate limiting on the
provider side, then it means we will never end up syncing very long ad
chains because the chances are not every call to fetch the content will
be successful.

To work around this, simply retry the HTTP calls with configurable
backoff and max retry times.

See:
- https://github.com/filecoin-project/go-legs/issues/128
- https://pkg.go.dev/github.com/hashicorp/go-retryablehttp


gammazero (approved):
I am wondering if this is better implemented in go-legs, [here](https://github.com/filecoin-project/go-legs/blob/main/httpsync/sync.go#L41), especially if the eventual goal is to have go-legs handle retries for HTTP and libp2p sync failure.

masih:
> I am wondering if this is better implemented in go-legs, [here](https://github.com/filecoin-project/go-legs/blob/main/httpsync/sync.go#L41), especially if the eventual goal is to have go-legs handle retries for HTTP and libp2p sync failure.

Tried that first but it was a bigger code change and involved adding http sync options to make retry parameters configurable.

Considering go legs makes the entire http client replaceable, adding it here seemed to me like the path of least resistance.

PR #521, opened 2022-05-25
Re-allow all peers on `prod` and separate the config from `base`

masih:
Re allow all the peers on production.

Fork the configuration in `prod` overlay so that we don't have to change
the base kustomization to impact production config. This change is
already done for the `dev` environment.


willscott (approved):

deploy/manifests/prod/us-east-2/tenant/storetheindex/config
```diff
@@ -0,0 +1,67 @@
...
+    "ValueStoreType": "sth",
+    "GCInterval": "30m0s"
+  },
+  "Ingest": {
+    "PubSubTopic": "/indexer/ingest/mainnet",
+    "StoreBatchSize": 4096,
+    "SyncTimeout": "2h0m0s",
+    "IngestWorkerCount": 10
+  }
+}
```
masih, line 67: This is identical to the previous config with the peer ID exclusion removed.

Issue #485, opened 2022-05-13
runtime updates to config


willscott:
would be great to be able push ratelimiting and allow-list config changes during runtime without needing to restart the indexer

masih:
For config to change without triggering redeployment 2 things need to get done:
1. application level reloading of content, say from watching the config file and reflecting any update at runtime.
2. change in the manifests to use a fixed ConfigMap name such that when the config changes the statefulset manifests does not change. It changes today because we generate the ConfigMap using a generator and that, by convention, concatenates a short SHA of content to the name of ConfigMap. This is a common practice since in a typical case one would want the pods to rotate when config changes.

@gammazero I can take the second item.

willscott:
the http endpoint PR  @gammazero made provides #1 as i understand it

masih:
#551 completed the item 2. Closing.

PR #529, opened 2022-05-26
Reload policy and rate limit settings without restarting daemon

gammazero:
## Context
Use the admin sub-command `reload-config`, or send the daemon process a SIGHUP signal, to reload policy and rate limit settings from the config file.

## Proposed Changes
- New command `admin reload-config` to reload parts of config.
- Handles SIGHUP to signal daemon to reload portions of config.

## Tests
Manual testing for success and failure
 
## Revert Strategy
Change is safe to revert.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Administrative:
 - `admin` Perform admin activities with an indexer
   - `allow` Allow advertisements and content from peer
   - `block` Block advertisements and content from peer
-  - `reload` Reload the policy from the configuration file
+  - `reload` Reload the policy and ingest settings from the configuration file
   - `sync` Sync indexer with provider
 - `init` Initialize or upgrade indexer node config file
 
--- a/api/v0/admin/client/http/client.go
+++ b/api/v0/admin/client/http/client.go
@@ -120,9 +120,9 @@ func (c *Client) Sync(ctx context.Context, peerID peer.ID, peerAddr multiaddr.Mu
 	return c.ingestRequest(ctx, peerID, "sync", http.MethodPost, data, q...)
 }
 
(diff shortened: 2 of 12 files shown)


willscott (commented):
there are going to be some ops config changes that will also need to happen so that the updated config can be changed against the running pod without triggering a full re-deploy, right?

gammazero:
> there are going to be some ops config changes that will also need to happen so that the updated config can be changed against the running pod without triggering a full re-deploy, right?

There probably should be if it is possible for ops to determine whether a reload is sufficient or whether a full restart is needed.  Currently, reload only reloads the allow/block policy for publishers and providers and reloads rate-limit settings. If there are changes to the config outside of those, then a redeploy is needed. @masih WDYT?

I may be able to reload the entire go-legs subscriber and network servers and registry with more work. Although the more that gets reloaded, the more potential for disruption.

gammazero:
@willscott Even if this does not work with our CI, it still allows us to log into an indexer node and update the policy and rate-limiting without having to restart the indexer.  So, it is still useful for emergencies.

masih:
Right now any change to config will retriever a pod redeployment. This is because the config map is generated and its name contains a short sha of its content, which is then automatically updated in the statefulset manifest that in turn triggers a redeployment. 

I have added a [comment](https://github.com/filecoin-project/storetheindex/issues/485#issuecomment-1139396606) on the related ticket to capture what needs to happen on ops side.

gammazero:
If the daemon is run with the `--watch-config` flag, or the environ var `INDEXER_WATCH_CONFIG` is set, then the indexer will check every 30 seconds to see if the config file has changed. If it has, then it will reload the reloadable sections.

This option is off by default.

Issue #523, opened 2022-05-26
Investigate high CPU usage on production


masih:
After rolling out the segmented sync and http retry CPU usage seems to be high with significant increase in open file descriptors on prod. 

Investigate why

PR #542, opened 2022-05-30
Defer handling of announce until ad chain processing is finished

masih:
In a case where a provider has a consistently growing ad chain where
each ad is explicitly announced while the indexer is in the progress of
processing previously announced ads a race-condition could occur where
the ingestion logic competes with the announce logic for syncing CIDs.
The former competes for syncing the ad entries so that they can be
processed, while the latter competes for syncing the newly announced
ads. As a result, the processing of ads slows down and CPU usage
increases significantly.

To avoid this postpone processing of any explicit announce request while
an ad chain is being processed. Once the ad chain processing finished,
then handle the announce request. Note that we only need to handle the
latest announce request because of the chaining of advertisements; as
long as we know the latest head we can discover the rest.

Relates to: #523

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -57,6 +57,12 @@ type adProcessedEvent struct {
 
 type providerID peer.ID
 
+// pendingAnnounce captures an announcement received from a provider that await processing.
+type pendingAnnounce struct {
+	addrInfo peer.AddrInfo
+	nextCid  cid.Cid
+}
+
 // Ingester is a type that uses go-legs for the ingestion protocol.
 type Ingester struct {
 	host    host.Host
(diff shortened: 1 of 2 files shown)


internal/ingest/ingest.go
```diff
@@ -716,16 +753,17 @@ func (ing *Ingester) ingestWorker() {
 		case <-ing.closeWorkers:
 			return
 		case provider := <-ing.toWorkers:
-			ing.ingestWorkerLogic(peer.ID(provider))
+			pid := peer.ID(provider)
+			pc := ing.providersBeingProcessed[pid]
+			pc <- struct{}{}
```
gammazero, line 758: Looks like there is still a race between when processing the ad chain finishes and the channel is cleared, [here](https://github.com/filecoin-project/storetheindex/commit/cf70a62e2ece09f18021fa940cb0c9fe2238d045#diff-50d2db719b550ec2183b831d2727d9184635a5647247ea8872ff82d91b6e31bdR432), and when the channel is again blocked for processing entries on the line above.

What happens if a new announce arrives after the channel is cleared, and before the channel is blocked above?   It looks execution would be blocked here waiting to write to the channel until the newly arrived announce was processed.  Could that cause a deadlock because then the `ing.toWorkers` channed is not being read?

PR #551, opened 2022-06-02
Enable automatic config reload

gammazero:
- Set the environ variable to enable automatic reload when config file changes
- Rename environ var to `STORETHEINDEX_WATCH_CONFIG`
- Longer time for graceful shutdown


willscott (approved):

command/daemon.go
```diff
@@ -40,7 +40,7 @@ const (
 	// configCheckInterval is the time between config update checks.
 	configCheckInterval = 30 * time.Second
 	// shutdownTimeout is the duration that a graceful shutdown has to complete.
-	shutdownTimeout = 5 * time.Second
+	shutdownTimeout = 10 * time.Second
```
masih, line 43: Side note: when storetheindex runs on K8S we need to make sure that the value configured here is not less than `terminationGracePeriodSeconds` specified in statefulset pod template. The default value for it is 30 seconds. 

Ultimately `terminationGracePeriodSeconds` is the value that decides how long to wait before force killing  the pod if `shutdownTimeout` is longer.

masih:
@gammazero  The current changes will trigger redeployment in `dev` and `prod` since the base kustomisation is modified. I am going to modify this PR to limit changes to `dev` only so that we can make sure the reload works in `dev` before triggering a redeployment in `prod`.

command/flags.go
```diff
@@ -80,9 +80,11 @@ var daemonFlags = []cli.Flag{
...
 	},
 	&cli.BoolFlag{
-		Name:     "watch-config",
-		Usage:    "Watch for changes to config file and automatically reload",
-		EnvVars:  []string{"INDEXER_WATCH_CONFIG"},
+		Name:  "watch-config",
+		Usage: "Watch for changes to config file and automatically reload",
+		// Note: `INDEXER_WATCH_CONFIG` is added for backward compatibility.
+		//       Remove it ince `STORETHEINDEX_WATCH_CONFIG` is rolled out in both environments.
+		EnvVars:  []string{"INDEXER_WATCH_CONFIG", "STORETHEINDEX_WATCH_CONFIG"},
```
masih, line 87: @gammazero See commit [message](https://github.com/filecoin-project/storetheindex/pull/551/commits/96b284ed40a2d95d669c47c0490151cca8b2c818) on why we need to keep env var name changes backward compatible.

command/flags.go
```diff
@@ -80,9 +80,11 @@ var daemonFlags = []cli.Flag{
 		Required: false,
 	},
 	&cli.BoolFlag{
-		Name:     "watch-config",
-		Usage:    "Watch for changes to config file and automatically reload",
-		EnvVars:  []string{"INDEXER_WATCH_CONFIG"},
+		Name:  "watch-config",
+		Usage: "Watch for changes to config file and automatically reload",
+		// Note: `INDEXER_WATCH_CONFIG` is added for backward compatibility.
+		//       Remove it ince `STORETHEINDEX_WATCH_CONFIG` is rolled out in both environments.
```
willscott, line 86:
```suggestion
		//       Remove it once `STORETHEINDEX_WATCH_CONFIG` is rolled out in both environments.
```

willscott (approved):

masih:
Relates to https://github.com/filecoin-project/storetheindex/issues/485

Issue #53, opened 2021-08-31
Ingestion: don't ask for cid lists of retracted indexes


willscott:
When walking back from the head of an index announcement chain, if the indexer encounters a notice of retraction of a previous index, and then later encounters the previous index, it should not ask for the list of CIDs composing that previous index. Since the index has been removed, the provider may not have it any more, and will not reliably be able to supply the CID list anymore.
We should either not fail everything and ignore the failure, or not ask for the CID list at all

willscott:
Right. there will be an error expected, since the provider will not have the data anymore, but in the common case the indexer can know not request it, since in playing back the log from the most recent, it should have seen in the retraction from the same provider before seeing it advertised

gammazero:
Fixed by #553

PR #553, opened 2022-06-02
Skip ads with context deleted later in chain

gammazero:
Prevent ingesting entries from ads that are deleted by another ad later (toward head) in chain.

When the chain of ads is first read from head to tail, remember context IDs that are deleted and mark ads deeper in the chain, that have those context IDs, as needing to be skipped.  Then when processing the chain from tail to head, skip ingesting entries for the ads that are marked for skipping and mark those ads as processed.

Fixes issue #53

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -812,6 +812,10 @@ func (ing *Ingester) ingestWorkerLogic(provider peer.ID) {
 	}
 	assignment := assignmentInterface.(workerAssignment)
 
+	rmCtxID := make(map[string]struct{})
+	var skips []int
+	skip := -1
+
 	// Filter out ads that are already processed, and any earlier ads.
 	splitAtIndex := len(assignment.adInfos)
 	for i, ai := range assignment.adInfos {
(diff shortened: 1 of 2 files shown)


internal/ingest/ingest.go
```diff
@@ -825,15 +829,51 @@ func (ing *Ingester) ingestWorkerLogic(provider peer.ID) {
 			splitAtIndex = i
 			break
 		}
+		ctxIdStr := string(ai.ad.ContextID)
```
willscott, line 832: is there a reason to key the map as `string` rather than `[]byte`?

internal/ingest/ingest.go
```diff
@@ -825,15 +829,51 @@ func (ing *Ingester) ingestWorkerLogic(provider peer.ID) {
...
+		// This ad was deleted by a later remove. Push previous onto skips
+		// stack, and set latest skip.
+		if _, ok := rmCtxID[ctxIdStr]; ok {
```
willscott, line 835: there's something odd about this skip stack:
* the first `skip` that will be put into `skips` will be '-1', the default value.
* that will then never matched in the subsequent loop through `i`, meaning that there seems to be at least 1 entry that isn't going to be skipped when it does match.

internal/ingest/ingest.go
gammazero: Map keys may be of any type that is comparable. This does not include slices, maps, and functions.

willscott (approved):

PR #554, opened 2022-06-02
Confirm that invalid multihashes are ingested by the indexer

gammazero:
Write a test that confirms advertisements with invalid multihashes are
ingested and discoverable by the indexer.

**Only merge this if we _want_ to allow invalid multihashes**. Compare with #555

### DIFF
--- a/internal/ingest/invalid_mh_ingest_test.go
+++ b/internal/ingest/invalid_mh_ingest_test.go
@@ -1,0 +1,53 @@
+package ingest
+
+import (
+	"context"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


masih:
Might be easier to rebase [this](https://github.com/filecoin-project/storetheindex/pull/555) branch with the test branch. That way both the fix and the test will be in the same PR, and that PR will have 2 commits: one is the test, and the other is the fix.

gammazero:
The fix that I think we want is in PR #555, and it has its own version of this test.  So, I think we can just close this PR.  Please close if you concur.

PR #555, opened 2022-06-02
Ignore invalid multihashes

gammazero:
Test that invalid multihashes are ignored.

**Only merge this if we _do not want_ to allow invalid multihashes**. Compare with #554

### DIFF
--- a/test/typehelpers/typehelpers.go
+++ b/test/typehelpers/typehelpers.go
@@ -2,11 +2,11 @@ package typehelpers
 
 import (
 	"fmt"
+	"github.com/filecoin-project/storetheindex/test/util"
 	"math/rand"
 	"testing"
 
 	"github.com/filecoin-project/storetheindex/api/v0/ingest/schema"
-	"github.com/filecoin-project/storetheindex/test/util"
 	"github.com/ipld/go-ipld-prime"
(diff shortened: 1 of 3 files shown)


willscott (approved):
ignoring bad mh's seems like the preferable path to me

test/typehelpers/typehelpers.go
```diff
@@ -197,10 +209,11 @@ func AllMultihashesFromAdChainDepth(t *testing.T, ad *schema.Advertisement, lsys
...
 			}
-			_, mh, err := multihash.MHFromBytes(b)
-			if err != nil {
-				return err
-			}
+			mh := multihash.Multihash(b)
+			//_, mh, err := multihash.MHFromBytes(b)
+			//if err != nil {
+			//	return err
+			//}
```
masih, line 216: These look like comments from my test branch; please feel free to remove the commented codes. Thanks

test/typehelpers/typehelpers.go
```diff
@@ -2,11 +2,11 @@ package typehelpers
 
 import (
 	"fmt"
+	"github.com/filecoin-project/storetheindex/test/util"
```
masih, line 5: Group with non go sdk imports?

PR #564, opened 2022-06-08
Log levels are configurable in config file and reloadable

gammazero:
Log levels are now configurable in the config file. An overall log level setting is configurable, as well as the settings for any logger that should not have the same level as the overall setting. These config settings are reloaded when the config file is reloaded at runtime.

If there is no logging configuration, the default log level is set to "info" and the loggers "basichost", "bootstrap", "dt_graphsync", "dt-impl", and "graphsync" are set to level "warn".

### DIFF
--- a/config/logging.go
+++ b/config/logging.go
@@ -1,0 +1,41 @@
+package config
+
+// Logging configures overall and logger-specific log levels. Level values are
+// case-insensitive and include the following in order of importance: "fatal",
+// "panic", "dpanic", ""error", "warn", "info", "debug"
+type Logging struct {
+	// Level sets the log level for all loggers that do not have a setting in
+	// Loggers. The default value is "info".
+	Level string
+	// Loggers sets log levels for individual loggers.
+	Loggers map[string]string
+}
+
+// NewLogging returns Logging with values set to their defaults.
+func NewLogging() Logging {
+	return Logging{
+		Level: "info",
+		Loggers: map[string]string{
+			"basichost":    "warn",
+			"bootstrap":    "warn",
+			"dt_graphsync": "warn",
+			"dt-impl":      "warn",
+			"graphsync":    "warn",
+		},
+	}
+}
+
+// populateUnset replaces zero-values in the config with default values.
+func (c *Logging) populateUnset() {
+	def := NewLogging()
+
+	if c.Level == "" {
+		c.Level = def.Level
+
+		// If not default logging level was set, and no level set for loggers,
+		// then set levels for certain loggers.
+		if len(c.Loggers) == 0 {
+			c.Loggers = def.Loggers
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


config/logging.go
```diff
@@ -0,0 +1,41 @@
+package config
+
+// Logging configures overall and logger-specific log levels. Level values are
+// case-insensitive and include the following in order of importance: "fatal",
+// "panic", "dpanic", ""error", "warn", "info", "debug"
+type Logging struct {
+	// Level sets the log level for all loggers that do not have a setting in
+	// Loggers. The default value is "info".
+	Level string
```
masih, line 9: I'm curious why make an exceptional case for logging subsystem `*` here? Why not add a default of `info` for `*` subsystem in the map below and let the user override it like any other subystem?

config/logging.go
```diff
@@ -0,0 +1,41 @@
...
+// NewLogging returns Logging with values set to their defaults.
+func NewLogging() Logging {
+	return Logging{
+		Level: "info",
+		Loggers: map[string]string{
+			"basichost":    "warn",
+			"bootstrap":    "warn",
+			"dt_graphsync": "warn",
+			"dt-impl":      "warn",
+			"graphsync":    "warn",
```
masih, line 23: It might be more flexible to expose regex-based log level configuration and set it using `SetLogLevelRegex` instead of `SetLogLevel`.

masih (approved):
One thing I thought of is potentially conflicting settings specified via [`GOLOG_*` env vars](https://github.com/ipfs/go-log#environment-variables), since they are read internally by go-log. But then I suspect storetheindex config ends up overriding any conflicting settings?

config/logging.go
gammazero: My preference is to match exactly. If the user does not know all the possible logger names, then a regex could unintentionally match a name and some important message would be discarded.

config/logging.go
gammazero: That would also work, but using "*" may imply globbing or regex name matching, which I do not want to do (see next comment).  Do you think that without globbing that using "*" would still be better?

masih: I have no preference. Both work :)

config/logging.go
masih: The downside is that people whom know exactly what they want the logging to do would be forced to list subsystems individually which could become painful for core libraries like libp2p with multiple sub packages. 

We can certainly leave it as is and change when needed. This is by no means a blocker.

gammazero:
@masih  | But then I suspect storetheindex config ends up overriding any conflicting settings?

The `GOLOG_LOG_LEVEL` does currently override the overall log level, but only because it is used as a flag environ var [here](https://github.com/filecoin-project/storetheindex/blob/main/command/flags.go#L36).  I will remove this flag so that _only_ the config file has authority. Otherwise, if that flag is used or the environ var is set, then the overall log level can not be changed by reloading the config.

PR #569, opened 2022-06-10
More config items

gammazero:
## Context
Some additional time settings were moved into the config file in order to change them without a software update. The `ShutdownTimeout` may need to be adjusted as it can take longer to shutdown an indexer if it is very busy.

Automatic config reload is now enabled by default.  There really is no reason to have this disabled, and it is much worse to accidentally not enable it should it become necessary to reload config items without a restart.

Fixed a problem configuring indexer core cache to be disabled.  Previously, setting the cache size to 0 was supposed to disable the cache, but instead only configured the default.  Now, `-1` is used to disable the cache, and the meaning of value 0 (use default) is consistent with all other `Indexer` config items.

## Proposed Changes
- Move `CheckConfigInterval` into config file
- Move `ShutdownTimeout` into config file
- Make the above values reloadable
- Fix the ability to configure disabled core cache
- Update config doc to describe logging config
- Enable `watch-config` by default

## Tests
- Manual tests of cache size configuration, `ConfigCheckInterval`, and `ShutdownTimeout`.
- Manual test to check the watch-config on by default.

## Revert Strategy
Change is safe to revert.

### DIFF
--- a/command/flags.go
+++ b/command/flags.go
@@ -63,12 +63,10 @@ var daemonFlags = []cli.Flag{
 		Required: false,
 	},
 	&cli.BoolFlag{
-		Name:  "watch-config",
-		Usage: "Watch for changes to config file and automatically reload",
-		// Note: `INDEXER_WATCH_CONFIG` is added for backward compatibility.
-		//       Remove it once `STORETHEINDEX_WATCH_CONFIG` is rolled out in both environments.
-		EnvVars:  []string{"INDEXER_WATCH_CONFIG", "STORETHEINDEX_WATCH_CONFIG"},
-		Value:    false,
+		Name:     "watch-config",
+		Usage:    "Watch for changes to config file and automatically reload",
+		EnvVars:  []string{"STORETHEINDEX_WATCH_CONFIG"},
+		Value:    true,
 		Required: false,
 	},
 }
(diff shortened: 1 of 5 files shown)


command/flags.go
```diff
@@ -64,12 +63,10 @@ var daemonFlags = []cli.Flag{
...
 	&cli.BoolFlag{
-		Name:  "watch-config",
-		Usage: "Watch for changes to config file and automatically reload",
-		// Note: `INDEXER_WATCH_CONFIG` is added for backward compatibility.
-		//       Remove it once `STORETHEINDEX_WATCH_CONFIG` is rolled out in both environments.
-		EnvVars:  []string{"INDEXER_WATCH_CONFIG", "STORETHEINDEX_WATCH_CONFIG"},
-		Value:    false,
+		Name:     "watch-config",
+		Usage:    "Watch for changes to config file and automatically reload",
+		EnvVars:  []string{"STORETHEINDEX_WATCH_CONFIG"},
```
willscott, line 68: we've switched over to the new env var on deployments at this point?

willscott (approved):

command/flags.go
gammazero: Yes, confirmed this was also done on prod

PR #574, opened 2022-06-14
Add command to import providers from another indexer

gammazero:
The "admin import-providers" subcommand imports providers from another indexer. The other indexer is specified using the `--from` flag and specifying the host:port of the other indexer.

After importing publishers, sync with them immediately.


willscott:
an immediate sync is probably a good idea

Issue #420, opened 2022-04-22
Support a new prefix-map entries format for ingest


willscott:
currently, the entries cid of an advertisement being ingested is expected to be structured as a list in the form `[[mh, mh,...], next]`.

In addition, it would be useful to allow support for entries to be structured in the form `{'prefix':next, 'prefix':next, ...}`.

This would allow for efficient distribution of subsets of advertisement data to different index shards handling different prefix-spaces.

gammazero:
Would the prefix length need to be specified by the indexer (indexer federation) since only the indexers would know their current sharding configuration?  Or, are the advertisement entries communicated in something more like a trie (prefix tree) letting the indexers decide how far down the tree to walk to determine which shard a multihash is part of?

willscott:
I think i was imagining a trie, or that even if the hamt-ness of a provider doesn't quite match the layering of a given index instance, it'll still be better than no sharding

PR #578, opened 2022-06-16
Ingest Entries of type IPLD HAMT

masih:
Expand the indexer protocol to accept and ingest `Entries` link in
advertisements that point to IPLD HAMT ADL in addition to the existing
interlinked chain of `EntryChunk`s. This is to facilitate a
backward-compatible ingestion logic that not only can process the
existing chains but also allows the entries link kind to differ across
advertisements in a single chain.

The entries processing logic is modified to automatically detect the
kind of entries and process them accordingly. The logic:
 1. Syncs the root node to which the `Entries` link in an advertisement
    points.
 2. Checks the kind of the synced node.
 3. If HAMT:
  a. Recursively syncs any remaining links in the HAMT.
  b. Batches the keys in the hamt in `Ingest.StoreBatchSize` sizes.
  c. Sends each batch to the existing indexing logic.
  d. Deletes the HAMT from the local datastore.
 4. If `EntryChunk`:
  a. Processes the root node that is synced already as before.
  b. recursively syncs through remaining interlinked `EntryChunk`s as
     before.

The maximum entries recursion depth is modified to 1 plus the configured
max depth. This is because the root entries link is synced separately to
probe the kind of entries we are dealing with before syncing the
remaining chain. A `TODO` is left in code to investigate whether it's
worth modifying the selector on the fly and reducing the recursion depth
by one so that the maximum configured value is respected.

The segmented sync for HAMT entries is disabled and postponed to future
work in order to reduce the scope of this change. A `TODO` is left to
investigate how useful the segmented sync would be for a HAMT before
implementing the logic.

Testing utilities are introduced to generate ad chains with HAMT
entries. All existing tests are modified to use a mix of HAMT and
`EntryChunk` chains for testing purposes.

The IPLD selectors used by the ingester are refactored into a struct
variable better re-usability/readability.

Note that the dependencies use specific versions of the latest
go-data-transfer, go-ipld-prime, and go-ipld-adh-hamt; none of which
have been released yet. It is not possible to use the latest released
versions due to the need for bug fixes.

Fixes: #420


go.mod
```diff
@@ -8,24 +8,25 @@ require (
...
 	github.com/hashicorp/go-retryablehttp v0.7.1
-	github.com/ipfs/go-cid v0.1.0
+	github.com/ipfs/go-cid v0.2.0
 	github.com/ipfs/go-datastore v0.5.1
 	github.com/ipfs/go-delegated-routing v0.2.2
 	github.com/ipfs/go-ds-leveldb v0.5.0
 	github.com/ipfs/go-ipfs v0.12.1
 	github.com/ipfs/go-log/v2 v2.5.1
-	github.com/ipld/go-ipld-prime v0.16.0
+	github.com/ipld/go-ipld-adl-hamt v0.0.0-20220614114547-20a727441376
```
masih, line 22: Depends on the latest fixes in https://github.com/ipld/go-ipld-adl-hamt/pull/41

go.mod
```diff
@@ -8,24 +8,25 @@ require (
...
+	github.com/ipld/go-ipld-prime v0.16.1-0.20220607093021-c88f0b441e80
```
masih, line 23: Depends on unreleased commit in `master` that contains `bindnode` improvements/fixes.

go.mod
```diff
@@ -59,7 +60,7 @@ require (
 	github.com/elastic/gosigar v0.12.0 // indirect
 	github.com/felixge/httpsnoop v1.0.1 // indirect
 	github.com/filecoin-project/go-cbor-util v0.0.0-20191219014500-08c40a1e63a2 // indirect
-	github.com/filecoin-project/go-data-transfer v1.15.1 // indirect
+	github.com/filecoin-project/go-data-transfer v1.15.2-0.20220615120701-59f73123719f // indirect
```
masih, line 63: Depends on a fix that would have otherwise caused a panic with the latest ipld-prime dependency, fixed here: https://github.com/filecoin-project/go-data-transfer/pull/335

internal/ingest/linksystem.go
```diff
@@ -204,35 +213,146 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 
 	startTime := time.Now()
 
-	// Traverse entries based on the entries selector that limits recursion depth.
+	// The ad.Entries link can point to either a chain of EntryChunks or a HAMT.
+	// Sync the very first entry so that we can check which type it is.
+	// Note, this means the maximum depth of entries traversal will be 1 plus the configured max depth.
+	// TODO: See if it is worth detecting and reducing depth the depth in entries selectors by one.
```
masih, line 219: We _can_ detect if the selector has a depth configured and reduce the depth by one similar to the work I did [here](https://github.com/filecoin-project/go-legs/blob/main/selector.go#L78-L172). But I am not sure if it is worth doing. 

Left a TODO to revisit, later since regardless of whether it's worth doing or not I'd like to attend to it in a separate PR to reduce the scope of work here.

masih (commented):
Self review.

I have randomly changed the tests to use a mix of entries kinds for better coverage and building up confidence in the correctness of logic as well as backward compatibility.

api/v0/ingest/schema/schema.ipldsch
```diff
@@ -18,7 +18,15 @@ type Advertisement struct {
     Addresses [String]
     # Signature is the signature of this advertisement.
     Signature Bytes
-    # Entries is a link to the chained EntryChunk instances that contain the multihashes advertised. 
+    # Entries is a link to a data structure that contains the advertised multihashes.
```
willscott, line 21: Update also in https://github.com/filecoin-project/storetheindex/blob/main/doc/ingest.md ?

internal/ingest/linksystem.go
willscott: todo for now seems fine

internal/ingest/linksystem.go
```diff
@@ -204,35 +213,146 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
...
+	if isHAMT(node) {
+		log = log.With("entriesKind", "hamt")
+		// Keep track of all CIDs in the HAMT to remove them later when the processing is done.
+		// This is equivalent behaviour to ingestEntryChunk which removes an entry chunk right afrer
+		// it is processed.
+		hamtCids := []cid.Cid{syncedFirstEntryCid}
+		gatherCids := func(_ peer.ID, c cid.Cid, _ legs.SegmentSyncActions) {
+			hamtCids = append(hamtCids, c)
+		}
+		defer func() {
```
willscott, line 240: i don't see where this method would have previously remove entry chunks before return,
and i don't see where it does it in the non-hamt case. Why is it different here?

internal/ingest/linksystem.go
```diff
@@ -449,10 +571,16 @@ func decodeIPLDNode(codec uint64, r io.Reader, prototype ipld.NodePrototype) (ip
...
-// Checks if an IPLD node is an advertisement, by looking to see if it has a
+// isAdvertisement checks if an IPLD node is an advertisement, by looking to see if it has a
 // "Signature" field. Additional checks may be needed if the schema is extended
 // with new types that are traversable.
 func isAdvertisement(n ipld.Node) bool {
 	indexID, _ := n.LookupByString("Signature")
 	return indexID != nil
 }
+
+// isHAMT checks if the given IPLD node is a HAMT by looking for "HashAlg" field in the given node.
```
willscott, line 582: comment and behavior are different

willscott (approved):

internal/ingest/linksystem.go
masih: [Here](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/linksystem.go#L259-L268) it is.

PR #591, opened 2022-07-05
Add a /stats endpoint with an estimate of the number of entries

willscott:
not sure i got this fully plumbed through, but and endpoint with this estimate is desired to be public for the refreshed ux.


gammazero:
There are two multihash counts that are interesting for describing indexing:
- Number of unique multihashes
- Number of multihash-to-metadata mappings (number of indexes)

The first tells how many unique pieces of data are indexed, and the second tells how much indexing information is stored in the indexer.  Consider the case of an indexer that indexes the same data for two providers. The number of unique multihashes will be half the number of multihash-to-metadata mappings.

I think this PR tries to capture the second metric, but is a very rough approximation without a clear margin of error.  I think an error factor could be computed by starting a new indexer and comparing the exact count of indexes to the value computed by this PR. Without knowing that error factor I have no idea if this PR is close enough to the actual number to be considered passable.

I can create another PR to instrument the indexer core with the ability to get exact counts of the index metrics above when starting from an empty indexer.  Then we can see how close much this estimate diverges from the actual.  If we just need the endpoint now and are willing to accept unknown inaccuracy knowing it will get better, then we can merge this, after fixing build errors.

willscott:
the plan you propose sounds good. Having an endpoint now that the ux refresh can build against will be useful. being able to update this to a more accurate count of unique multihashes would be preferable, but can happen over a longer timeframe.

PR #596, opened 2022-07-11
add http router mux that was missed

willscott:
we were missing a connecting link in the recently added stats handler


willscott:
closing in favor of #597

PR #598, opened 2022-07-11
update dev to not only allow specific ids

willscott:


gammazero (approved):
Removing the whitelisting is going to cause a rush of indexing.  It may be preferable to allow additional peer IDs.

Rebase on main to remove build failures

willscott:
I think we want dev to eventually be a non-permissioned instance to allow new providers to be able to test against. we need to open it up at some point to do that

PR #603, opened 2022-07-19
don't return providers after polling has stopped

### LINKED ISSUE #594: Filter provider from responses during unsuccessful polling
After receiving no updates for a provider for the configured `PollInterval` time, the indexer will start polling the provider's last publisher to get the latest advertisement.  This continues for up to `PollStopAfter` time, as long as the publisher cannot be reached, and there have been no updates for the provider.  

After the first unsuccessful poll, the provider should be filtered from find responses as well as in both the list and get responses to `/providers`. The provider should not be included in the responses until a response from the publisher has been received.

### LINKED ISSUE #595: Remove provider content after unsuccessful polling
After polling a provider's publisher unsuccessfully and reaching the `PollStopAfter` time, polling is stopped.  At this point the provider's index content should also be removed from the indexer.

willscott:
## Context
Providers are currently listed at `/providers` even when not live

## Proposed Changes
After polling has started and still unresponsive after at least one poll attempt, exclude provider from find responses and from `/providers` in both list and get responses (#594)
 
When polling has stopped with no responses remove the provider data from the indexer. (#595)
- Remove provider from registry's list of providers, in memory and datastore.
- Remove latest sync data (for provider's publisher) from ingester datastore.
- Remove handler from legs subscriber.
- Remove provider content from indexer core. Delete by contextID when removed provider seen in query result.

Fixes #594 
Fixes #595

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -83,7 +83,7 @@ func daemonCommand(cctx *cli.Context) error {
 	}
 
 	// Create a valuestore of the configured type.
-	valueStore, err := createValueStore(cfg.Indexer)
+	valueStore, err := createValueStore(cctx.Context, cfg.Indexer)
 	if err != nil {
 		return err
 	}
@@ -455,7 +455,7 @@ func fileChanged(filePath string, modTime time.Time) (time.Time, bool, error) {
 	return modTime, false, nil
 }
 
-func createValueStore(cfgIndexer config.Indexer) (indexer.Interface, error) {
+func createValueStore(ctx context.Context, cfgIndexer config.Indexer) (indexer.Interface, error) {
 	dir, err := config.Path("", cfgIndexer.ValueStoreDir)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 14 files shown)


willscott:
part of #594

gammazero (commented):
There is a bit more that needs to be done here:
- Should remove provider from registry's list of providers
- Should remove provider from indexer core (either delete or filter from response).
- Update datastore so provider does not return upon indexer start
- Tests.

I will take over PR and add these.

willscott (commented):
(can't approve since i started the PR, but +1 to merge)

Would be nice to add a test to confirm removal of a PR after the timeout elapses

gammazero:
Existing tests ensure that provider is removed from registry and that correct request is made to ingester to remove publisher info.  Added a new tests to ensure that after removing provider from registry that get provider returns not found and multihash query returns nothing for removed provider.

gammazero (approved):

PR #608, opened 2022-07-22
expand dev disk

willscott:
Like #607  but for dev


willscott:
is it dev-v2 or just dev?

masih (approved):

rvagg (approved):

Issue #612, opened 2022-07-28
cloudfront is stripping headers


willscott:
The internal golang server sets some relevant headers, e.g. https://github.com/filecoin-project/storetheindex/blob/main/internal/httpserver/util.go#L17-L19

These are not returned to clients, and appear to be removed at the cloudfront level.
We need to allow them through in order to get expected JS api access and caching.

masih:
This is not a cloudfront issue. The finder server does not seen to return the headers set as part of response handling. Digging deeper. I think this is a gorilla/mux issue

masih:
The finder server responses are fixed and deployed in `dev`. It'll take a bit of time for the deployment to roll out. Once that's done, I'll pick this issue back up, there may be changes we need in cloudfront too. Unclear until deployment rollout is complete.

masih:
It looks like the changes in code did the trick, no need for cloudfront changes. see CORS test [here](https://cors-test.codehappy.dev/?url=https%3A%2F%2Fdev.cid.contact%2Fstats&method=get)

The response content type is now fixed too:

```bash
$ curl -sSL -D - https://dev.cid.contact/providers -o /dev/null        

HTTP/2 200 
date: Thu, 04 Aug 2022 15:17:43 GMT
content-type: application/json; charset=utf-8
access-control-allow-methods: GET,OPTIONS
access-control-allow-origin: *
strict-transport-security: max-age=15724800; includeSubDomains
```

PR #618, opened 2022-08-04
Fix handler content type and add CORS tests

masih:
Fix an issue where the set `Content-Type` by `WriteJsonResponse` handler
had no impact since HTTP status was being set before the headers.

Simplify the CORS handing in `mux` router and add tests that assert the
CORS handling as well as expected content type for every finder
endpoint.

Relates to:
- #612

### DIFF
--- a/internal/httpserver/util.go
+++ b/internal/httpserver/util.go
@@ -13,8 +13,8 @@ import (
 var log = logging.Logger("indexer/http")
 
 func WriteJsonResponse(w http.ResponseWriter, status int, body []byte) {
-	w.WriteHeader(status)
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(status)
 	if _, err := w.Write(body); err != nil {
 		log.Errorw("cannot write response", "err", err)
 		http.Error(w, "", http.StatusInternalServerError)
--- a/server/finder/http/server.go
+++ b/server/finder/http/server.go
@@ -49,53 +49,28 @@ func New(listen string, indexer indexer.Interface, registry *registry.Registry, 
 	h := newHandler(indexer, registry)
 
 	// Client routes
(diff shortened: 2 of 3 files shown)


internal/httpserver/util.go
```diff
@@ -13,8 +13,8 @@ import (
 var log = logging.Logger("indexer/http")
 
 func WriteJsonResponse(w http.ResponseWriter, status int, body []byte) {
-	w.WriteHeader(status)
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(status)
```
masih, line 17: @willscott  this was the root of all evil.

server/finder/http/server.go
```diff
@@ -49,34 +49,22 @@ func New(listen string, indexer indexer.Interface, registry *registry.Registry,
...
+	r.HandleFunc("/providers/{providerid}", h.getProvider).Methods(http.MethodGet)
+	r.HandleFunc("/stats", h.getStats).Methods(http.MethodGet)
 
 	reframeHandler := reframe.NewReframeHTTPHandler(indexer, registry)
 	r.HandleFunc("/reframe", reframeHandler)
 
+	cors := handlers.CORS(handlers.AllowedOrigins([]string{"*"}))
 	server := &http.Server{
-		Handler:      r,
+		Handler:      cors(r),
```
masih, line 67: @gammazero I'd love your eagle eye on the changes here. I think the changes I have made simplifies the way mux router is set up while keeping the intended functionality the same. Many thanks

willscott (approved):

server/finder/http/server.go
gammazero: I do not know why we did not do this in the first place.  I am guessing we thought we did not want to have CORS for all endpoints.  Anyway, this is much cleaner and looks correct.

PR #621, opened 2022-08-05
Configure the web page of the index page

willscott:
* load the static external frame as a template and inject in a new config option to set the domain.

### DIFF
--- a/config/addresses.go
+++ b/config/addresses.go
@@ -8,6 +8,9 @@ type Addresses struct {
 	// Finder is the finder http isten address. Set to "none" to disable this
 	// server for both http and libp2p.
 	Finder string
+	// FinderWebpage is a domain to display when the homepage of the finder
+	// is accessed over http.
+	FinderWebpage string
 	// Ingest is the index data ingestion http listen address. Set to "none"
 	// to disable this server for both http and libp2p.
 	Ingest string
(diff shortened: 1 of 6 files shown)


config/addresses.go
```diff
@@ -21,10 +24,11 @@ type Addresses struct {
 // NewAddresses returns Addresses with values set to their defaults.
 func NewAddresses() Addresses {
 	return Addresses{
-		Admin:   "/ip4/127.0.0.1/tcp/3002",
-		Finder:  "/ip4/0.0.0.0/tcp/3000",
-		Ingest:  "/ip4/0.0.0.0/tcp/3001",
-		P2PAddr: "/ip4/0.0.0.0/tcp/3003",
+		Admin:         "/ip4/127.0.0.1/tcp/3002",
+		Finder:        "/ip4/0.0.0.0/tcp/3000",
+		FinderWebpage: "https://web.cid.contact/",
```
masih, line 29: We should parameterise this and set it per environment. What are the impacts of hard coding this for:
* people whom want to run their own indexer under some other domain
* `dev` vs `prod` deployment

masih (approved):
I think this is a stopgap. What'd recommend doing is to pull out static HTML or visual stuff completely out of the container and handle it all externally via K8S and AWS control plane.  The container service should _only_ be exposing the API.

The undesirables in the current approach are:
* `cid.contact` hardcoded into the container which will be problematic for anyone wanting to run their own indexer
* Though better than before, we are not completely safe from having to redeploy the entire service just to change visual sugar on the landing page.

config/addresses.go
willscott: this is meant as a default - are these structs not loaded from per-instance configs?

masih: If the default is not overridden in corresponding configs e.g. [here](https://github.com/filecoin-project/storetheindex/blob/main/deploy/manifests/dev/us-east-2/tenant/storetheindex/indexer-config.yaml#L14) both dev and prod end using the same path. Is that the intention? If so, how would the new UI know which API to call?

willscott: https://github.com/willscott/cid.contact/blob/main/index.html#L247-L248

willscott: I'm happy to also change the dev config you point to to have it embed a different root

masih: I see thank you

masih: I honestly think we should rethink this and pull UI stuff out. but for now iframe seems like a reasonable approach to get  the new UI rolled out.

PR #627, opened 2022-08-07
Registration of new publisher with no addresses

gammazero:
Handle registering a new publisher for a provider, when no new provider addresses are given. Handle the case where the publisher is different from the publisher, and where the provider is the same as the publisher.

This also fixes another issue where the new provider addresses, if given, need to be applied before setting an empty publisher address to that of the provider with the same ID.

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -448,9 +448,10 @@ func (r *Registry) RegisterOrUpdate(ctx context.Context, providerID peer.ID, add
 	if len(addrs) != 0 {
 		maddrs, err := stringsToMultiaddrs(addrs)
 		if err != nil {
-			panic(err)
+			log.Errorw("Invalid provider address", "err", err)
+		} else if len(maddrs) != 0 {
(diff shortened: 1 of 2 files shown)


internal/registry/registry.go
```diff
@@ -442,6 +445,16 @@ func (r *Registry) RegisterOrUpdate(ctx context.Context, providerID peer.ID, add
 		}
 	}
 
+	if len(addrs) != 0 {
+		maddrs, err := stringsToMultiaddrs(addrs)
+		if err != nil {
+			panic(err)
```
willscott, line 451: the addrs are supplied by the provider - it's presumably a failure to ingest but not a panic if they're not valid?

gammazero: If an invalid multiaddress got this far, then it must be a programming error. However, anyone who sees that would want to re-validate that is the case, so yeah, better to handle it as an error.

willscott (approved):

PR #638, opened 2022-08-10
Added more metrics into the sync process

ischasny:
Added more metrics into sync / ingestion process

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -308,6 +308,9 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 		log = log.With("entriesKind", "EntryChunk")
 		// We have already peaked the first EntryChunk as part of probing the entries type.
 		// So process that first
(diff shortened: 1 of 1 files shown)


internal/ingest/linksystem.go
```diff
@@ -348,6 +355,8 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 				return adIngestError{adIngestSyncEntriesErr, fmt.Errorf("failed to sync entries: %w", err)}
 			}
 		}
+
+		log = log.With("ingestTotalTime", ingestTotalTime/1000.0).With("totalChunksCount", totalChunksCount)
```
masih, line 359: This is a great start thanks @ischasny 👍 
* I think we need grafana metrics, so it would be great to report them that way too.
* we need metrics for both HAMT and EntryChunk processing time, tagged by entry kind.

ischasny: Yes, agreed. That was a quick attempt to understand where the problem is, not the final solution by any means :)

PR #644, opened 2022-08-10
Filter non-routable IP addresses

gammazero:
Remove any private, loopback, and unspecified IP addresses from provider and publisher addresses when they are written to the registry.

### DIFF
--- a/deploy/manifests/base/storetheindex/indexer-config.yaml
+++ b/deploy/manifests/base/storetheindex/indexer-config.yaml
@@ -44,6 +44,7 @@ data:
         "Dir": "/data/datastore"
       },
       "Discovery": {
+        "FilterIPs": true,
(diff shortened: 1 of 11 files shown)


deploy/manifests/base/storetheindex/indexer-config.yaml
```diff
@@ -44,6 +44,7 @@ data:
         "Dir": "/data/datastore"
       },
       "Discovery": {
+        "FilterIPs": true,
```
willscott, line 47: having the key present before the deploy of the version that recognizes it won't cause problems, right?

willscott (approved):

deploy/manifests/base/storetheindex/indexer-config.yaml
gammazero: Right, it gets ignored.

PR #645, opened 2022-08-11
Start metric server at the beginning of the process

### LINKED ISSUE #323: Metrics should start sooner
Right now we wait until the end of the startup steps to start the admin http server. The admin http server runs the metrics endpoint for the process. This means we don't have any metrics until storetheindex finishes booting up.

Ideally we start a metrics server as soon as we can so we get obs into the startup process. For example, right now startup takes a long time and we are missing useful metrics (maybe the CPU and ram are pegged?).

This change involves moving the metrics handler out of admin and into its own thing. This will also involve changing all the endpoints that reference that endpoint to use the new one.

willscott:
Allows for pprof and metrics during startup

fix #323

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -82,6 +82,32 @@ func daemonCommand(cctx *cli.Context) error {
 		return fmt.Errorf("only levelds datastore type supported, %q not supported", cfg.Datastore.Type)
 	}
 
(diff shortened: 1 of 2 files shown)


ischasny (approved):
LGTM

PR #648, opened 2022-08-11
Allow init with a different STH bit prefix

willscott:
currently this is fixed at 24, but may be worth experimenting with a bigger version

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -476,7 +476,12 @@ func createValueStore(ctx context.Context, cfgIndexer config.Indexer) (indexer.I
 
 	switch cfgIndexer.ValueStoreType {
 	case vstoreStorethehash:
-		return storethehash.New(ctx, dir, storethehash.GCInterval(time.Duration(cfgIndexer.GCInterval)))
+		return storethehash.New(
+			ctx,
+			dir,
+			storethehash.GCInterval(time.Duration(cfgIndexer.GCInterval)),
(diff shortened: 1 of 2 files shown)


command/daemon.go
```diff
@@ -476,7 +476,12 @@ func createValueStore(ctx context.Context, cfgIndexer config.Indexer) (indexer.I
 
 	switch cfgIndexer.ValueStoreType {
 	case vstoreStorethehash:
-		return storethehash.New(ctx, dir, storethehash.GCInterval(time.Duration(cfgIndexer.GCInterval)))
+		return storethehash.New(
+			ctx,
+			dir,
+			storethehash.GCInterval(time.Duration(cfgIndexer.GCInterval)),
```
masih, line 482: What would happen if bitesize is changed on existing data? 

We probably should error out, and maybe a TODO to do a migration ?

config/indexer.go
```diff
@@ -34,6 +36,7 @@ func NewIndexer() Indexer {
 		ShutdownTimeout:     Duration(10 * time.Second),
 		ValueStoreDir:       "valuestore",
 		ValueStoreType:      "sth",
+		STHBits:             24,
```
masih, line 39: Add comment about changing this value for an existing datastore and its consequences?

masih (approved):

command/daemon.go
willscott: https://github.com/filecoin-project/go-indexer-core/issues/62

PR #656, opened 2022-08-12
Run prod instances explicitly on `r5b.4xl` node and increase resources

masih:
Now that there are `r5b.4xl` worker nodes in the cluster, set affinity
for `prod` nodes to them and increase the memory and CPU resources to
utilise what the worker nodes offer.


willscott (approved):
looks right. how do you roll it out one at a time?

masih:
> how do you roll it out one at a time?

That's handled by k8s just like any other deployment

PR #661, opened 2022-08-13
Index gc

gammazero:
Use storethehash that has advanced GC

### DIFF
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/patch-indexer.yaml
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/patch-indexer.yaml
@@ -47,6 +47,7 @@ spec:
         - name: identity
           secret:
             secretName: indexer-identity
+      - terminationGracePeriodSeconds: 300
   volumeClaimTemplates:
     - metadata:
         name: data
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/patch-indexer.yaml
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/patch-indexer.yaml
@@ -5,6 +5,7 @@ metadata:
 spec:
   template:
     spec:
+      terminationGracePeriodSeconds: 300
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
@@ -47,7 +48,6 @@ spec:
         - name: identity
           secret:
             secretName: indexer-identity
-      - terminationGracePeriodSeconds: 300
   volumeClaimTemplates:
     - metadata:
         name: data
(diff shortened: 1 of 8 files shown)


internal/ingest/ingest.go
```diff
@@ -829,6 +839,16 @@ func (ing *Ingester) ingestWorkerLogic(provider peer.ID) {
 	// Filter out ads that are already processed, and any earlier ads.
 	splitAtIndex := len(assignment.adInfos)
 	for i, ai := range assignment.adInfos {
+		if ctx.Err() != nil {
+			log.Infow("Ingest worker canceled while ingesting ads", "provider", provider)
```
masih, line 843: log error too? Ditto in second check loop below.

internal/ingest/ingest.go
```diff
@@ -799,14 +809,14 @@ func (ing *Ingester) ingestWorker() {
 			pc := ing.providersBeingProcessed[pid]
 			ing.providersBeingProcessedMu.Unlock()
 			pc <- struct{}{}
-			ing.ingestWorkerLogic(pid)
-			ing.handlePendingAnnounce(pid)
+			ing.ingestWorkerLogic(ctx, pid)
```
masih, line 812: Should the `select` above also select on `ctx.Done()`?

internal/ingest/ingest.go
```diff
@@ -203,6 +205,8 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr indexer.Interface, reg *re
 	if cfg.IngestWorkerCount == 0 {
 		return nil, errors.New("ingester worker count must be > 0")
 	}
+
+	ing.workersCtx, ing.cancelWorkers = context.WithCancel(context.Background())
```
masih, line 209: `ing.cancelWorkers` never called?

gammazero: Oops, I meant to move it to the top of Close(), there now.

internal/ingest/ingest.go
gammazero: I think that this cannot exit while items are still being written to `ing.toWorkers`, otherwise those things will block on writing the channel.

internal/ingest/ingest.go
gammazero: Done

deploy/manifests/dev/us-east-2/tenant/storetheindex/patch-indexer.yaml
```diff
@@ -47,6 +47,7 @@ spec:
         - name: identity
           secret:
             secretName: indexer-identity
+      - terminationGracePeriodSeconds: 300
```
masih, line 50:
```suggestion
      terminationGracePeriodSeconds: 300
```

masih (approved):
Approving for trial deployment to `dev`

PR #665, opened 2022-08-16
Configure replica count in `dev` via patch for finer control

masih:
The `images` kustomization transform does not support image
transformation filtering by `kind`. Since the image name across both
single-instances are the same, the replica is changed to 2 across the
deployments and statefulsets.

Set the replica via patch for finer control over the resource it is
applied to.


ischasny (approved):
Lgtm

PR #671, opened 2022-08-16
add metric for how many ads are queued

willscott:
Q: is it okay to make this queue non-blocking?


internal/ingest/ingest.go
```diff
@@ -156,7 +157,7 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr indexer.Interface, reg *re
 
 		providersBeingProcessed: make(map[peer.ID]chan struct{}),
 		providerAdChainStaging:  make(map[peer.ID]*atomic.Value),
-		toWorkers:               make(chan providerID),
+		toWorkers:               make(chan providerID, 1000),
```
gammazero, line 160: What does this solve by putting the providers in the channel without a worker available?  Having the channel unbuffered caused backpressure on ingestion if there are no workers available.  With this change, I am concerned that it ends up loading 1000 providers into the channel, and then causing back-pressure as before.  That would use up more memory without solving any problem.

internal/ingest/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+	q.lk.Lock()
+	defer q.lk.Unlock()
+
+	if _, ok := q.states[p]; !ok {
+		q.order = append(q.order, p)
+	}
+	q.states[p] = a
+
+	if q.waiters > 0 {
+		q.notifyChan <- struct{}{}
```
gammazero, line 36: Instead of using `q.waiters` to determine if the channel can be written, what about using select:
```suggestion
		select {
		case q.notifyChan <- struct{}{}:
		default:
		}
```

That should do the same thing and then `q.waiters` is not needed.

internal/ingest/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+	// used if someone is pulling an empty queue
+	waiters    uint
+	notifyChan chan struct{}
+	doneChan   chan struct{}
+}
+
+func NewPriorityQueue() *Queue {
+	return &Queue{
+		states:     make(map[providerID]adInfo),
+		order:      []providerID{},
```
gammazero, line 19: minor: No need to initialize this.  The zero-value of `nil` for this slice will have a `len` of 0 and this will be allocated with the first call to
```go
q.order = append(q.order, p)
```
It works just fine to do: `slice = append(nil, element)`

gammazero (approved):
Comments about minor tweaks

internal/ingest/priority_queue.go
```diff
@@ -0,0 +1,95 @@
+package ingest
+
+import "sync"
+
+type Queue struct {
+	lk     sync.Mutex
+	states map[providerID]adInfo
```
gammazero, line 7: Where is the `adInfo` part of `states` getting used?

internal/ingest/priority_queue.go
```diff
@@ -0,0 +1,95 @@
...
+		case <-q.doneChan:
+			return ""
+		case <-q.notifyChan:
+		}
+		q.lk.Lock()
+		q.waiters--
+	}
+
+	p := q.order[0]
+	q.order = q.order[1:]
```
gammazero, line 56: Each time `q.order` is advanced by `q.order = q.order[1:]` and then appended to, this will allocate new underlying memory and copy the contents of the old memory into the new.  This creates unnecessary GC.  Here is an example to demonstrate: https://go.dev/play/p/6fYrZTZiUeF

Might I suggest using a ring buffer instead.  [Here is one](https://github.com/gammazero/deque#deque) that can be pushed and popped and will reuse the same memory.

willscott:
@gammazero  are you okay with me doing generics as you suggested and breaking the 1.17 ci until the CI gets upgraded to only use 1.18?

gammazero:
I am OK with the use of generics.  After all, go1.19 has been out for a while now.  If we want the tests to pass for now, you can depend on deque v0.1.2.  That uses `interface{}` instead of generics.  It will not be as efficient, but should be fine for this application.

PR #673, opened 2022-08-18
Introduce binary format valustore serialization

masih:
Upgrade to the latest `go-indexer-core` which introduces binary
serialization format for values that offers much better performance in
comparison with JSON.

Add configuration values in storetheindex so that the valuestore
serialization format can be configured, with default of JSON for
backward-compatibility.

### DIFF
--- a/config/indexer.go
+++ b/config/indexer.go
@@ -7,7 +7,7 @@ import (
 // Indexer holds configuration for the indexer core. Setting any of these items
 // to their zero-value configures the default value.
 type Indexer struct {
-	// Maximum number of CIDs that cache can hold. Setting to -1 disables the
+	// CacheSize is the maximum number of CIDs that cache can hold. Setting to -1 disables the
 	// cache.
 	CacheSize int
 	// ConfigCheckInterval is the time between config file update checks.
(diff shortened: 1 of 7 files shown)


config/indexer.go
```diff
@@ -21,14 +21,21 @@ type Indexer struct {
...
-	// Type of valuestore to use, such as "sth" or "pogreb".
+	// ValueStoreType specifies type of valuestore to use, such as "sth" or "pogreb".
 	ValueStoreType string
-	// bits for bucket size in store the hash. Note: this should not be changed
-	// from it's value at initialization or the datastore will be corrupted.
+	// STHBits is bits for bucket size in store the hash. Note: this should not be changed
+	// from its value at initialization or the datastore will be corrupted.
 	STHBits uint8
+	// ValueStoreSerde configures the marshalling format of values stored by the valuestore.
+	// It can be one of "json" or "binary". If unspecified, json format is used by default.
```
gammazero, line 33: If left unspecified, could the functionality instead be to use whatever the existing value store uses?  If there is no existing value store, then use binary by default.

masih: That needs extra engineering in go-indexer-core to probe the existing serialisation format when the file exists

gammazero: Just aksing to consider if that is worth it.

gammazero (approved):

config/indexer.go
masih: Thank you for pointing that out. Left a todo in code.

PR #680, opened 2022-08-18
Deploy a fresh `prod` instance with binary valuestore codec

masih:
Deploy `xabi` a fresh prod instance in us-east2b AZ that uses binary
serialization format with 30 bit bucket size.


deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/kustomization.yaml
```diff
@@ -8,3 +8,4 @@ resources:
   - pod-monitor.yaml
   - romi
   - tara
+  - xabi
```
willscott, line 11: some mapping of short name to behavior would be great :)

willscott (approved):

PR #688, opened 2022-08-22
Exclude Go `1.17` from CI builds

masih:
The changes in worker queueing system requires Go 1.18 and later.
Exclude Go 17.17 until the unified CI is upgraded in order to avoid CI
noise.

See:
 - https://github.com/filecoin-project/storetheindex/pull/671


willscott (approved):
hopefully this doesn't make the CI workflows unhappy with us 🤞

masih:
It opens PRs that tries to revert it back I think. We can ignore those until upgraded.

PR #690, opened 2022-08-22
Deploy latest to mya

gammazero:
This has sti queue PR and sth that uses less memory


willscott (approved):

masih:
Instances Readme needs to be updated with the version. Please feel free to remove that column if you think it's not useful

PR #694, opened 2022-08-23
Fix reframe path join issue in indexstar prod

masih:
See:
 - https://github.com/filecoin-shipyard/indexstar/pull/9


ischasny (approved):
lgtm

PR #696, opened 2022-08-23
Upgrade to latest indexstar with upgraded reframe dependency

masih:
To avoid potential DAG-JSON decoding issues when multiple find provider
results are returned from a server run the latest indexstar with
upgraded reframe dependency.


masih:
Please feel free to merge on approval if I was not online. Thanks

PR #697, opened 2022-08-24
Added Remove All message support

ischasny:
This PR adds support for Remove All ad that removes all entries for the given provider.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -839,7 +839,9 @@ func (ing *Ingester) ingestWorkerLogic(ctx context.Context, provider peer.ID) {
 	}
 	assignment := assignmentInterface.(workerAssignment)
 
-	rmCtxID := make(map[string]struct{})
+	rmCtxID := make(map[string]bool)
+	rmProviderID := make(map[string]bool)
+
 	var skips []int
 	skip := -1
 
@@ -869,7 +871,7 @@ func (ing *Ingester) ingestWorkerLogic(ctx context.Context, provider peer.ID) {
 		ctxIdStr := string(ai.ad.ContextID)
 		// This ad was deleted by a later remove. Push previous onto skips
 		// stack, and set latest skip.
-		if _, ok := rmCtxID[ctxIdStr]; ok {
+		if rmCtxID[ctxIdStr] || rmProviderID[ai.ad.Provider] {
 			skips = append(skips, skip)
 			skip = i
 			continue
@@ -877,7 +879,11 @@ func (ing *Ingester) ingestWorkerLogic(ctx context.Context, provider peer.ID) {
 		// If this is a remove, do not skip this ad, but skip all earlier
 		// (deeped in chain) ads with the same context ID.
 		if ai.ad.IsRm {
-			rmCtxID[ctxIdStr] = struct{}{}
+			if len(ai.ad.ContextID) != 0 {
+				rmCtxID[ctxIdStr] = true
+			} else {
+				rmProviderID[ai.ad.Provider] = true
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


gammazero:
When ingesting ads for a provider, if an ad is received that instructs the indexer to delete everything for the provider, normally this should not be a problem.  Normally there should not be any additional ads for that provider to ingest following the delete-all ad.  However, there are some edge cases that we want to make sure do not break the indexer.
- What if the ad chain has ad(s) following delete all ad?
- What if there is a pending announce for an ad previous to the delete all ad. The announce may have been delivered late, and if the provider was deleted it will look like the indexer needs to ingest everything for a new provider up to the announced ad.

internal/ingest/ingest.go
```diff
@@ -869,15 +871,19 @@ func (ing *Ingester) ingestWorkerLogic(ctx context.Context, provider peer.ID) {
 		ctxIdStr := string(ai.ad.ContextID)
 		// This ad was deleted by a later remove. Push previous onto skips
 		// stack, and set latest skip.
-		if _, ok := rmCtxID[ctxIdStr]; ok {
+		if rmCtxID[ctxIdStr] || rmProviderID[ai.ad.Provider] {
```
gammazero, line 874: If the ad is for deleting all data for the provider, then not only should the previous ads in the chain be skipped, but any ads in the entire chain for the provider.  Granted, there should no be any ads for a provider following an ad to delete that provider, but the indexer should catch that case.  If it does not, then it will try to ingest those ads and fail in strange ways.

I would suggest using a separate map of deleted providers, and check that in addition to checking the next ad to skip, [here](https://github.com/filecoin-project/storetheindex/pull/697/files#diff-50d2db719b550ec2183b831d2727d9184635a5647247ea8872ff82d91b6e31bdL904)

gammazero (commented):
There is an edge condition I think we should handle.

internal/ingest/ingest.go
ischasny: Makes sense, thanks. Fixed in the latest commit.

internal/ingest/ingest.go
```diff
@@ -869,15 +874,19 @@ func (ing *Ingester) ingestWorkerLogic(ctx context.Context, provider peer.ID) {
...
 			continue
 		}
 		// If this is a remove, do not skip this ad, but skip all earlier
 		// (deeped in chain) ads with the same context ID.
 		if ai.ad.IsRm {
-			rmCtxID[ctxIdStr] = struct{}{}
+			if len(ai.ad.ContextID) != 0 {
+				rmCtxID[ctxIdStr] = true
+			} else {
+				ing.deletedProviders[ai.ad.Provider] = true
```
gammazero, line 888: When an ad to delete a provider is seen, handle that ad immediately, because it is the only one in the chain we new to handle for this provider. None of this provider's other ads matter since the provider is getting deleted.
```suggestion
				// Mark the provider as deleted to skip any other ads.
				deletedProviders[ai.ad.Provider] = true
				// Process this ad now, since it is the only one that
				// matters for the provider that is being deleted.
				if markErr := ing.markAdProcessed(assignment.publisher, ai.cid); markErr != nil {
					log.Errorw("Failed to mark ad as processed", "err", markErr)
				}
				ing.inEvents <- adProcessedEvent{
					publisher: assignment.publisher,
					headAdCid: assignment.adInfos[0].cid,
					adCid:     ai.cid,
				}
			}
```

internal/ingest/ingest.go
```diff
@@ -160,6 +163,7 @@ func NewIngester(cfg config.Ingest, h host.Host, idxr indexer.Interface, reg *re
 		providerAdChainStaging:  make(map[peer.ID]*atomic.Value),
 		toWorkers:               NewPriorityQueue(),
 		closeWorkers:            make(chan struct{}),
+		deletedProviders:        make(map[string]bool),
```
gammazero, line 166: This can be local to `ingestWorkerLogic`.  I think it is only necessary to remember deleted providers during the processing of the current chain.

gammazero (commented):
Gave some more examples to explain about not using skip stack.

PR #704, opened 2022-08-25
Interpret shutdown timeout of zero as no limit

masih:
Storetheindex at runtime can have 2 shutdown timeouts: one from K8S,
i.e. the termination grace period, and one at application level. The two
can conflict each other. So on deployment to K8S one has to make sure
the value of latter is always higher than the value of former.

To avoid having to make sure the value is higher, simply interpret zero
as no timeout. That way, the user can disable shutdown timeout fully and
avoid conflicting config depending on runtime environment.

Change the default shutdown timeout to no-timeout.

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -86,32 +86,6 @@ func daemonCommand(cctx *cli.Context) error {
 		return fmt.Errorf("only levelds datastore type supported, %q not supported", cfg.Datastore.Type)
 	}
 
-	// Create admin HTTP server
-	var adminSvr *httpadminserver.Server
-	if cfg.Addresses.Admin != "" && !cctx.Bool("noadmin") {
-		maddr, err := multiaddr.NewMultiaddr(cfg.Addresses.Admin)
-		if err != nil {
-			return fmt.Errorf("bad admin address in config %s: %s", cfg.Addresses.Admin, err)
-		}
-		adminAddr, err := manet.ToNetAddr(maddr)
-		if err != nil {
-			return err
-		}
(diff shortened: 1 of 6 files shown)


gammazero (approved):
I think you want to avoid calling `cancel()` on [line 456](https://github.com/filecoin-project/storetheindex/pull/704/files#diff-4be1265b79e88b7bd2cecd63f7f6e8ff9f1a59aa806ba52635de084462cb58caR456) if it was not set with a shutdown timer. Otherwise, it is calling the wrong cancel.  I think there is no harm, but it is confusing which cancel function is supposed to be called.

masih:
I am going to remove that cancel call. both are called via defer. I don't see what's the point of explicitly calling cancel in line 456 just before return. WDYT? does that make sense?

gammazero (approved):

gammazero:
The purpose of calling that cancel is to prevent the error message and error exit here:
https://github.com/filecoin-project/storetheindex/pull/704/files#diff-4be1265b79e88b7bd2cecd63f7f6e8ff9f1a59aa806ba52635de084462cb58caR409

masih:
Does that line not explicitly check if the error is to do with timeout? Cancelation would return a different error no?

gammazero:
That is correct, cancellation returns a different error and is ignored... Which is why `cancel()` is called, so that the context does not timeout and cause the daemon to exit with error when everything has correctly shutdown.  This is probably never a problem because the daemon would need to finish shutting down  very close to the timeout.

PR #708, opened 2022-08-26
Add a minimum multihash digest length check

gammazero:
This is necessary for value stores (e.g. storethehash) that require a minimum key length. Additionally, the minimum multihash digest length can be configured in the indexer config, should some minimum size be desired.

This PR allows ingestion to proceed when there are multihashes that are too short.  Previously this would lead to ingestion failing, and the whole ad chain being reprocessed later, repeatedly, leading to the indexer backing up with all the ad chain ingestion retries.


config/ingest.go
```diff
@@ -41,6 +41,10 @@ type Ingest struct {
 	// IngestWorkerCount sets how many ingest worker goroutines to spawn. This
 	// controls how many concurrent ingest from different providers we can handle.
 	IngestWorkerCount int
+	// MinimumKeyLengt causes any multihash, that has a digest length less than
+	// this, to be ignored. If using storethehash, this value is automatically
+	// set to 4 if it was configured to be anything less.
+	MinimumKeyLength int
```
masih, line 47: This will default to `0` right? 

Should we explicitly set its value across instance configs to `4`?

masih: Side note and out of the scope of this PR: we probably want a maximum size config too to limit malicious behaviour.

masih (commented):
Looks good The only thing left is to set the min length value in instance configurations. right?

masih (approved):
I see for sth value store the minimum is automatically set to 4; thank you :+1:

PR #711, opened 2022-08-26
Peer indexers to the two main instances

gammazero:
Enable ResendDirectAnnounce for the main indexer instances. This allows them to publish direct announce messages, received over to HTTP, to the other peer indexers via pubsub.

In the dev and prod cluster, peer the individual indexers to two main indexer nodes for the cluster.

**DO NOT MERGE UNTIL `indexer-0` and `indexer-1` ADDRESSES ARE WORKING**


masih (approved):
[Updated](https://github.com/filecoin-project/storetheindex/pull/711/commits/d7c3a7329288926519348e97e682be795f1317cc) the addresses. They should be working now; no need for separate services 👍

PR #712, opened 2022-08-27
STH option to enable GC free index file scan

gammazero:


masih:
Builds on top of https://github.com/ipld/go-storethehash/pull/67


**Do Not Merge**: the changes here are being tested in `romi` and `xabi` in `prod`.

masih:
Tested; all is fine. Bubbling up the dependency from sth

masih (approved):
Thank you for adding this option so quickly @gammazero 🍻

PR #739, opened 2022-09-03
Asynchronously handle entry chunks by default

gammazero:
Handle entry chunks asynchronously, unless `Config.Ingest.SyncWriteEntries` is true. The asynchronous behavior allows the indexer to fetch the next entry chunk without having to wait for the current one to complete processing. An asynchronous goroutine (one per worker) processes the entry chunks while they are being queued.

Previously, the `EntryPutConcurrency` setting configured a pool of goroutines, shared by all workers, to handle processing entry chunks. A problem with this was that one provider/worker could submit many entry chunks to his pool and keep all the goroutines busy for itself, starving the other provider/workers of the ability to process entry chunks.

With this PR, each provider/worker has its own async goroutine and can asynchronously process entry chunks without starving any other provider/worker. The asynchronous behavior can be enabled or disabled using the `Ingest.SyncWriteEntries` setting (false=async, true=sync). This configuration value is run-time reloadable.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -17,7 +17,6 @@ import (
 	"github.com/filecoin-project/storetheindex/internal/metrics"
 	"github.com/filecoin-project/storetheindex/internal/registry"
 	"github.com/filecoin-project/storetheindex/peerutil"
-	"github.com/gammazero/workerpool"
 	"github.com/hashicorp/go-retryablehttp"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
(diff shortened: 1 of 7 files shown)


willscott (approved):

internal/ingest/ingest.go
```diff
@@ -137,9 +136,11 @@ type Ingester struct {
...
 	// Multihash minimum length
 	minKeyLen int
-
-	// entryWP is a worker pool for asynchronous calls to indexer.Put.
-	entryWP *workerpool.WorkerPool
+	// syncWriteEntries is accessed through Ingester.SetSyncWriteEntries() and
+	// Ingester.syncWriteEntries(). It tells the Ingester to handle write entry
+	// chunks synchronously, waiting for each to complete before fetching the
+	// next.
+	_syncWriteEntries uint32
```
masih, line 143: `_` prefix a new naming convention? I don't recall it's use elsewhere.

masih (approved):

internal/ingest/ingest.go
gammazero: The reason for the `_` prefix is because this member variable should not be accessed directly by things in the same package, but through the functions, `syncWriteEntries()` and `SetSyncWriteEntries()`

masih: I recommend adding that as a comment, and update the godoc to use the same name in the beginning.

PR #741, opened 2022-09-03
New version and updated config doc

gammazero:
The number of entry chunks queued for async processing is limited to 1. This prevents queuing large chains of entry chunks awaiting processing and exhausting memory.

If testing does not expose any problems, merge this so that there is a version tag associated with the async entries processing and core put concurrency.


gammazero:
Closing: replacing with non-versioned PR.

PR #742, opened 2022-09-04
Limit entry chunks queued for async processing

gammazero:
The number of entry chunks queued for async processing is limited to 1. This prevents queuing large chains of entry chunks awaiting processing and exhausting memory.

Other changes:
- Daemon logs config reload
- Documentation updates


masih (approved):

willscott:
Do we really only want 1? I wonder if we know that processing will always take longer than downloading the next chunk? we may want to allow this buffer to be proportional to the size of the worker pool, maybe?

PR #750, opened 2022-09-05
Keep entries chunk worker goroutines running

gammazero:
Do not start and stop entries chunk worker goroutines with each ad processed. Instead, if enabled, use a global pool that is at least as big as the worker pool size. The worker goroutines compete fairly to use the entries chunk workers, so one will not starve others.

The config setting `Ingest.EntriesChunkConcurrency` controls the number of entries chunk workers and is runtime reloadable. A value of 0 (the default) sets this to the same as the ingest worker pool size. A value of -1 disables it and makes entries chunk processing synchronous.

### DIFF
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -17,7 +17,7 @@ type Ingest struct {
 	// size set by SyncSegmentDepthLimit. AdvertisementDepthLimit sets the
 	// limit on the total number of advertisements across all segments.
 	AdvertisementDepthLimit int
-	// EntriesChunkConcurrency, tells the indexer to process chunks of
+	// EntriesChunkConcurrency tells the indexer to process chunks of
 	// multihash entries asynchronously using this number of goroutines, where
 	// this number is at least as large the worker pool size. If n is -1,
 	// asynchronous processing is disabled. If n < IngestWorkerCount, then n is
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -17,11 +17,11 @@ type Ingest struct {
 	// size set by SyncSegmentDepthLimit. AdvertisementDepthLimit sets the
 	// limit on the total number of advertisements across all segments.
 	AdvertisementDepthLimit int
-	// EntriesChunkConcurrency is the number of additional goroutines for each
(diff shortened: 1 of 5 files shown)


config/ingest.go
```diff
@@ -17,11 +17,11 @@ type Ingest struct {
 	// size set by SyncSegmentDepthLimit. AdvertisementDepthLimit sets the
 	// limit on the total number of advertisements across all segments.
 	AdvertisementDepthLimit int
-	// EntriesChunkConcurrency is the number of additional goroutines for each
-	// publisher/worker to asynchronously process entry chunks. This allows
-	// fetching the next entry chunk without waiting for the current one to
-	// finish being written. A value of 1 means no concurrency, and zero uses
-	// the default. This value is reloadable.
+	// EntriesChunkConcurrency, tells the indexer to process chunks of
```
willscott, line 20:
```suggestion
	// EntriesChunkConcurrency tells the indexer to process chunks of
```

internal/ingest/ingest.go
```diff
@@ -136,11 +136,11 @@ type Ingester struct {
...
-	// chunkConcurrency is accessed through SetEntriesChunkConcurrency() and
-	// entriesChunkConcurrency(). It tells the Ingester to handle writing entry chunks
-	// asynchronously using this main additional goroutines per
-	// publisher/worker.
-	chunkConcurrency int32
+
+	chunkConcurrency int
+	chunkWorkQueue   chan func()
+	// chunkConcurrencyRWMute protects access to chunkConcurrency and chunkWorkQueue.
+	chunkConcurrencyRWMutex sync.RWMutex
```
willscott, line 143: this mutex is only in place to be able to set the concurrency?

willscott (approved):

internal/ingest/ingest.go
willscott: one q: is there any chance that processing of a chunk can generate new work? are there issues with the changing of chunk size holding the write lock while finishing the processing queue?

gammazero: Yes. If it were not runtime configurable, then it would be unnecessary.

PR #761, opened 2022-09-06
Write entries chunks synchronously or asynchronously

gammazero:
There is no performance benefit to having a pool of goroutines over having one asynchronous goroutine per publisher/worker to asynchronously process and entries chunks. Having one additional goroutine provides a simpler implementation that uses less memory. It is also simpler to configure since the choice is to enable or disable the asynchronous behavior.

The default behavior is asynchronous.

Make the configuration setting `Indexer.CorePutConcurrency` run-time reloadable.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -77,7 +77,6 @@ config file at runtime.
   },
   "Ingest": {
     "AdvertisementDepthLimit": 33554432,
-    "EntriesChunkConcurrency": 0,
     "EntriesDepthLimit": 65536,
     "HttpSyncRetryMax": 4,
     "HttpSyncRetryWaitMax": "30s",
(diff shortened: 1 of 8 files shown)


doc/config.md
```diff
@@ -94,7 +93,8 @@ config file at runtime.
     "ResendDirectAnnounce": true,
     "StoreBatchSize": 4096,
     "SyncSegmentDepthLimit": 2000,
-    "SyncTimeout": "2h0m0s"
+    "SyncTimeout": "2h0m0s",
+    "SyncWriteEntries": false
```
masih, line 97: Does this mean entries never get synced?

gammazero: No, it means that we are not forcing it to write synchronously, and allowing the default async behavior.  Better name suggestion? (I would like a default value of `false`)

masih: OK we have the word "Sync" in `SyncTimeout` implying the meaning "fetch", and right below it `SyncWriteEntries` implying synchronous.

Would it make sense to rename either so that it is less confusing? Or consistently use one or the other to mean the same thing.

Issue #766, opened 2022-09-07
Configure resource limit and request for index star


masih:
Set resource limit and request for indexstar deployment in `prod`. 

Prerequisite: [Read docs on pod resource management](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)

masih:
Notes: 
* use the same resources for `limit` and `request` to guarantee that the pod will be scheduled on a node that has capacity to grow up to the limit. 
* start by `2Gi` memory and `0.5` CPU

PR #772, opened 2022-09-07
Set memory and CPU limits for indexstar in PROD

ischasny:
Fixes https://github.com/filecoin-project/storetheindex/issues/766

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/indexstar/deployment.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/indexstar/deployment.yaml
@@ -21,3 +21,10 @@ spec:
             - '--backends=http://xabi-indexer:3000/'
             - '--backends=http://vega-indexer:3000/'
             - '--backends=http://dido-indexer:3000/'
+          resources:
+            limits:
+              cpu: "0.5"
+              memory: 2Gi
+            requests:
+              cpu: "0.5"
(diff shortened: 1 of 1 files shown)


masih:
LGTM; As an exercise I'd like you to build the YAML files from indexstar kustomization. `cd` into the `indexstar` directory which should contain a `kustomization.yaml` and run `kustomize build .`.

ischasny:
> LGTM; As an exercise I'd like you to build the YAML files from indexstar kustomization. `cd` into the `indexstar` directory which should contain a `kustomization.yaml` and run `kustomize build .`.

Done - it builds fine

masih (approved):
Thank you for building this locally. The point of the exercise was to generate questions which it did; discussed via DM. 


Summary:
* `kustomize build .` validates that YAML patching is successful but it does not verify that the schema of the YAML produced is valid to the target K8S API
* To truly validate the YAML, one can apply the generated YAML via `dry-run` 
* There are two modes for `dry-run`: `client` (default) and `server`.
* The former only checks general K8S schema
* The latter fully verifies it against the remote K8S API server and would work for things like CRD (Custom Resource Definitions, i.e. extensions to the core K8S YAML schema/api) that are cluster-specific.
* Example of server-side validation: `kustomize build . | kubectl apply --dry-run=server -f -`
  * For this to work, `kubectl` context should be pointing to the target cluster.

PR #773, opened 2022-09-07
add pod monitor for indexstar

willscott:
Allow tracking metrics emitted by the load splitter


deploy/manifests/base/indexstar/pod-monitor.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: monitoring.coreos.com/v1
```
masih, line 1: Include `pod-monitor.yaml` in `kustomization.yaml` at this directory under `resources`.

masih (commented):
If you'd like to actually roll out the changes in the same PR we also need to update the [image](https://github.com/filecoin-project/storetheindex/blob/main/deploy/manifests/prod/us-east-2/tenant/storetheindex/indexstar/kustomization.yaml#L20) deployed in prod to [the latest built](https://github.com/filecoin-shipyard/indexstar/runs/8233017166?check_suite_focus=true#step:3:12) that exposes the metrics.

masih (approved):

PR #785, opened 2022-09-08
Deploy latest to `dido` and `oden`

masih:
Deploy latest to pebble-backed nodes and use the same ingest settings as vega.


gammazero (approved):
Is there a README update as well?

PR #797, opened 2022-09-09
add metric for number of active workers

willscott:
to know how many active chains are being pulled


masih (approved):

willscott:
are we okay that this breaks on go 1.18?

gammazero:
We need this to build with go1.19, but we probably do not want to require that just yet.

willscott:
switched to the older atomic structure

PR #799, opened 2022-09-09
Fix/indexstar update

willscott:
attempt rolling out updated indexstar with updated scatter/gather. also use it for reframe.


masih:
This PR seems to be redundant considering #800 contains the same changes. 

Merging #800 will do I think

PR #800, opened 2022-09-09
Update vega policy

willscott:
this provider is important for autoretrieve testing, and seems to have fallen off of prod pair


masih (approved):

willscott:
note that this includes #799  - should i merge that first / exclude?

PR #803, opened 2022-09-12
have vega poll nft storage quickly

willscott:


masih (approved):

masih:
Note that `Discovery` config is not auto reloaded. This means for the changes to take effect nodes need to be redeployed. Please go ahead ad redeploy `Vega` since it is not consuming anything. Hold off on `oden` until the ingest is interrpupted since it is currently busy

PR #804, opened 2022-09-12
propose spec for provider families

willscott:
Initial proposal for how we extended advertisement chains for provider families.

* does the signing setup make sense?
* do the semantics around how the family is applied make sense?

### DIFF
--- a/doc/ingest.md
+++ b/doc/ingest.md
@@ -26,6 +26,7 @@ type Advertisement struct {
     ContextID Bytes
     Metadata Bytes
     IsRm Bool
+    ExtendedProvider optional ExtendedMetadata
 }
 ```
 
@@ -39,6 +40,7 @@ type Advertisement struct {
   * If a ContextID is used with different metadata, all previous CIDs advertised under that ContextID will have their metadata updated to the most recent.
   * If a ContextID is used with the `IsRm` flag set, all previous CIDs advertised under that ContextID will be removed.
 * Metadata represents additional opaque data that is returned in client query responses for any of the CIDs in this advertisement. It is expected to start with a `varint` indicating the remaining format of metadata. The opaque data is send to the provider when retrieving content for the provider to use to retrieve the content. Storetheindex operators may limit the length of this field, and it is recommended to keep it below 100 bytes.
+* If ExtendedProvider is specified, indexers which understand the `ExtendedMetadata` extension should ignore the `Provider`, `Addresses` and `Metadata` specified in the advertisement in factor of those specified in the extended metadata. The values in the direct advertisement should still be set to a compatible endpoint for content routers which do note understand full `ExtendedMetadata` semantics.
 
 #### Entries data structure
 
@@ -85,6 +87,29 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
 * http
   * the proposed `uvarint` protocol is `0x3D0000`.
   * the following bytes are not yet defined.
+
+#### ExtendedMetadata
+
+The `ExtendedMetadata` field allows for specification of provider families, in cases where a provider operates multiple PeerIDs, perhaps with different access methods between them, but over the same database of content.
+
+```
+type ExtendedMetadata struct {
+    Family [ExtendedProvider]
+}
+
+type ExtendedProvider struct {
+    Provider String
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If Metadata is not specified for an `ExtendedProvider`, the metadata for an individual `Advertisement` will be used instead.
+* If an `ExtendedProvider` listing is written with no `ContextID`, those peers will be returned forall published advertisements for the advertisement Publisher.
+* If an `ExtendedProvider` listing is written on an advertisement with a `ContextID`, it will over-ride any chain-globals, and that specific set of records will be returned for that context ID.
+* The `Signature` for an `ExtendedProvider` is made using the key of that `Provider` over the serialized `Advertisement` containing the following subset of fields:
+    * `PreviousID`, `Provider`, `Addresses`, `Entries`, `ContextID`, `Metadata`.
 
 ### Advertisement transfer
 


doc/ingest.md
```diff
@@ -86,6 +88,29 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If Metadata is not specified for an `ExtendedProvider`, the metadata for an individual `Advertisement` will be used instead.
+* If an `ExtendedProvider` listing is written with no `ContextID`, those peers will be returned forall published advertisements for the advertisement Publisher.
+* If an `ExtendedProvider` listing is written on an advertisement with a `ContextID`, it will over-ride any chain-globals, and that specific set of records will be returned for that context ID.
+* The `Signature` for an `ExtendedProvider` is made using the key of that `Provider` over the serialized `Advertisement` containing the following subset of fields:
```
masih, line 111: Would this leave `ExtendedProvider` open to ad-hoc modification?

doc/ingest.md
```diff
@@ -26,6 +26,7 @@ type Advertisement struct {
     ContextID Bytes
     Metadata Bytes
     IsRm Bool
+    ExtendedProvider optional ExtendedMetadata
```
masih, line 29: Why nesting of `ExtendedMetadata` struct type? Why not `ExtendedProviders optional [ExtendedProvider]`? is there an explicit requirement for "family" grouping?

doc/ingest.md
```diff
@@ -86,6 +88,29 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If Metadata is not specified for an `ExtendedProvider`, the metadata for an individual `Advertisement` will be used instead.
+* If an `ExtendedProvider` listing is written with no `ContextID`, those peers will be returned forall published advertisements for the advertisement Publisher.
+* If an `ExtendedProvider` listing is written on an advertisement with a `ContextID`, it will over-ride any chain-globals, and that specific set of records will be returned for that context ID.
+* The `Signature` for an `ExtendedProvider` is made using the key of that `Provider` over the serialized `Advertisement` containing the following subset of fields:
+    * `PreviousID`, `Provider`, `Addresses`, `Entries`, `ContextID`, `Metadata`.
```
ischasny, line 112: Do we need `Addresses` and `Entries` to be signed over given that those are ignored if `ExtendedMetadata` is present?

willscott: updated signature to be safer

willscott:
Expanded on this a bit - ready for a next round of eyes

doc/ingest.md
```diff
@@ -39,6 +40,7 @@ type Advertisement struct {
   * If a ContextID is used with different metadata, all previous CIDs advertised under that ContextID will have their metadata updated to the most recent.
   * If a ContextID is used with the `IsRm` flag set, all previous CIDs advertised under that ContextID will be removed.
 * Metadata represents additional opaque data that is returned in client query responses for any of the CIDs in this advertisement. It is expected to start with a `varint` indicating the remaining format of metadata. The opaque data is send to the provider when retrieving content for the provider to use to retrieve the content. Storetheindex operators may limit the length of this field, and it is recommended to keep it below 100 bytes.
+* If ExtendedProvider is specified, indexers which understand the `ExtendedMetadata` extension should ignore the `Provider`, `Addresses` and `Metadata` specified in the advertisement in factor of those specified in the extended metadata. The values in the direct advertisement should still be set to a compatible endpoint for content routers which do note understand full `ExtendedMetadata` semantics.
```
masih, line 43:
```suggestion
* If ExtendedProvider is specified, indexers which understand the `ExtendedMetadata` extension should ignore the `Provider`, `Addresses` and `Metadata` specified in the advertisement in factor of those specified in the extended metadata. The values in the direct advertisement should still be set to a compatible endpoint for content routers which do not understand full `ExtendedMetadata` semantics.
```

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+
+type ExtendedProvider struct {
+    Provider String
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If Metadata is not specified for an `ExtendedProvider`, the metadata for an individual `Advertisement` will be used instead.
```
masih, line 109:
```suggestion
* If `Metadata` is not specified for an `ExtendedProvider`, the metadata from the encapsulating  `Advertisement` will be used instead.
```

Sanity check: Is this what the sentence means?

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+* If an `ExtendedProvider` listing is written with no `ContextID`, those peers will be returned for all published advertisements for the advertisement Publisher.
+  * If `Override` is set on an `ExtendedProvider` entry on an advertisement with a `ContextID`, that indicates that any set chain-level set of providers should not be returned for that context ID.
```
masih, line 111:
```suggestion
  * If `Override` is set on an `ExtendedProvider` entry on an advertisement with a `ContextID`, it indicates that any specified chain-level set of providers should not be returned for that context ID.
```

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+  * If `Override` is not set on an entry for an advertisement with a `ContextID`, it will extended any set chain-level `ExtndedProviders`.
+* The `Signature` for an `ExtendedProvider` is made using the key of that extended `Provider`.
```
masih, line 113:
```suggestion
* The `Signature` for an `ExtendedProvider` is signed by the key of that extended `Provider`.
```

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+  * The full advertisement object is serialized, with all instances of `Signature` replaces with an empty array of bytes.
+  * This serialization is then hashed, and the hash is then signed.
```
masih, line 115: Would this break the verification for indexer nodes that don't support `Extended*` features (yet)?

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
   * the proposed `uvarint` protocol is `0x3D0000`.
   * the following bytes are not yet defined.
 
+#### ExtendedMetadata
+
+The `ExtendedMetadata` field allows for specification of provider families, in cases where a provider operates multiple PeerIDs, perhaps with different access methods between them, but over the same database of content.
+
+```
+type ExtendedMetadata struct {
+    Family [ExtendedProvider]
```
masih, line 97: Is there an existing term we could use to avoid introducing "Family" as a new word? `Providers` maybe?

doc/ingest.md
masih, line 29: I wonder how well the field name `ExtendedProvider` with type name `ExtendedMetadata` fit together.

Why the word `Metadata` in the type name?

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If Metadata is not specified for an `ExtendedProvider`, the metadata for an individual `Advertisement` will be used instead.
+* If an `ExtendedProvider` listing is written with no `ContextID`, those peers will be returned for all published advertisements for the advertisement Publisher.
+  * If `Override` is set on an `ExtendedProvider` entry on an advertisement with a `ContextID`, that indicates that any set chain-level set of providers should not be returned for that context ID.
+  * If `Override` is not set on an entry for an advertisement with a `ContextID`, it will extended any set chain-level `ExtndedProviders`.
```
ischasny, line 112: What is meant by "extended" here? Is that about ignoring provider data from the advertisement itself?

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+type ExtendedProvider struct {
+    Provider String
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If Metadata is not specified for an `ExtendedProvider`, the metadata for an individual `Advertisement` will be used instead.
+* If an `ExtendedProvider` listing is written with no `ContextID`, those peers will be returned for all published advertisements for the advertisement Publisher.
```
ischasny, line 110: What would be behaviour of `Override` flag with no context id?

doc/ingest.md
```diff
@@ -86,6 +88,32 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+  * If `Override` is set on an `ExtendedProvider` entry on an advertisement with a `ContextID`, that indicates that any set chain-level set of providers should not be returned for that context ID.
+  * If `Override` is not set on an entry for an advertisement with a `ContextID`, it will extended any set chain-level `ExtndedProviders`.
+* The `Signature` for an `ExtendedProvider` is made using the key of that extended `Provider`.
+  * The full advertisement object is serialized, with all instances of `Signature` replaces with an empty array of bytes.
```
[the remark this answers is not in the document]
willscott: It would - but with the signature sections in it removed - that does force choosing the full set of identities in advance of the signing, but prevents subsequent partial sub-identity addition/removals

doc/ingest.md
willscott: it does encompass the `Metadata` field for the additional provider identities, right?

doc/ingest.md
willscott: I think it does - do we think that's okay?

doc/ingest.md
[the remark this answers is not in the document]
ischasny: Do we not want to allow "partial sub-identity addition/removals" given that in the estuary usecase this new identities can come and go dynamically?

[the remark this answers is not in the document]
willscott: by this do you mean just specifying a delta (eg "the identity d is no longer providing") as opposed to writing out the current family each time there is a delta as would be specified currently (eg "the current family is now a,b, and c" no longer including d)

doc/ingest.md
masih: True while encompassing other things like the addrs and peer ID :)

Just now we use the term `Metadata` to refer to a set of bytes that describe retrieval protocol; it has a very specific meaning. But `ExtendedMetadata` is not just `Metadata`; it also includes addresses and peer IDs.

The name `ExtendedProvider` makes sense to me. Because, so far we have used  the term `Provider` to describe Peer ID, addrs and metadata combination. 

Thinking out loud would it make sense to use the same name for both the field and they type? if not it would be nice to avoid terminology overlap.

This is by no means a blocker I just point it out because it'd be hard to change later and might come across as confusing to the reader of docs/schema

doc/ingest.md
masih: It is certainly undesirable and would make the sentences like "indexers which understand the `ExtendedMetadata` extension should ignore the `Provider`" a bit redundant, since indexers that do not understand it will cease to ingest any further ads from the point at which the extensions are advertised and would have to upgrade.

Can we avoid this? It would be nice if nodes that have not upgraded continue to be able to ingest while nodes that have, get the extra features.

We could:
* keep the ad signature calculation the same.
* Then include the ad signature in the `ExtendedProvider` signature calculation.

That means we can keep old nodes working, and avoid attacks like stitching `ExtendedProvider` from one ad onto another ad.

Do you think that would work?

doc/ingest.md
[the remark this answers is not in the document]
ischasny: It'd be good to allow publishers to add / remove extended providers without having to re-obtain a signature from each of them for each change. I.e. given that a provider has confirmed (signed an AD) that they can provide a certain piece of data, they don't necessarily need to confirm it again once another party can provide the same piece of data too. 

So maybe extended providers could sign only over the AD content, without `ExtendedMetadata` included?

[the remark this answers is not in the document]
willscott: the extended provider needs to sign over it's own multi-addrs to make sure that it isn't being mis-represented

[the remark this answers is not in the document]
ischasny: Maybe a provider could sign the AD without `ExtendedMetadata` but with *their* details filled in as if they were a primary provider? Or signing with `ExtendedMetadata` included only for *that* provider could work too. Wdyt?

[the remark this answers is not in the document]
willscott: see if you think the current setup is reasonable

doc/ingest.md
willscott: see if you think the current setup is reasonable

doc/ingest.md
willscott: renamed

doc/ingest.md
```diff
@@ -86,6 +88,34 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+* If a `Provider` listing is written with no `ContextID`, those peers will be returned for all published advertisements for the advertisement Publisher.
+  * If `Override` is set on an `ExtendedProvider` entry on an advertisement with a `ContextID`, it indicates that any specified chain-level set of providers should not be returned for that context ID.
+  * If `Override` is not set on an entry for an advertisement with a `ContextID`, it will be combined as a union with any set chain-level `ExtndedProviders`.
+  * If `Override` is set on `ExtendedProvider` for an advertisement without a `ContextID`, the entry can be considered invalid and should be ignored.
+* The `Signature` for an `ExtendedProvider` is signed by the key of that extended `Provider`.
+  * The full advertisement object is serialized, with all instances of `Signature` replaces with an empty array of bytes.
+  * This serialization is then hashed, and the hash is then signed.
+  * The provider from the surrounding advertisement must be present as a `Provider` in the `ExtendedProvider` object, and must sign in this way as well. It does not need to provide multiaddrs or transports, but must be present.
```
masih, line 117:
```suggestion
  * The `Provider` from the encapsulating advertisement must be present in the `Providers` of the `ExtendedProvider` object, and must sign in this way as well. It may omit `Metadata` and `Addresses` if they match the values already set at the encapsulating advertisement. However, `Signature` must be present.
```

We used "encapsulating" earlier. Feel free to change all to "surrounding" if that reads better. Consistent terminology is the nice thing to have.

masih (approved):
LGTM 👍 

Left a bunch of comments on wording.
I _think_ we are on the same page about `Signature` backward compatibility. I left a suggestion to harden extended provider association with ads by including ad-level signature when calculating signature of extended providers.

Excited to get the ball rolling on the implementation of this 🚀

ischasny (approved):
Thanks for making these changes. LGTM now

doc/ingest.md
```diff
@@ -86,6 +88,34 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+The `ExtendedProvider` field allows for specification of provider families, in cases where a provider operates multiple PeerIDs, perhaps with different access methods between them, but over the same database of content.
+
+```
+type ExtendedProvider struct {
+    Providers [Provider]
+    Override bool
+}
+
+type Provider struct {
+    Provider String
```
willscott, line 102:
```suggestion
    ID String
```

doc/ingest.md
willscott: yeah, that was my intention. i'll add a sentence saying that more clearly

doc/ingest.md
```diff
@@ -86,6 +88,35 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+    ID String
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If `Metadata` is not specified for a `Provider`, the metadata from the encapsulating  `Advertisement` will be used instead.
+* If a `Provider` listing is written with no `ContextID`, those peers will be returned for all advertisements published by the publisher.
+  * If `Override` is set on an `ExtendedProvider` entry on an advertisement with a `ContextID`, it indicates that any specified chain-level set of providers should not be returned for that context ID. `Providers` will be returned Instead.
```
gammazero, line 111: This seems like it overlaps in functionality with a removal (`IsRm`) advertisement for provider and context ID.  If that is the case, then maybe it should be more explicit (set the `IsRm` flag) to specify that a context ID is being deleted for a set of providers.  Or, is there a different use case for this?

In any case, the semantics of `ExtendedProvider` in a removal advertisement should be specified.

doc/ingest.md
```diff
@@ -86,6 +88,35 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+  * If `Override` is not set on an entry for an advertisement with a `ContextID`, it will be combined as a union with any set chain-level `ExtndedProvider`s.
+  * If `Override` is set on `ExtendedProvider` for an advertisement without a `ContextID`, the entry is invalid and should be ignored.
```
gammazero, line 113: What about the case of a removal advertisement (`IsRm == true`) that does not have a context ID?  Would that be valid and indicate that a set of providers is being removed, or should that be done with separate removal advertisements?

doc/ingest.md
```diff
@@ -86,6 +88,35 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
+```
+
+* If `Metadata` is not specified for a `Provider`, the metadata from the encapsulating  `Advertisement` will be used instead.
+* If a `Provider` listing is written with no `ContextID`, those peers will be returned for all advertisements published by the publisher.
+  * If `Override` is set on an `ExtendedProvider` entry on an advertisement with a `ContextID`, it indicates that any specified chain-level set of providers should not be returned for that context ID. `Providers` will be returned Instead.
+  * If `Override` is not set on an entry for an advertisement with a `ContextID`, it will be combined as a union with any set chain-level `ExtndedProvider`s.
```
gammazero, line 112: > union with any set chain-level `ExtndedProvider`s.

Should this say, "union with any set chain-level provider information (Addresses, Context ID, Metadata)"?

doc/ingest.md
```diff
@@ -86,6 +88,35 @@ The network indexer nodes expect that metadata begins with a `uvarint` identifyi
...
+    Providers [Provider]
+    Override bool
+}
+
+type Provider struct {
+    ID String
+    Addresses [String]
+    Metadata optional Bytes
+    Signature Bytes
+}
```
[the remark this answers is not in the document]
willscott: context id will be from the advertisement containing this new provider record.

doc/ingest.md
willscott: added line indicating `ExtendedProviders` should be ignored in `IsRm` advertisements.

gammazero (approved):

PR #816, opened 2022-09-15
Update to use storethehash with multi-file primary with GC

gammazero:
Newer storethehash splits primary storage across multiple files, and has garbage collections for the primary storage.

Conversion from the old to the new primary is done automatically when the new indexer starts up.  This process will take significant time as the entire index needs to be updated with the new locations of primary data.  Progress is logged.


masih:
**Do not Merge: being tested in dev first**

.github/workflows/ecr-publisher.yml
```diff
@@ -14,6 +14,7 @@ on:
       - 'doc/**'
     branches:
       - main
+      - primary-gc # FIXME: remove prior to PR merge; added to build containers for testing on dev.
```
willscott, line 17: this still needs to get removed, yeah?

gammazero: Nice catch!! Thanks, done.

gammazero:
I am going to merge this PR as it has become clear that it is necessary if we are going to use STH.

PR #843, opened 2022-09-23
Set peering for oden; no custom auto-sync

gammazero:
Use peering to get republished HTTP announce messages, instead of auto-syncing every 5 minutes.

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/oden/config.json
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/oden/config.json
@@ -50,12 +50,7 @@
     "PollInterval": "24h0m0s",
     "PollRetryAfter": "5h0m0s",
     "PollStopAfter": "168h0m0s",
-    "PollOverrides": [{
-      "ProviderID": "QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6ZmC",
-      "Interval": "5m0s",
-      "RetryAfter": "3m0s",
-      "StopAfter": "168h0m0s"
-    }],
+    "PollOverrides": null,
     "RediscoverWait": "5m0s",
     "Timeout": "2m0s"
   },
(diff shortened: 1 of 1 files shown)


masih (commented):
The purpose of this config is to stay on top of nft.storage. Same is configured for vega.

gammazero:
If indexer-0 or index-1 get the HTTP announce, then that is relayed to peers and the auto-sync should not be needed.  If those indexers do not get HTTP announce from NFT storage, then we can keep this previous config.  With the old config the autosync will continue to trigger while syncing long ad chains, or anything that takes over 5 minutes.  That will build up goroutines waiting to perform a sync.

If we intend to use auto-sync like this, then there should be some synchronization to check if there is already an auto-sync in progress.

masih:
Ok let's try it

gammazero:
I figure it is OK since it is what the dido node is already doing.

willscott (approved):
should monitor when merged/applied to see if peering is sufficient - otherwise should add back in the poll over-ride.


What from the peer to the other indexers will cause forwarding of the announcements currently?

willscott:
> if indexer-0 or index-1 get the HTTP announce, then that is relayed to peers

i didn't realize we had the conversion from http announce to gossip already in legs. that's great

gammazero:
> What from the peer to the other indexers will cause forwarding of the announcements currently?

[This configuration](https://github.com/filecoin-project/storetheindex/blob/main/deploy/manifests/prod/us-east-2/tenant/storetheindex/indexer-config.yaml#L85) tells the indexer-0 and indexer-1 to forward HTTP announce messages to any peers, either peers in their config or others that are peers to the them.

masih:
@gammazero note there are no announcements flowing on `oden` after the changes here. Can I ask you to take a look and see why announcements are not being forwarded to oden?

gammazero:
Primarily because of this:https://github.com/filecoin-project/storetheindex/commit/2a038a4030d74fc381e65d36255a4e1e6b262673

Also, indexer-1 is currently down for upgrade and indexer-0 has an inoperable valuestore until it finishes upgrade.

masih:
I see, thank you. I'll restore this setting perhaps with an interval bigger than 5 minutes tomorrow while indexer 0 and 1 are upgrading. We can revisit when they are back up.

PR #854, opened 2022-09-27
Increase worker count on `dido`

masih:
Investigate high queue times on `dido` by increasing worker count.

While at it, set the rate limit to zero to make it clear that it is disabled.


willscott:
retroactive 'looks good' - did increasing workers here help?

masih:
[No](https://protocollabs.grafana.net/d/pPROWOD7z/providers?orgId=1&var-Source=storetheindex-prod&var-Jobs=storetheindex%2Findexer-single&var-pod=dido-indexer&from=1664272102085&to=1664282730133) unfortunately. There was still high queue time but it's not straightforward to conclude facts from metrics since the queue percent depends on the load on node, sync latency among other things

PR #855, opened 2022-09-27
Record time to store multihashes

### LINKED ISSUE #744: Add ingest time per multihash metric
The time spent ingesting a batch of multihashes is logged by the ingester. Ingester also logs the size of the batch in a different line. To find out the normalised ingest time one has to do Loki query gymnastics like this:
```
sum(
sum_over_time ({environment="prod", namespace="storetheindex", pod_name=~"(dido|vega).+"} |~ "elapsed" | json | unwrap elapsed [1m])
) by (pod_name)

/

sum(
sum_over_time ({environment="prod", namespace="storetheindex", pod_name=~"(dido|vega).+"} |~ "Put multihashes in" | json | unwrap count [1m])
) by (pod_name)
```

Add a metric that reports the normalised ingest time per multihash for a processed batch.

gammazero:
Report metric for average nanoseconds to store a multihash. This is the average over all the multihashes stored for the last ad processed.

Also, log a message that says how long the ad took to ingest and the count of multihashes that were ingested for that ad. Example:
```
2022-09-27T08:59:15.364-0700    INFO    indexer/ingest  ingest/linksystem.go:152        Finished syncing advertisement  {"publisher": "12D3KooWKRJVaDiWcwAvvPNbzWPH8D49qzV5ExxNAGpsdRsABLmr", "adCid": "baguqeeraeigva2nffpv4xymg7x7sssfklenb4kuoz44q4ls5cn6nt66rjlhq", "contextID": "AXESIEHjf2X4P3ruFlU8HpSIR2rH3/qx2x9nALdjIhyI3oOL", "provider": "12D3KooWKRJVaDiWcwAvvPNbzWPH8D49qzV5ExxNAGpsdRsABLmr", "entriesKind": "EntryChunk", "elapsed": "2.31674075s", "multihashes": 25481}
```

- Cleanup some log messages
- Log elapsed time when finished ingesting advertisement

Fixes #744

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"strings"
 	"time"
 
@@ -159,8 +160,8 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 		stats.Record(context.Background(), metrics.EntriesSyncLatency.M(elapsedMsec))
 
 		// Record average time to store 10000 multihashes.
-		elapsedPer10KMh := float64(entsStoreElapsed.Nanoseconds()/int64(mhCount)) / (1e6 / 1e4)
-		stats.Record(context.Background(), metrics.MhStore10KLatency.M(elapsedPer10KMh))
+		elapsedPerMh := int64(math.Round(float64(entsStoreElapsed.Nanoseconds()) / float64(mhCount)))
+		stats.Record(context.Background(), metrics.MhStoreNanoseconds.M(elapsedPerMh))
 	}()
 
 	// Get provider ID from advertisement.
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -159,7 +159,8 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 		elapsedMsec = float64(elapsed.Nanoseconds()) / 1e6
 		stats.Record(context.Background(), metrics.EntriesSyncLatency.M(elapsedMsec))
 
-		// Record average time to store 10000 multihashes.
+		// Record average time to store one multihash, for all multihahses in
+		// this ad's entries.
 		elapsedPerMh := int64(math.Round(float64(entsStoreElapsed.Nanoseconds()) / float64(mhCount)))
 		stats.Record(context.Background(), metrics.MhStoreNanoseconds.M(elapsedPerMh))
 	}()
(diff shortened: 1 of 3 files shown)


internal/ingest/linksystem.go
```diff
@@ -135,13 +134,35 @@ func verifyAdvertisement(n ipld.Node, reg *registry.Registry) (peer.ID, error) {
...
+		if mhCount == 0 {
+			return
+		}
+		// Record how long entries sync took.
+		elapsed = now.Sub(entsSyncStart)
+		elapsedMsec = float64(elapsed.Nanoseconds()) / 1e6
+		stats.Record(context.Background(), metrics.EntriesSyncLatency.M(elapsedMsec))
+
+		// Record average time to store 10000 multihashes.
+		elapsedPer10KMh := float64(entsStoreElapsed.Nanoseconds()/int64(mhCount)) / 100
```
willscott, line 162: the `/100` is incorporating both the `ns-ms` conversion and the `10k` batching, right? maybe break those two factors out for clarity?

willscott (approved):

masih:
Can we avoid all of this and report the latency as nanoseconds without unit as a histogram?

Is there something that forces us to specify unit for the metric here?

gammazero:
I can change that to a unitless value, and use nanoseconds or microseconds. The histogram was used here because that is how all the other latency metrics are captured. Do we want to change the other latency metrics similarly?

gammazero:
@masih OK, how about now?

masih (approved):
I'm sorry there must be something I am missing: Why change ad ingest latency and entries sync latency to unitless? 

Using _a_ unitless metric to would be plenty good to satisfy #744 no? 

No blockers; thank you for picking this up

gammazero:
@masih 
> Why change ad ingest latency and entries sync latency to unitless?

Only to make them reported exactly like the multihash store time metric.

... Maybe I am not getting what #744 is asking for, [a new metric](https://github.com/filecoin-project/storetheindex/pull/855/files#diff-42c7bc9f1f4c699aabdad03f6522878dd826967a4d33545502a6d0a2a88cb927R37) that records only the average time to _store_ a multihash or [additional data to be logged](https://github.com/filecoin-project/storetheindex/pull/855/files#diff-9224e87eb2202ee2db2e74a111cf0cb420d9a05619897c77d8dadd57cbaa3af3R152) that tells how long an ad took to ingest and how many multihashes were associated with the add.

In this PR I did both (see links above). If the `MhStoreNanoseconds` metric is not needed then it should be removed.

masih:
Here is how it occurs for me: #744 asks for `MhStoreNanoseconds` which this PR delivers. The reason for that metric being unitless as far as I understand is that average storage time per multihash may be lower than the smallest time unit allowed by  the metric. As for the other additional metrics added in this PR, fantastic; thank you for adding those. I don't understand why they need to be limitless other than consistency? The nanosecond metric is the exception here, I am not sure why we avoid unit for the other ones. 

I have no hard feeling about either approach; only trying to discover what value does consistency in unitless reporting of time would bring when we know `MhStoreNanoseconds` is an exceptional case. I for one am fine with only reporting that metric as unitless instead of changing everything because of it.

Issue #786, opened 2022-09-08
Investigate nginx ingress restarts 


masih:
After giving nginx more resource it seems to restart under load. Understand why.

PR #856, opened 2022-09-28
Enable prometheus metric collection from nginx ingress

masih:
Export prometheus monitoring port on ngnix pods and write metrics out to PL grafana.

Relates to #786


ischasny (approved):
LGTM

PR #860, opened 2022-09-28
Increase CPU resource for nginx ingress in prod

masih:
The nginx pods were evicted earlier this morning because they went over the configured limit for CPU usage.

Increase CPU limit by 4X.

See [metrics](0).

[0]: https://protocollabs.grafana.net/d/pPROWOD7z/providers?orgId=1&var-Source=storetheindex-prod&var-Jobs=storetheindex%2Findexer-single&var-pod=dido-indexer&var-pod=oden-indexer&from=1664347369311&to=1664350773579&viewPanel=87


ischasny (approved):
LGTM

PR #861, opened 2022-09-28
Validate advertisement

gammazero:
## Context
This checks that the metadata and context ID is within the allowed limit, and if not causes the ad sync to fail. This was not done previously and there is evidence that some oversized data was ingested on some indexers.

The sync is ended with an error, instead of skipping the advertisement, to limit the number of possibly huge unusable ads that are downloaded for ingestion.

## Proposed Changes
Verification of the advertisement metadata length and context ID length is added to the advertisement verification that is done in the linksystem `StorageWriteOpener`

## Tests
A new test checks that an advertisement with too large of metadata does not get ingested.

## Revert Strategy
Change is safe to revert.

### DIFF
--- a/internal/ingest/ingest_test.go
+++ b/internal/ingest/ingest_test.go
@@ -671,7 +671,7 @@ func TestSync(t *testing.T) {
 	defer pub.Close()
 	connectHosts(t, h, pubHost)
 
-	c1, mhs, providerID := publishRandomIndexAndAdv(t, pub, lsys, false)
+	c1, mhs, providerID := publishRandomIndexAndAdv(t, pub, lsys, false, nil)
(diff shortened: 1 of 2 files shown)


masih:
> evidence that some oversized data was ingested on some indexers

Is there a known provider ID / ad CID?

masih (approved):

gammazero:
> Is there a known provider ID / ad CID?

Unfortunately no.

The file `storethehash.data.2509` on prod indexer-1 is one of the files that is signaifcantly larger than it should be. It would be necessary to walk the sth index records to find indexes into that primary file, and then read those primary records to see which one(s) are oversized.

Issue #863, opened 2022-09-29
Uncached find requests time out


honghaoq:
Find requests from Hydra for which no cached entry is present, time out. Note the timeout set by Hydra client is 1 second. 
Understand how to reduce origin latency in indexstar.

masih:
Relates to https://github.com/filecoin-shipyard/indexstar/issues/12 

indexer nodes that are unavailable should be excluded from the backend. Otherwise every request hitting indexstar will pay a n internal timeout cost which reslts in higher origin latency as observed in ColudFront caching layer and ultimately a timeout on hydra side.

masih:
Changes in #864 significantly reduced the timeouts and origin response latency visible on both indexer and hydra metrics:

<img width="759" alt="image" src="https://user-images.githubusercontent.com/301855/193024040-daa0f622-3906-4536-8963-e3e14c002e12.png">
<img width="572" alt="image" src="https://user-images.githubusercontent.com/301855/193024074-631d7e83-0ab4-43c9-8d3a-da9262ecd7cc.png">
<img width="616" alt="image" src="https://user-images.githubusercontent.com/301855/193024101-9dc9aa0e-62e1-4fd9-ab61-46aac3df498d.png">


Question @honghaoq: What is the target QoS we are aiming for here? Is the reduction above good enough? if not please quantify.

gammazero:
This has been fixed by internal routing improvements. Including high-performance provider information cache in indexstar and in dhfind/dhstore, and much more.

PR #864, opened 2022-09-29
Exclude `rami` from indexstar and adjust resources

masih:
`rami` is currently going through an index file upgrade; exclude it from indexstar to avoid expensive timeouts until indexstar implements proper exclusion of unhealthy backends.

The indexstar CPU usage was hovering at 70%; increase the CPU resource to give it enough headroom at the face of request fluctuations. As a result, up the max worker count in relevant EKS nodegroup to allow scheduling of pods with higher CPU limit.

Reduce instance counts to 3 since most of the pods were mostly idle now that resource limit is elevated.

Enable compression for `/reframe` endpoint on CloudFront; unrelated to the timeout investigation. Added here since it was changed in the past and terraform update was missing.

Relates to #863

### DIFF
--- a/deploy/infrastructure/prod/us-east-2/cloudfront.tf
+++ b/deploy/infrastructure/prod/us-east-2/cloudfront.tf
@@ -73,7 +73,7 @@ resource "aws_cloudfront_distribution" "cdn" {
   }
 
   ordered_cache_behavior {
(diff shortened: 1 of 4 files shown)


ischasny (approved):
Lgtm

PR #868, opened 2022-09-30
Restore small poll interval in `oden`

masih:
There seem to be no announcements flowing while indexer-0 and indexer-1 are performing primary upgrade. Reduce poll interval in order to avoid `oden` falling behind due to lack of announcements.

Revisit when those nodes are back up.

Relates to:
 - https://github.com/filecoin-project/storetheindex/pull/843#issuecomment-1262581983


ischasny (approved):
Lgtm

PR #869, opened 2022-09-30
Handle IPLD `SkipMe` errors as content not found during ad ingest

### LINKED ISSUE #866: Graphsync skip error disrupts ingest
The graphsync skip error causes the ingest to stop from a node with error: `graphsync request failed to complete: skip`.

Understand if this error is "skippable" and if so skip the error and proceed with ingest.

masih:
IPLD `SkipMe` errors returned during traversal via GraphSync signal partial availability of blocks. We already skip over "content not found" errors that occur while fetching entries associated to an ad. Therefore, it makes sense to do the same on `SkipMe` errors. To summarise, this means their ingest of ads that their entries are not traversable will not be processed. Otherwise, ad ingest for a provider may halt indefinitely.

Fixes #866

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -274,10 +274,15 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 		//
 		// Use string search until then.
 		wrappedErr := fmt.Errorf("failed to sync first entry while checking entries type: %w", err)
-		if strings.Contains(err.Error(), "content not found") {
+		msg := err.Error()
+		switch {
+		case
+			strings.Contains(msg, "content not found"),
+			strings.Contains(msg, "graphsync request failed to complete: skip"):
 			return adIngestError{adIngestContentNotFound, wrappedErr}
+		default:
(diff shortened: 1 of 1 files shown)


internal/ingest/linksystem.go
```diff
@@ -274,10 +274,15 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 		//
 		// Use string search until then.
 		wrappedErr := fmt.Errorf("failed to sync first entry while checking entries type: %w", err)
-		if strings.Contains(err.Error(), "content not found") {
+		msg := err.Error()
+		switch {
+		case
+			strings.Contains(msg, "content not found"),
```
ischasny, line 280: I wonder if there is a better way to distinguish between graph sync errors, so that it won't break once an error message changes?

ischasny (approved):
Approving, but please see the comment inline.

internal/ingest/linksystem.go
masih: if the error is not wrapped that's the best we can do. See comment above. Having said that I am not sure if `SkipMe` is wrapped or not.

gammazero (approved):

Issue #865, opened 2022-09-29
Manual sync from previously unknown peers fail with `no good addresses`


masih:
When a provider over datatransfer is completely unknown to the indexer, manual sync seems to fail with `no good addresses` error. This is because the datatransfer libp2p host peerstore is not updated with the address from the explicit sync and therefore the host fails to connect.

Update the libp2p peerstore with the address provided via the sync command before manual sync.

gammazero:
I have not observed this and can do a manual sync with a previously unknown publisher.  Check that you are not using an unroutable local network address, or set `Discovery.FilterIPs` to `false`.

masih:
I observed this on all nodes except indexer 0 and 1 on prod. I confirmed that peerstore is an issue by adding the peer multiaddr to the bootstrap list. After that the sync started working.

This should be straightforward to reproduce.

masih:
OK this is happening again; here is a concrete address example:

```
{"level":"error","ts":"2022-09-30T12:23:10.947Z","logger":"indexer/ingest","caller":"ingest/ingest.go:415","msg":"Failed to sync with provider","provider":"12D3KooWDtiA9w5c37MnFpGWr2M12m8neoUqDroHEMsJVuf3ELi7","peerAddr":"/ip4/195.167.147.43/tcp/9999/p2p/12D3KooWDtiA9w5c37MnFpGWr2M12m8neoUqDroHEMsJVuf3ELi7","depth":0,"resync":false,"err":"cannot query head for sync: Get \"http://unused.invalid/head\": no good addresses. Possibly incorrect topic configured"}
```

I am going to pick this up next.

masih:
After having spoken with Andrew, reducing priority of this to P3 because nodes seem to recover from it after a while and I can't reproduce this locally.

[...]

PR #872, opened 2022-09-30
Reproduce #865 using a unit test

masih:
Write a test that asserts explicitly syncing from an unknown provider updates the registry.


masih:
Can't reproduce locally.

PR #877, opened 2022-10-03
Disable indexer core cache on `oden` to avoid panic

masih:
A panic occurs when comparing values returned by cache. Disable it on `oden` to avoid panic. Aside from that the cache is hardly used by the node.

See:
 - https://github.com/filecoin-project/go-indexer-core/issues/80


ischasny (approved):
LGTM

Issue #883, opened 2022-10-04
Panic during ingestion


gammazero:
A panic happened during ingestion, [log here](https://protocollabs.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Benvironment%3D%5C%22prod%5C%22,%20namespace%3D%5C%22storetheindex%5C%22,%20pod_name%3D%5C%22indexer-1%5C%22%7D%20%7C%3D%20%5C%22panic%5C%22%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221664914716271%22,%22to%22:%221664914789729%22%7D%7D). There was not any error logged nearby that is clearly associated.

```
panic: send on closed channel                                                                                                                  
                                                                                                                                               
goroutine 88 [running]:                                                                                                                        
github.com/filecoin-project/storetheindex/internal/ingest.(*Ingester).ingestAd.func6({0x1645dd3?, 0x41e963?}, {{0xc0b5cb3c50?, 0xc00efcd0e0?}}\
, {0x1aac2d0, 0xc00b3ad7a0})                                                                                                                   
        /home/indexer/storetheindex/internal/ingest/linksystem.go:461 +0x333                                                                   
github.com/filecoin-project/go-legs.(*handler).handle.func1({0xc00b4155f0?, 0xc00b583ce0?}, {{0xc0b5cb3c50?, 0x26?}})                          
        /home/indexer/go/pkg/mod/github.com/filecoin-project/go-legs@v0.4.15/subscriber.go:956 +0x116                                          
github.com/filecoin-project/go-legs.wrapBlockHook.func1({0xc00b4155f0, 0x26}, {{0xc0b5cb3c50?, 0x2763c00?}})                                   
        /home/indexer/go/pkg/mod/github.com/filecoin-project/go-legs@v0.4.15/subscriber.go:180 +0xdd                                           
github.com/filecoin-project/go-legs/dtsync.addIncomingBlockHook.func1({0xc00b4155f0, 0x26}, {0x1ab6340, 0xc087392000}, {0x1ab4cf0?, 0xc0b0eb6b\
10?}, {0x1ab1208, 0xc0b0eb6b40})                                                                                                               
[...]

PR #884, opened 2022-10-05
Write entries synchronously to avoid #883

gammazero:
#883 happened on indexer-1, so setting config to write entries synchronously until that is fixex.


willscott:
how much of a perf impact does this have?

PR #905, opened 2022-10-07
Test zero-copy marshal fixes high CPU on `dido`

masih:
Deploy the changes from go-indexer-core that use zero-copy marshaller in pebble to test it fixes the high CPU on `dido`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/cockroachdb/pebble v0.0.0-20220726144858-a78491c0086f
 	github.com/filecoin-project/go-address v0.0.5
 	github.com/filecoin-project/go-dagaggregator-unixfs v0.2.0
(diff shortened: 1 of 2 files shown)


masih:
CPU usage remains high but for a different reason: the recursive unpacking of nested value-key marshals. This is the self correction of values by the new merger. 

Right now I'd like to keep it as is considering the node is responsive: both ingesting and finder works. I will keep an eye on it over the weekend.

.github/workflows/ecr-publisher.yml
```diff
@@ -14,6 +14,7 @@ on:
       - 'doc/**'
     branches:
       - main
+      - masih/pebble_zero_copy_marshal_dido # TODO remove before merge
```
willscott, line 17: rebase / clean up before merge.

willscott (approved):

PR #908, opened 2022-10-10
Added creation of ECR public repo

ischasny:
This PR defines a resource for creation of ECR public repo


.gitignore
```diff
@@ -2,3 +2,6 @@
 *~
 *.car
 *.out
+.terraform
+.vscode
+*.code-workspace
```
masih, line 7: Look at the above changes `.vscode` and `.code-worskace` look IDE specifc?
if so, please avoid IDE specific gitignore entries. Instead, consider setting up a global gitignore on your machine that filters out all such files across every repo for the IDE you use.

deploy/infrastructure/modules/ecr_public/main.tf
```diff
@@ -0,0 +1,10 @@
+provider "aws" {
+  alias  = "us_east_1"
```
masih, line 2: Use [existing](https://github.com/filecoin-project/storetheindex/blob/main/deploy/infrastructure/common/main.tf#L3) alias instead of re-declaring AWS provider.

deploy/infrastructure/common/ecr.tf
```diff
@@ -10,3 +10,10 @@ module "ecr_ue2" {
   ]
   tags = local.tags
 }
+
+module "ecr_public_ue1" {
```
masih, line 14: Avoid `module` considering it contains a single resource and LOC would be halved without it. Instead, use the `resource` straight up.

masih (changes requested):

masih (approved):

Issue #826, opened 2022-09-19
Investigate step index size growth rate on `romi`


masih:
The `romi` index size in `prod` seems to be growing very fast. This node uses JSON value codec which explains higher index size but not necessarily the growth. Understand why.

masih:
From 12th to 14th of Sep. there [doesn't seem to be an increase in ingest throughput](https://protocollabs.grafana.net/d/pPROWOD7z/providers?orgId=1&from=1662937200000&to=1663196399000&var-Source=storetheindex-prod&var-Jobs=storetheindex%2Findexer-single&var-pod=romi-indexer) on `romi`:

<img width="748" alt="image" src="https://user-images.githubusercontent.com/301855/190996915-5d7a0393-7338-4d0e-a784-ee68d62ef0d6.png">

But there is a consistent increase in disk usage and index size:
<img width="751" alt="image" src="https://user-images.githubusercontent.com/301855/190997083-845b2532-0378-472c-82e2-19fc7b81b082.png">

<img width="761" alt="image" src="https://user-images.githubusercontent.com/301855/190997218-dc0de291-0500-4d6b-9fc8-1a5371301a48.png">

[Disk was expanded to double the size](https://github.com/filecoin-project/storetheindex/pull/811) when `romi` hit >80%.

masih:
This seems to be happening again where the multihash ingest rate graph doesn't quite match the increase in disk usage. In one day the disk utilisation in romi increased by ~25% ( just over 10TB). where the mh ingest rate looks relatively flat.
[Here](https://protocollabs.grafana.net/d/pPROWOD7z/providers?orgId=1&from=1665356400000&to=1665529199000&var-Source=storetheindex-prod&var-Jobs=storetheindex%2Findexer-single&var-pod=romi-indexer) is the link to grafana dashboard for the referenced timespan.

@gammazero  any thoughts on this?

<img width="774" alt="image" src="https://user-images.githubusercontent.com/301855/195319077-4176b71c-21bd-4b44-acbd-735e7411adf2.png">
<img width="768" alt="image" src="https://user-images.githubusercontent.com/301855/195319218-f0a0dc59-7ca5-4a03-88ea-ab85178da3d9.png">
<img width="778" alt="image" src="https://user-images.githubusercontent.com/301855/195319347-75564de1-a242-4558-a7d4-7d798d484ef9.png">

ischasny:
Has such behaviour been observed on `pebble` storage too?

[...]

PR #915, opened 2022-10-11
Increase storage on romi

gammazero:


willscott (approved):

masih:
THis change has not been reflected on the K8S it looks like. Investigating. Cc @gammazero

masih:
This change did not get applied because we hit a maximum volume modification quota for io2. I have requested an increase which should get that resolved automatically. 

One thing to point out (also posted on slack): maximum volume size for io2 is 64Ti. I think we should look into [this](https://github.com/filecoin-project/storetheindex/issues/826) before any further disk expansion on `romi`.

masih:
Resolved; volume is now expanded.

PR #919, opened 2022-10-12
Exclude `dido` from indexstar while investigating merger CPU footprint

masih:
Exclude `dido` backend from indexstar while working on improving CPU footprint of f pebble merger. This is to offer a stable upstream find latency in indexstar.


ischasny (approved):
lgtm

PR #920, opened 2022-10-12
Deploy `indexstar` on `dev`

masih:
Stand up an instance of indexstar, similar to the set up in `prod`.

Relates to #856


masih:
Note the changes here only deploy indexstar; it does not change the routing on dev to point to it. Once service is up, a separate PR will make necessary changes to dev cloudfront.

willscott (approved):

PR #921, opened 2022-10-12
Point `dev.cid.contact` to CloudFront backed by `indexstar`

masih:
Change the routing of requests in `dev` environment, such that all requests to `dev.cid.contact` go through CloudFront caching, and routed via indexstar internally. This matches the routing mechanism currently employed in `prod`.

As a result of changes, ingress to `indexer-0` and `indexer-1` is changed to point to `indexer.dev.cid.contact` instead.

Fixes #856


deploy/infrastructure/dev/us-east-2/cloudfront.tf
```diff
@@ -50,6 +72,20 @@ resource "aws_cloudfront_distribution" "cdn" {
     max_ttl                = 86400
   }
 
+  ordered_cache_behavior {
+    path_pattern = "reframe"
+    # CloudFront does not support configuring allowed methods selectively.
+    # Hence the complete method list.
+    allowed_methods        = ["GET", "HEAD", "OPTIONS", "PUT", "DELETE", "PATCH", "POST"]
+    cached_methods         = ["GET", "HEAD"]
```
willscott, line 80: options could be cached, right?

willscott (approved):

PR #924, opened 2022-10-12
Enable nginx ingress metrics on `dev`

masih:
Enable nginx monitoring and metrics on `dev` so that we can measure upstream latency from ingress. This is already set up for `prod`.


ischasny (approved):
lgtm

PR #925, opened 2022-10-12
Deploy latest indexstar to `prod` to roll out resilience improvements

masih:
Testing in `dev` seems fine though hard to be confident considering it receives very little traffic.


willscott (approved):
🚀  - we should monitor as this is merged since i'm not sure we've fully tested the same load in the dev roll-out that we're handling in prod

masih:
Thank you for the review; I'll roll this out tomorrow with a fresh brain.

PR #929, opened 2022-10-13
Define an EKS node group dedicated to us-east2a for autoretrieve

masih:
The node group primarily getting used by autoretrieve is configured to use all subnets. This means ASG can spin up new nodes in any AZ which may not match the AZ on which PV was originally created.

The work here creates a node group with AZ dedicated to the zone on which the autoretrieve PV is currently presenent, i.e. `us-east2a`, in order to avoid unschedulable pods caused by AZ mismatch between worker node and PV.

The minimum instance count in the existing ASG is decreased to zero in order to avoid running idle workers once the autoretrieve pod migrates to the newly set up node group.


ischasny (approved):
Lgtm

Issue #931, opened 2022-10-14
High find latency with error `file already closed` in sth backed nodes


masih:
`indexer-0` and `indexer-1` in `prod` environment started experiencing [high find latency](https://protocollabs.grafana.net/d/pPROWOD7z/providers?orgId=1&from=1665701845694&to=1665736832881&var-Source=storetheindex-prod&var-Jobs=storetheindex%2Findexer&var-pod=All&viewPanel=22). Looking at the [logs](https://protocollabs.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Benvironment%3D%5C%22prod%5C%22,%20pod_name%3D~%5C%22indexer-1.%2A%5C%22%7D%20%7C~%20%5C%22error%5C%22%20%21~%20%5C%22datatransfer%5C%22%20%21~%20%5C%22graphsync%5C%22%20%21~%20%5C%22dt-%5C%22%5Cn%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221665717334901%22,%22to%22:%221665736686463%22%7D%7D) in oldest-first order it seems like the node came across bad index files which were removed. After that there seems to be `file already closed` errors occurring; example:  

```
cannot get multihash from store: error reading index records from disk: read /data/valuestore/storethehash.index.427016: file already closed
```

masih:
In the meantime, I will restart those nodes to mitigate the impact on ingress upstream latency.

gammazero:
[...]

PR #932, opened 2022-10-14
Exclude indexer 0 and 1 on prod from indexstar

masih:
Exclude `indexer-0` and `1` from `indexstar` on prod until they are restarted. This is a measure to reduce the impact on upstream latency caused by #931.


ischasny (approved):
Lgtm

PR #936, opened 2022-10-16
Health check should store and retrieve index

### LINKED ISSUE #900: Health check should store and retrieve index
Currently, the health check only checks that the admin server is running. This will not detect if the value store is inoperable. The health check should also check that values are being written to and read from the valuestore without error.

arnabguin:
Currently the health check endpoint in admin server only checks if server is running. It is now enhanced to verify that writes/reads to value store are working correctly as well.

Fixes #900


server/admin/http/handler.go
```diff
@@ -373,13 +373,23 @@ func batchIndexerEntries(batchSize int, putChan <-chan multihash.Multihash, valu
 // ----- admin handlers -----
 
 func (h *adminHandler) healthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	w.WriteHeader(http.StatusOK)
-	// TODO: Report on indexer core health?
-	_, err := w.Write([]byte("\"OK\""))
+	healthStatus := &HealthStatus{Server: "OK", ValueStore: "OK"}
```
gammazero, line 376: The HealthStatus contains two status values, yet the code only ever sets one or the other. So, in the case of an error, it seems sufficient to return only a single error that describes the failure.  Also, if there is no error, then returning a 200 should be sufficient to declare that there is no error, and returning `HealthStatus{Server: "OK", ValueStore: "OK"}` is not necessary.

When `json.Marshal` fails the error is returned as a string, and when there is a valuestore error the error is returned as json.  If think it would be more consistent to always return the same type of data, especially when the content type header is not being set.

If there is no need to return a parsable structure, then I would prefer to do what is simplest and return a string with a 500 response code in the case of any error, and a 200 in the case of no error.

server/admin/http/handler.go
```diff
@@ -373,13 +373,23 @@ func batchIndexerEntries(batchSize int, putChan <-chan multihash.Multihash, valu
...
+		healthStatus.Server = fmt.Sprintf("FAILED:Marshal failed:%s", err)
+		http.Error(w, healthStatus.Server, http.StatusInternalServerError)
+		return
+	}
+	if errVS != nil {
+		http.Error(w, string(bytes), http.StatusInternalServerError)
 		return
 	}
+	w.WriteHeader(http.StatusOK)
+	w.Write(bytes)
```
gammazero, line 392: Suggest:
```go
	w.WriteHeader(http.StatusOK)
	if _, err = w.Write([]byte("\"OK\"")); err != nil {
		log.Errorw("Cannot write HealthCheck response:", "err", err)
	}
```

server/admin/http/handler.go
```diff
@@ -394,3 +404,50 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	val := indexer.Value{
+		ProviderID:    provider,
+		ContextID:     []byte(mhash),
+		MetadataBytes: []byte("p-metadata"),
+	}
+	return mhash, &val, nil
+}
```
gammazero, line 453: Since this data is always the same, instead of decoding the peer ID and creating the value on every call, it would be better to create a static `indexer.Value`

```suggestion
var healthCheckMH multihash.Multihash
var healthCheckValue indexer.Value

func init() {
	provider, err := peer.Decode("12D3KooWBUNzpAz1Jfvnaag1nHBi6gbG5Q1BQm8kCfcpmt7bGKa6")
	if err != nil {
		panic(err.Error())
	}
	healthCheckMH = multihash.Multihash("2DrjgbFdhNiSJghFWcQbzw6E8y4jU1Z7ZsWo3dJbYxwGTNFmAj")
	healthCheckValue = indexer.Value{
		ProviderID:    provider,
		ContextID:     []byte(healthCheckMH),
		MetadataBytes: []byte("healthcheck-metadata"),
	}
}
```

server/admin/http/healthstatus.go
```diff
@@ -0,0 +1,17 @@
...
+
+type ValueStoreFailReason string
+
+const (
+	VS_GEN_FAILURE        ValueStoreFailReason = "Unable to generate valuestore datum"
+	VS_WRITE_FAILURE                           = "Cannot store value in valuestore"
+	VS_READ_FAILURE                            = "Cannot get value from valuestore"
+	VS_NO_DATA                                 = "Value not present in valuestore"
+	VS_READWRITE_MISMATCH                      = "Value stored does not match value retrieved from valuestore"
+	VS_REMOVE_FAILURE                          = "Unable to remove value from valuestore"
```
gammazero, line 16: Is there a reason all these consts, and their type, are exported? Are you expecting some other code to import this module to reference these?
https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

server/admin/http/healthstatus.go
```diff
@@ -0,0 +1,17 @@
...
+)
```
gammazero, line 17: I do not think the `HealthStatus` type is needed, as per my comments above. If you see it as adding value, please describe why.

If you do not think `HealthStatus` adds value, then this file can be removed.

server/admin/http/handler.go
```diff
@@ -394,3 +404,50 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
 	}
 	return peerID, true
 }
+
+func healthCheckValueStore(h *adminHandler) (cause ValueStoreFailReason, err error) {
+	mhash, val, err := createValueStoreDatum()
```
gammazero, line 409: You can use static values instead. See comment on `createValueStoreDatum()`

gammazero (changes requested):
Thank you for handling this issue. I have commented on some simplifications that I would prefer unless there is good reason to do otherwise.

server/admin/http/handler.go
```diff
@@ -394,3 +404,50 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
...
+		return VS_READ_FAILURE, err
+	}
+	if !present {
+		return VS_NO_DATA, err
+	}
+	if !val.Equal(rval[0]) {
+		return VS_READWRITE_MISMATCH, err
+	}
+
+	return "", nil
```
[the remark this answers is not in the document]
arnabguin: I was trying to introduce the concept of an error code but since these messages won't be used anywhere else, makes sense to go with just messages. Made necessary changes.

server/admin/http/handler.go
```diff
@@ -394,3 +404,50 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
...
+	if err != nil {
+		return VS_WRITE_FAILURE, err
+	}
+
+	defer func() {
+		if errorOnRemove := h.indexer.Remove(*val, mhash); errorOnRemove != nil {
+			cause = VS_REMOVE_FAILURE
+			err = errorOnRemove
+		}
+	}()
```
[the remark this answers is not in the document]
arnabguin: done

server/admin/http/handler.go
```diff
@@ -394,3 +404,50 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
...
+func healthCheckValueStore(h *adminHandler) (cause ValueStoreFailReason, err error) {
+	mhash, val, err := createValueStoreDatum()
+	if err != nil {
+		return VS_GEN_FAILURE, err
+	}
+
+	err = h.indexer.Put(*val, mhash)
+	if err != nil {
+		return VS_WRITE_FAILURE, err
+	}
```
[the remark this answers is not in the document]
arnabguin: moved to init() method and removed createValueStoreDatum in accordance with some of your latter review comments.

server/admin/http/handler.go
```diff
@@ -373,13 +373,23 @@ func batchIndexerEntries(batchSize int, putChan <-chan multihash.Multihash, valu
...
 
 func (h *adminHandler) healthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	w.WriteHeader(http.StatusOK)
-	// TODO: Report on indexer core health?
-	_, err := w.Write([]byte("\"OK\""))
+	healthStatus := &HealthStatus{Server: "OK", ValueStore: "OK"}
+	var errVS error
+	if cause, errVS := healthCheckValueStore(h); errVS != nil {
+		healthStatus.ValueStore = fmt.Sprintf("FAILED:%s:%s", cause, errVS)
+	}
```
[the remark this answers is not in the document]
arnabguin: sure

server/admin/http/handler.go
arnabguin: I was trying to distinguish between valuestore and server status but agree this makes it simpler. did changes.

server/admin/http/healthstatus.go
arnabguin: I was thinking of an enum that could be standardized and used elsewhere too. But if they are used only inside the indexer, makes sense to go with just messages.

server/admin/http/healthstatus.go
arnabguin: HealthStatus is not required anymore with simplification of the health handler code. removed.

server/admin/http/handler.go
arnabguin: done.

server/admin/http/handler.go
```diff
@@ -394,3 +395,42 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
...
+func init() {
+	provider, err := peer.Decode("12D3KooWBUNzpAz1Jfvnaag1nHBi6gbG5Q1BQm8kCfcpmt7bGKa6")
+	if err != nil {
+		panic(err.Error())
+	}
+	healthCheckMH = multihash.Multihash("2DrjgbFdhNiSJghFWcQbzw6E8y4jU1Z7ZsWo3dJbYxwGTNFmAj")
+	healthCheckValue = indexer.Value{
+		ProviderID:    provider,
+		ContextID:     []byte(healthCheckMH),
+		MetadataBytes: []byte("healthcheck-metadata"),
```
masih, line 411: Use a valid metadata to avoid lower level errors in case this is checked for correctness.

server/admin/http/handler.go
```diff
@@ -394,3 +395,42 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
 	}
 	return peerID, true
 }
+
+var healthCheckMH multihash.Multihash
+var healthCheckValue indexer.Value
+
+func init() {
+	provider, err := peer.Decode("12D3KooWBUNzpAz1Jfvnaag1nHBi6gbG5Q1BQm8kCfcpmt7bGKa6")
```
masih, line 403: We should exclude this provider+context ID from the find requests to avoid leaking bogus responses to the end user.

server/admin/http/handler.go
```diff
@@ -394,3 +395,42 @@ func decodePeerID(id string, w http.ResponseWriter) (peer.ID, bool) {
 	}
 	return peerID, true
 }
+
+var healthCheckMH multihash.Multihash
+var healthCheckValue indexer.Value
+
+func init() {
```
masih, line 402: minor comment, feel free to postpone for future improvements: 
Ideally, the data here should be mutated periodically: if kept fixed, an optimised version of the store could detect no changes and return without performing IO operations. This then reduces our confidence in healthcheck passing.

masih (commented):
Apologies for the late review; left some comments and suggestions. 

The top comment I recommend reflecting on is exclusion of healthcheck provider from results returned by Finder server (including `/providers`).

PR #944, opened 2022-10-18
Route `ingest/*` requests directly to backend nodes

masih:
`indexstar` does not multiplex over `ingest/*` requests. As a result announce requests won't reach the backend unless sent to explicit instance names.

Configure cloudfront such that `ingest/*` requests are sent directly to backend nodes, more specifically `indexer-0` and `indexer-1`. Because, those nodes are configured to further propagate HTTP announce requests over gossipsub which are then picked up by other instances.


ischasny (approved):
lgtm

Issue #953, opened 2022-10-21
Providers with low ad generation rate get excluded and marked as inactive


masih:
I have been manually triggering ingest from Sofia Miner. After a while it seems to disappear from the list of providers and none of the ingested records are findable. 

Taking a closer look at the registry logic in storetheindex, i suspect what’s happening is that the provider gets marked as “inactive” and excluded from everything as if it doesn’t exist. Providers marked as inactive typically are the ones that have not been seen for [2XretryAfter](https://github.com/filecoin-project/storetheindex/blob/b7e28d64dce97c53d1f88ae77fac72877aa890f6/internal/registry/registry.go#L882), i.e. 2 hours.

What marks a provider as seen is either registerOrUpdate (which happens on first discovery or address change) or a function call to Registry.Saw which seems to only be called if a provider [had non zero number of ads](https://github.com/filecoin-project/storetheindex/blob/b7e28d64dce97c53d1f88ae77fac72877aa890f6/internal/ingest/ingest.go#L815) over the past retryAfter ? I think this means providers that are perfectly contactable but haven’t had published any ads for an hour get marked as inactive?

I suspect this is the cause of low provider numbers on /providers (currently 63) compared to number of providers getting loaded into registry ([206 max over the last 2 days](https://protocollabs.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Benvironment%3D~%5C%22prod%5C%22,%20namespace%3D~%5C%22storetheindex%5C%22%7D%20%7C~%20%5C%22loaded%20providers%20into%20registry%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-2d%22,%22to%22:%22now%22%7D%7D)).

masih:
All known poroviders on nodes to which the fix was deployed now appear at /providers endpoint. For example, dido now has 206 known providers.

PR #954, opened 2022-10-21
Mark provider as Seen when sync is successful

masih:
A provider may be marked as inactive if its last contact time is not updated for 2x longer than a configured threshold. The last contact time was previously being updated only of there was some ads from a provider: an active provider that has not published any new ads for a while could get marked as inactive.

The changes here opportunistically marks a provider as seen: i.e. whenever sync from that provider doesn't return an error. If subscriber. Sync does not return an error it means there was some successful communication to either fetch data or get the head.

Relates to #953


willscott (approved):

masih:
Going to deploy this on a few nodes and observer if sofia miner remains in registry before cutting a release.

PR #956, opened 2022-10-21
Keep providers in active state until deleted by default

### LINKED ISSUE #953: Providers with low ad generation rate get excluded and marked as inactive

masih:
Introduce a new config parameter `DeactivateAfter` to replace the hardcoded `2xretryAfter` duration after which a provider is deactivated.

Change the default behaviour to set `DeactivateAfter` to the same value as `PollStopAfter`. This effectively disables the inactive state feature for providers: i.e. providers will remain findable via query APIs until deleted when unseen after `PollStopAfter` has elapsed.

Fixes #953

### DIFF
--- a/config/discovery.go
+++ b/config/discovery.go
@@ -29,11 +29,18 @@ type Discovery struct {
 	// there to be more than one poll attempt for a provider.
 	PollRetryAfter Duration
 	// PollStopAfter is the amount of time, from the start of polling, to
-	// continuing polling for the latest advertisment without getting a
+	// continuing polling for the latest advertisement without getting a
 	// response. After this time elapses with no updates from the provider, the
 	// provider's data is removed from the indexer and must be re-fetched if
 	// the provider returns.
 	PollStopAfter Duration
+	// DeactivateAfter is the duration after which an unseen provider will be
+	// excluded from find queries as well as providers list. Note, an inactive
+	// provider will not be deleted until PollStopAfter has elapses since last
+	// it was seen. If unspecified, providers will be removed once PollStopAfter
+	// has elapsed without entering inactive state grace period. If set to less
+	// than PollStopAfter this parameter will have no effect.
+	DeactivateAfter Duration
 	// PollOverrides configures polling for specific providers.
 	PollOverrides []Polling
 	// RediscoverWait is the amount of time that must pass before a provider
@@ -57,18 +64,22 @@ type Polling struct {
 	RetryAfter Duration
 	// StopAfter overrides Discovery.PollStopAfter.
 	StopAfter Duration
+	// DeactivateAfter overrides Discovery.DeactivateAfter.
+	DeactivateAfter Duration
 }
 
 // NewDiscovery returns Discovery with values set to their defaults.
 func NewDiscovery() Discovery {
+	const defaultStopAfter = Duration(7 * 24 * time.Hour)
 	return Discovery{
-		LotusGateway:   "https://api.chain.love",
-		Policy:         NewPolicy(),
-		PollInterval:   Duration(24 * time.Hour),
-		PollRetryAfter: Duration(5 * time.Hour),
-		PollStopAfter:  Duration(7 * 24 * time.Hour),
-		RediscoverWait: Duration(5 * time.Minute),
-		Timeout:        Duration(2 * time.Minute),
+		LotusGateway:    "https://api.chain.love",
+		Policy:          NewPolicy(),
+		PollInterval:    Duration(24 * time.Hour),
+		PollRetryAfter:  Duration(5 * time.Hour),
+		PollStopAfter:   defaultStopAfter,
+		DeactivateAfter: defaultStopAfter,
+		RediscoverWait:  Duration(5 * time.Minute),
+		Timeout:         Duration(2 * time.Minute),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


willscott (approved):

gammazero:
I would prefer not to add another config option around this behavior, and I think we should decide on one of the following:
1. Continue returning the provider in responses until they are delisted.
2. Stop returning providers during polling, if after at least one poll they are not responsive (no response to sync).

It seems that PR #954  is sufficient for option 2.

masih:
I think 1 is a more straight forward one. This PR simply parameterises the deactivation duration which allows both behaviours defaulting to 1. 

I am happy to remove code such that only 1 is possible. Though, i think retaining the ability to momentarily exclude from results considering it is already implemented until behaviour 1 has had some trial run.

gammazero:
If we go with option 1, then we should also remove the `inactive` flag.

config/discovery.go
```diff
@@ -57,18 +64,21 @@ type Polling struct {
...
-		PollRetryAfter: Duration(5 * time.Hour),
-		PollStopAfter:  Duration(7 * 24 * time.Hour),
-		RediscoverWait: Duration(5 * time.Minute),
-		Timeout:        Duration(2 * time.Minute),
+		LotusGateway:    "https://api.chain.love",
+		Policy:          NewPolicy(),
+		PollInterval:    Duration(24 * time.Hour),
+		PollRetryAfter:  Duration(5 * time.Hour),
+		PollStopAfter:   Duration(7 * 24 * time.Hour),
+		DeactivateAfter: Duration(7 * 24 * time.Hour),
```
gammazero, line 79: Maybe this should explicitly be set to the same value as `PollStopAfter` so that it tracks whatever that is set to:
```go
	const defaultPollStopAfter = Duration(7 * 24 * time.Hour)
...
		PollStopAfter:   defaultPollStopAfter,
		DeactivateAfter: defaultPollStopAfter,
```

internal/registry/registry.go
```diff
@@ -881,7 +884,7 @@ func (r *Registry) pollProviders(poll polling, pollOverrides map[peer.ID]polling
 				}
 				// Tell the ingester to remove data for the provider.
 				info.deleted = true
-			} else if sincePollingStarted >= 2*poll.retryAfter {
+			} else if sincePollingStarted >= poll.deactivateAfter {
```
gammazero, line 887: Update the comment below, because now it is not after at least one poll attempt, it is after `deactivateAfter`

internal/registry/registry_test.go
```diff
@@ -433,15 +434,15 @@ func TestPollProvider(t *testing.T) {
 	select {
 	case pinfo := <-r.SyncChan():
 		require.Equal(t, pinfo.AddrInfo.ID, peerID, "unexpected provider info on sync channel, expected %q got %q", peerID.String(), pinfo.AddrInfo.ID.String())
-		require.True(t, pinfo.Inactive(), "Expected provider to be marked inactive")
+		require.False(t, pinfo.Inactive(), "Expected provider to not be marked inactive")
```
gammazero, line 437: I think this means that the test that a provider becomes inactive in now missing.  Instead of changing the lin above, I think that `deactivateAfter` shold be set to a value that makes the provider become inactive and the original test condition to be met.

internal/registry/registry_test.go
```diff
@@ -433,15 +434,15 @@ func TestPollProvider(t *testing.T) {
...
 	case <-timeout:
 		t.Fatal("Expected sync channel to be written")
 	}
 
 	// Inactive provider should not be returned.
 	pinfo, _ := r.ProviderInfo(peerID)
 	require.NotNil(t, pinfo, "expected inactive provider to still be present")
-	require.True(t, pinfo.Inactive(), "expected provider to be inactive")
+	require.False(t, pinfo.Inactive(), "expected provider to not be inactive")
```
gammazero, line 445: Same comment as above.

gammazero (changes requested):
I think the test for a provider becoming inactive is being missed. See comments.

gammazero (approved):

PR #963, opened 2022-10-23
Expand `xabi` PVC to 40 Ti

masih:
`xabi` is at 93% utilisation.


masih:
Merging this before we hit the limit.

PR #965, opened 2022-10-25
Find latency metric should split 'findbatch' and 'find'

### LINKED ISSUE #179: find latency metric should split 'findbatch' and 'find'
they will take different amounts of time potentially

arnabguin:
Fixes #179

### DIFF
--- a/server/finder/http/handler.go
+++ b/server/finder/http/handler.go
@@ -80,9 +80,15 @@ func (h *httpHandler) getIndexes(w http.ResponseWriter, mhs []multihash.Multihas
 	var found bool
 	defer func() {
 		msecPerMh := coremetrics.MsecSince(startTime) / float64(len(mhs))
+		var measure *stats.Float64Measure
+		if len(mhs) == 1 {
+			measure = metrics.FindLatency
+		} else {
+			measure = metrics.FindBatchLatency
+		}
 		_ = stats.RecordWithOptions(context.Background(),
 			stats.WithTags(tag.Insert(metrics.Method, "http"), tag.Insert(metrics.Found, fmt.Sprintf("%v", found))),
-			stats.WithMeasurements(metrics.FindLatency.M(msecPerMh)))
+			stats.WithMeasurements(measure.M(msecPerMh)))
 	}()
 
 	response, err := h.finderHandler.Find(mhs)
--- a/internal/metrics/server.go
+++ b/internal/metrics/server.go
@@ -23,6 +23,7 @@ var (
 // Measures
 var (
 	FindLatency          = stats.Float64("find/latency", "Time to respond to a find request", stats.UnitMilliseconds)
+	FindBatchLatency     = stats.Float64("find/batchlatency", "Time to respond to a find batch request", stats.UnitMilliseconds)
 	IngestChange         = stats.Int64("ingest/change", "Number of syncAdEntries started", stats.UnitDimensionless)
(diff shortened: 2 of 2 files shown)


gammazero (commented):
@willscott Look at [comment](https://github.com/filecoin-project/storetheindex/issues/179#issuecomment-1292569006) in issue #179 and decide if this still has value. If yes, then code if fine.

server/finder/http/handler.go
```diff
@@ -80,9 +80,15 @@ func (h *httpHandler) getIndexes(w http.ResponseWriter, mhs []multihash.Multihas
 	var found bool
 	defer func() {
 		msecPerMh := coremetrics.MsecSince(startTime) / float64(len(mhs))
```
gammazero, line 82: Remove this, since the PR needs total elapsed time, not the average per MH:
```suggestion
```

server/finder/http/handler.go
```diff
@@ -80,9 +80,15 @@ func (h *httpHandler) getIndexes(w http.ResponseWriter, mhs []multihash.Multihas
...
+		var measure *stats.Float64Measure
+		if len(mhs) == 1 {
+			measure = metrics.FindLatency
+		} else {
+			measure = metrics.FindBatchLatency
+		}
 		_ = stats.RecordWithOptions(context.Background(),
 			stats.WithTags(tag.Insert(metrics.Method, "http"), tag.Insert(metrics.Found, fmt.Sprintf("%v", found))),
-			stats.WithMeasurements(metrics.FindLatency.M(msecPerMh)))
+			stats.WithMeasurements(measure.M(msecPerMh)))
```
gammazero, line 91:
```suggestion
			stats.WithMeasurements(measure.M(coremetrics.MsecSince(startTime))))
```

gammazero (changes requested):
A couple of changes are needed for this PR.

gammazero (approved):

server/finder/http/handler.go
```diff
@@ -79,10 +79,15 @@ func (h *httpHandler) getIndexes(w http.ResponseWriter, mhs []multihash.Multihas
...
+		var measure *stats.Float64Measure
+		if len(mhs) == 1 {
+			measure = metrics.FindLatency
+		} else {
+			measure = metrics.FindBatchLatency
+		}
 		_ = stats.RecordWithOptions(context.Background(),
 			stats.WithTags(tag.Insert(metrics.Method, "http"), tag.Insert(metrics.Found, fmt.Sprintf("%v", found))),
-			stats.WithMeasurements(metrics.FindLatency.M(msecPerMh)))
+			stats.WithMeasurements(measure.M(coremetrics.MsecSince(startTime))))
```
gammazero, line 90:
```suggestion
			stats.WithMeasurements(measure.M(coremetrics.MsecSince(startTime)/ float64(len(mhs)))))
```

gammazero (commented):
Do need avg time per mh here.

gammazero:
@masih I do not think we want this change. It will show a different metric when searching for a single multihash versus multiple multihashes.  A larger number of multihashes will always show a smaller average latency per MH since there are fewer calls to the core.

If we want a more useful metric then we should look at latency as a function of multihashes in the query. Here, we are not keeping the multihahs count, so I do not think this is very useful. I propose closing this PR and either closing #179, or rewriting #179 to ask for a latency per multihash set size (if that is still useful). WDYT?

PR #978, opened 2022-10-29
Include Bitswap providers with multiple transport in reframe response

### LINKED ISSUE #977: Reframe server should check if metadata contains bitswap
instead it expect the provider record to be exactly one and exactly bitswap [here](https://github.com/filecoin-project/storetheindex/blob/b7e28d64dce97c53d1f88ae77fac72877aa890f6/server/reframe/reframe.go#L89). This results in filtering out results unfairly.

masih:
The reframe server implementation only returned providers that supported exactly one transport protocol in their metadata: `TransportBitswap`.

This implementation was added at the time where Provider record metadata did not support multiple transports. Later, when the feature was added in Metadata spec, the reframe server was not updated to reflect it.

The changes here fix the issue by checking if the metadata is prefixed with `TransportBitswap` codec and if so it will include the records in response. Check for prefix is sufficient here because, metadata bytes must be sorted by codec as stated by the network indexer protocol.

Fixes #977

### DIFF
--- a/server/reframe/reframe.go
+++ b/server/reframe/reframe.go
@@ -58,7 +58,7 @@ func (x *ReframeService) FindProviders(ctx context.Context, key cid.Cid) (<-chan
 			continue
 		}
 		for _, pr := range mhr.ProviderResults {
-			if !isBitswapMetadata(pr.Metadata) {
+			if !containsTransportBitswap(pr.Metadata) {
 				continue
 			}
 			peerAddrs = append(peerAddrs, pr.Provider)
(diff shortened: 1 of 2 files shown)


server/reframe/reframe.go
```diff
@@ -85,6 +85,10 @@ func (x *ReframeService) Provide(context.Context, *client.ProvideRequest) (<-cha
...
 var BitswapMetadataBytes = varint.ToUvarint(uint64(multicodec.TransportBitswap))
 
-func isBitswapMetadata(meta []byte) bool {
-	return bytes.Equal(meta, BitswapMetadataBytes)
+func containsTransportBitswap(meta []byte) bool {
+	// Metadata must be sorted according to the specification; see:
+	// - https://github.com/filecoin-project/index-provider/blob/main/metadata/metadata.go#L143
+	// This implies that if it includes Bitswap, its codec must appear at the beginning
+	// of the metadata value. Hence, bytes.HasPrefix.
+	return bytes.HasPrefix(meta, BitswapMetadataBytes)
```
jacobheun, line 93: A note for future humans - this will be dependent on https://github.com/multiformats/multicodec/blob/5e275cdb00e68e76663bbcc996f2ede47bb5b301/table.csv#L138 maintaining order in the future, which is a reasonable assumption.

jacobheun (approved):
❤️ thanks for patching this!

server/reframe/reframe.go
willscott: :+1: good catch

PR #981, opened 2022-10-29
Deploy #977 fixes to indexstar

masih:
Deploy fixes to #977 on indexstar so that bitswap poviders are returned correctly over reframe queries.


masih:
**Note: Merge after IPFS Camp Workshop is done in case of disruptions.**

Issue #891, opened 2022-10-05
Install external snapshotter controller and CRDs on K8S


masih:
The [external snapshotter](https://github.com/kubernetes-csi/external-snapshotter) controller allows a user to create snapshots with some retention policy by simply defining YAML manifests. This is useful for testing purposes in storetheindex. For example, we can perform migrations or re-scanning of data without downtime, while defining the entire process in a single PR without the need to perform any manual steps. 

Also see: https://kubernetes.io/docs/concepts/storage/volume-snapshots/

PR #986, opened 2022-11-01
Set up external snapshotter controller on `dev`

masih:
Configure and run external snapshotter on `dev` cluster with a default volume snapshot class.

Relates to #891

### DIFF
--- a/deploy/manifests/base/external-snapshotter/kustomization.yaml
+++ b/deploy/manifests/base/external-snapshotter/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - github.com/kubernetes-csi/external-snapshotter/client/config/crd?ref=v6.1.0
+  - snapshot-controller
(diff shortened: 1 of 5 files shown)


willscott:
this is just to test for having these for prod, i guess?

Eventually we should aim to have a process (e.g. with mirroring) such that we're able to re-spin-up replicas cheaply enough that we don't need so much focus on reliability of individual instances 🤞

masih:
This is exactly why we are adding snapshotting capability; for example: to spin up a replica of `dido` with this work we now can take a snapshot of the volume by defining a YAML, and then mount that as a separate PVC onto a brand new node and get a new instance with  current state of dido without having to wait for nodes to catch up. 

Adding it to dev first to make sure it works before bubbling up the changes (plus the upgrade to CSI driver onto prod)

masih:
On replication, what we want is sharded handling of providers via the coordination service, and for that we will be spinning up new nodes. Using snapshots it means those new nodes don't need to have completely empty state. They can start with the state of one of the most up to date nodes.

willscott (approved):

PR #988, opened 2022-11-02
Upgrade CSI driver and install external snapshotter on `prod`

### LINKED ISSUE #890: Upgrade EBS CSI controller
Upgrade to the latest EBS CSI controller which exposes much more convenient storage class parameters for configuring io2 volumes; see: 
* https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/parameters.md

As part of the upgrade, make sure that all existing storage class configurations continue to function.

### LINKED ISSUE #891: Install external snapshotter controller and CRDs on K8S

masih:
Upgrade to the latest CSI driver on `prod` now that it have been tested on `dev` and deploy external snapshotter on `prod` so that we can take snapshot of existing node's PVC and mount it as the state of new nodes. This then would reduce time spent on catching up as well as making testing on real data much easier.

Fixes #890
Fixes #891

### DIFF
--- a/deploy/manifests/base/aws-ebs-csi-driver/kustomization.yaml
+++ b/deploy/manifests/base/aws-ebs-csi-driver/kustomization.yaml
@@ -2,4 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-  - github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable?ref=v1.5.1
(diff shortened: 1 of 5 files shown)


ischasny (approved):
Lgtm

PR #989, opened 2022-11-02
Deploy `kepa` as replacement of `indexer-1` backed by `pebble`

masih:
Deploy a new node, called `kepa`, to replace the existing `indexer-1`. The node reuses the identity of `indexer-1` since its identity is one of the ones whitelisted by lotus bootstrap nodes.

`kepa` uses `pebble` as the backing store, sourced from a snapshot of `dido` PVC taken on 02/11/2022. This reduces  the catch-up time, while immediately making `kepa` capable of serving clients as soon as it is up. To offer better resilient to AZ failures, `kepa` is configured to run on `us-east2b`, whereas `dido` is currently running on `us-east2a`.

On disk IOPS, `kepa` is configured to use 3 IOPS per GiB, 2 less than `dido`. The rationale for this change is to see the impact of IOPS on ingest throughput and potentially start reducing IOPS across nodes if under-utilised or makes no difference considering its cost implications.

Note that `kepa` does not have any peering configuration, with `ResendDirectAnnounce` enabled in order to propagate HTTP announcements over to other nodes via gossipsub, just like its predecessor `indexer-1`.


masih:
**Do Not Merge: Pending `dido` volume snapshot completion; currently at 18%. It should finish snapshotting by tomorrow.**

willscott (approved):

PR #999, opened 2022-11-05
Refactor announce handling logic from go-legs into announce package here.

gammazero:
Refactors functionality for receiving, filtering, and resending announce messages from go-legs to storetheindex/announce.

See also: https://github.com/filecoin-project/go-legs/pull/169

This package is an example of one that might be put into a shared `libipni` repo at some point.


masih (approved):
This seems to largely be a fork of code from go-legs. I read this PR as move to reduce effort in maintaining go-legs as a generic syncing protocol and instead focus on doing what makes most sense in the context of syncing dags for storetheindex. I welcome this :+1:

However, I am not sure if there is much value in having go-legs depend on storetheindex as expained [here](https://github.com/filecoin-project/go-legs/pull/169#discussion_r1016356238). 

I think we should either leave the go-legs and stoetheindex relationship as is (i.e. avoid this refactoing altogether), or fork the code from go-legs into storetheindex and reduce effort to maintain go-legs. But not both.

PR #1002, opened 2022-11-07
Add Extended Providers implementation

ischasny:
This PR adds implementation for extended providers. For more details refer to the spec [here](https://github.com/filecoin-project/storetheindex/pull/804/files).

This PR adds backward-compatible support on the indexer side only. There going to be a second part to it that will have user-facing changes in index-provider.

### DIFF
--- a/api/v0/ingest/schema/envelope.go
+++ b/api/v0/ingest/schema/envelope.go
@@ -19,6 +19,7 @@ var log = logging.Logger("indexer/schema")
 const (
 	adSignatureCodec  = "/indexer/ingest/adSignature"
 	adSignatureDomain = "indexer"
+	epSignatureCodec  = "/indexer/ingest/extendedProviderSignature"
 )
 
 type advSignatureRecord struct {
@@ -47,6 +48,35 @@ func (r *advSignatureRecord) MarshalRecord() ([]byte, error) {
 
 func (r *advSignatureRecord) UnmarshalRecord(buf []byte) error {
 	r.advID = buf
+	return nil
+}
+
+type epSignatureRecord struct {
+	domain  *string
+	codec   []byte
+	payload []byte
+}
+
+func (r *epSignatureRecord) Domain() string {
+	if r.domain != nil {
+		return *r.domain
+	}
+	return adSignatureDomain
+}
+
+func (r *epSignatureRecord) Codec() []byte {
+	if r.codec != nil {
+		return r.codec
+	}
+	return []byte(epSignatureCodec)
+}
+
+func (r *epSignatureRecord) MarshalRecord() ([]byte, error) {
+	return r.payload, nil
+}
+
+func (r *epSignatureRecord) UnmarshalRecord(buf []byte) error {
+	r.payload = buf
 	return nil
 }
 
@@ -89,8 +119,62 @@ func signaturePayload(ad *Advertisement, oldFormat bool) ([]byte, error) {
 	return multihash.Sum(sigBuf.Bytes(), multihash.SHA2_256, -1)
 }
 
+// extendedSignaturePayload generates the data payload used to compute the signature for a provider from the ExtendedProviders list.
+// This payload *doesn't* contain enough information to sign the ad itself.
+func extendedProviderSignaturePayload(ad *Advertisement, p *Provider) ([]byte, error) {
+	if ad.IsRm {
+		return nil, fmt.Errorf("rm ads are not supported for extended provider signatures")
+	}
+
+	bindex := cid.Undef.Bytes()
+	if ad.PreviousID != nil {
+		bindex = ad.PreviousID.(cidlink.Link).Cid.Bytes()
+	}
+	ent := ad.Entries.(cidlink.Link).Cid.Bytes()
+
+	// Extended signature is an authrorisation for the publisher of the main Ad to publish on behalf of the signee.
+	// For more details see https://github.com/filecoin-project/storetheindex/pull/804/files
+	// The signature must contain the following fields:
+	// - the main provider's identity
+	// - the signee's identity addrs and metadata so that they can't be misrepresented
+	// - contextID and override flag as they change the behaviour of how extended providers are interpreted (see the spec)
+
+	var sigBuf bytes.Buffer
+
+	var addrsLen int
+	for _, addr := range p.Addresses {
+		addrsLen += len(addr)
+	}
+
+	sigBuf.Grow(len(bindex) + len(ent) + len(ad.Provider) + len(ad.ContextID) + len(p.ID) + addrsLen + len(p.Metadata) + 1)
+	sigBuf.Write(bindex)
+	sigBuf.Write(ent)
+	sigBuf.WriteString(ad.Provider)
+	sigBuf.Write(ad.ContextID)
+	sigBuf.WriteString(p.ID)
+	for _, addr := range p.Addresses {
+		sigBuf.WriteString(addr)
+	}
+	sigBuf.Write(p.Metadata)
+	if ad.ExtendedProvider.Override {
+		sigBuf.WriteByte(1)
+	} else {
+		sigBuf.WriteByte(0)
+	}
+
+	return multihash.Sum(sigBuf.Bytes(), multihash.SHA2_256, -1)
+}
+
 // Sign signs an advertisement using the given private key.
+// This function will return an error if used to sign an ad with extended providers.
 func (ad *Advertisement) Sign(key crypto.PrivKey) error {
+	if ad.ExtendedProvider != nil {
+		return fmt.Errorf("the ad can not be signed because it has extended providers")
+	}
+	return ad.signAd(key)
+}
+
+func (ad *Advertisement) signAd(key crypto.PrivKey) error {
 	advID, err := signaturePayload(ad, false)
 	if err != nil {
 		return err
@@ -108,12 +192,68 @@ func (ad *Advertisement) Sign(key crypto.PrivKey) error {
 	return nil
 }
 
+// SignWithExtendedProviders signs an advertisement by the main provider as well as by all extended providers if they are present.
+func (ad *Advertisement) SignWithExtendedProviders(key crypto.PrivKey, extendedProviderKeyFetcher func(string) (crypto.PrivKey, error)) error {
+	err := ad.signAd(key)
+	if err != nil {
+		return err
+	}
+
+	if ad.ExtendedProvider == nil {
+		return nil
+	}
+
+	seenTopLevelProvider := false
+
+	for i := range ad.ExtendedProvider.Providers {
+		p := &ad.ExtendedProvider.Providers[i]
+		if p.ID == ad.Provider {
+			seenTopLevelProvider = true
+		}
+
+		payload, err := extendedProviderSignaturePayload(ad, p)
+		if err != nil {
+			return err
+		}
+
+		var privKey crypto.PrivKey
+		if p.ID == ad.Provider {
+			privKey = key
+		} else {
+			privKey, err = extendedProviderKeyFetcher(p.ID)
+			if err != nil {
+				return err
+			}
+		}
+
+		envelope, err := record.Seal(&epSignatureRecord{payload: payload}, privKey)
+		if err != nil {
+			return err
+		}
+
+		sig, err := envelope.Marshal()
+		if err != nil {
+			return err
+		}
+
+		p.Signature = sig
+	}
+
+	if !seenTopLevelProvider && len(ad.ExtendedProvider.Providers) > 0 {
+		return fmt.Errorf("extended providers must contain provider from the encapsulating advertisement")
+	}
+
+	return nil
+}
+
 // VerifySignature verifies that the advertisement has been signed and
 // generated correctly.  Returns the peer ID of the signer.
 //
 // The signer may be different than the provider ID in the advertisement, so
 // the caller will need to check if the signer is allowed to sign this
 // advertisement.
+//
+// Extended providers signatures are also verified.
 func (ad *Advertisement) VerifySignature() (peer.ID, error) {
 	// sigSize is the size of the current signature.  Any signature that is not
 	// this size is the old signature format.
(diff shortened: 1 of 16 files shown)


api/v0/ingest/schema/envelope.go
```diff
@@ -150,5 +290,37 @@ func (ad *Advertisement) VerifySignature() (peer.ID, error) {
 		log.Warnw("advertisement has deprecated signature format", "signer", signerID)
 	}
 
+	if ad.ExtendedProvider != nil {
+		rec := &epSignatureRecord{}
+		// The top level provider must appear in the list of extended providers otherwise the ad is considered invalid
+		seenTopLevelProv := false
+		for _, p := range ad.ExtendedProvider.Providers {
```
gammazero, line 297: Do we need to check for duplicate providers here?

api/v0/ingest/schema/schema.ipldsch
```diff
@@ -29,9 +57,14 @@ type Advertisement struct {
     #  * https://ipld.io/specs/advanced-data-layouts/hamt/spec/#use-as-a-set
     Entries Link
     # ContextID is the unique identifier for the collection of advertised multihashes.
+    # If a Provider listing is written with no ContextID and IsRm=false, peers from ExtendedProvider 
+    # will be returned for all advertisements published by the publisher.
```
gammazero, line 61: I assume that the indexer's list is updated every time there is a provider listing with no context ID, and that the indexer only returns the latest version of the list of providers, regardless of what CID is queried. Is that correct?

internal/ingest/linksystem.go
```diff
@@ -191,29 +192,72 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
...
 
 	provider := peer.AddrInfo{
 		ID:    providerID,
 		Addrs: maddrs,
 	}
-	err = ing.reg.RegisterOrUpdate(context.Background(), provider, adCid, publisher)
+
+	ctx := context.Background()
+
+	err = ing.reg.RegisterOrUpdate(ctx, provider, adCid, publisher)
```
gammazero, line 204: I think that there should only be one call to the registry to register the provider as well as any extended providers.  I think it would make sense to add an `extendedProviders` argument to `RegisterOrUpdate`

internal/ingest/linksystem.go
```diff
@@ -191,29 +192,72 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
...
 
-	log = log.With("provider", providerID)
+	if ad.ExtendedProvider != nil {
+
+		if ad.IsRm {
+			return adIngestError{adIngestIndexerErr, fmt.Errorf("rm ads can not have extended providers")}
+		}
+
+		if len(ad.ContextID) == 0 && ad.ExtendedProvider.Override {
+			return adIngestError{adIngestIndexerErr, fmt.Errorf("override can not be set on extended provider without context id")}
```
gammazero, line 216: Since the ad is not being ingested, it probably makes sense to call `RegisterOrUpdate` _after_ knowing the ad is acceptable. As per the comment above, it also makes sense to register all the information from the ad in one call to the registry.

internal/ingest/linksystem.go
```diff
@@ -631,3 +674,19 @@ func isHAMT(n ipld.Node) bool {
 	h, _ := n.LookupByString("hamt")
 	return h != nil
 }
+
+func stringsToMultiaddrs(addrs []string) []multiaddr.Multiaddr {
```
gammazero, line 678: This can probably be moved to `storetheindex/mautil` if you happen to rebase before you merge.

internal/registry/registry.go
```diff
@@ -30,20 +30,23 @@ import (
...
-	providers map[peer.ID]*ProviderInfo
-	sequences *sequences
+	actions           chan func()
+	closed            chan struct{}
+	closeOnce         sync.Once
+	closing           chan struct{}
+	dstore            datastore.Datastore
+	filterIPs         bool
+	providers         map[peer.ID]*ProviderInfo
+	extendedProviders map[peer.ID]*ExtendedProviders
```
gammazero, line 48: What is the reason for keeping  a separate `extendedProviders` map instead of adding a `**ExtendedProviders` to the `ProviderInfo` type?

internal/registry/registry.go
```diff
@@ -548,6 +651,23 @@ func (r *Registry) ProviderInfo(providerID peer.ID) (*ProviderInfo, bool) {
...
+		}
+		close(infoChan)
+	}
+	extendedProviders := <-infoChan
+	if extendedProviders == nil {
+		return nil
+	}
+	return extendedProviders
+}
+
```
gammazero, line 670: There may be a couple of functions below there need modification:
- I think the `ImportProviders` function needs to be able to import extended providers as well.
- The `RemoveProvider`function should also remove all the associated extended providers.

These concerns avove would automatically be handled if `ExtendedPRoviders` was part of the `ProviderInfo` type

server/finder/handler/finder_handler.go
```diff
@@ -56,40 +57,60 @@ func (h *FinderHandler) Find(mhashes []multihash.Multihash) (*model.FindResponse
...
-					// If provider not in registry, do not return in result.
-					continue
-				}
-				// Omit provider info if not allowed or marked as inactive.
-				if !allowed || pinfo.Inactive() {
-					continue
-				}
-				addrs = pinfo.AddrInfo.Addrs
-				provAddrs[provID] = addrs
+				epRecord = h.registry.ExtendedProviders(provID)
```
gammazero, line 71: If would be more efficient to get the provider addrs and the extended provider addrs from the registry at the same time, otherwise you have to enter the critical goroutine twice for each response.  Same would be true of a critical mutex. This could cause contention which is already a concern with a high read load.

server/finder/handler/finder_handler.go
```diff
@@ -56,40 +57,60 @@ func (h *FinderHandler) Find(mhashes []multihash.Multihash) (*model.FindResponse
...
+			// If override is set to true at the context level then the chain level EPs should be ignored for this context ID
+			override := false
+
+			// Adding context-level EPs if they exist
+			if contextualEpRecord, ok := epRecord.ContextualProviders[string(iVal.ContextID)]; ok {
+				override = contextualEpRecord.Override
+				for _, epInfo := range contextualEpRecord.Providers {
+					provResult, err := createExtendedProviderResult(epInfo, iVal)
+					if err != nil {
+						return nil, err
```
gammazero, line 95: If there is an error creating the `provResutl` I assume that means the data is unusable. Maybe it should be removed?  Also, should the bad data be skipped here to return what good data there is?

server/finder/handler/finder_handler.go
```diff
@@ -56,40 +57,60 @@ func (h *FinderHandler) Find(mhashes []multihash.Multihash) (*model.FindResponse
...
+
+			if override {
+				continue
+			}
+
+			// Adding chain-level EPs if such exist
+			for _, epInfo := range epRecord.Providers {
+				provResult, err := createExtendedProviderResult(epInfo, iVal)
+				if err != nil {
+					return nil, err
```
gammazero, line 109: Same as previous comment

gammazero (commented):
This looks great so far, and I mostly have a number of questions, There may be some more work needed in the registry (see comments).

api/v0/ingest/schema/envelope.go
```diff
@@ -89,8 +119,62 @@ func signaturePayload(ad *Advertisement, oldFormat bool) ([]byte, error) {
...
+	}
+
+	return multihash.Sum(sigBuf.Bytes(), multihash.SHA2_256, -1)
+}
+
 // Sign signs an advertisement using the given private key.
+// This function will return an error if used to sign an ad with extended providers.
 func (ad *Advertisement) Sign(key crypto.PrivKey) error {
+	if ad.ExtendedProvider != nil {
+		return fmt.Errorf("the ad can not be signed because it has extended providers")
```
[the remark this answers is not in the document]
ischasny: Noted, thanks

api/v0/ingest/schema/envelope.go
```diff
@@ -89,8 +119,62 @@ func signaturePayload(ad *Advertisement, oldFormat bool) ([]byte, error) {
...
+		sigBuf.WriteByte(0)
+	}
+
+	return multihash.Sum(sigBuf.Bytes(), multihash.SHA2_256, -1)
+}
+
 // Sign signs an advertisement using the given private key.
+// This function will return an error if used to sign an ad with extended providers.
 func (ad *Advertisement) Sign(key crypto.PrivKey) error {
+	if ad.ExtendedProvider != nil {
```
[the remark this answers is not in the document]
ischasny: To sign an ad with extended providers one will need to have access to all of the other identities' keys (that are scattered across different hosts, so fetching them automatically isn't an option). Creating an ad with extended providers is envisioned to be an infrequent and deliberately manual process that would require human intervention. Having said that - I didn't want to pollute the main API (`Sign`) with extra arguments that won't be required in 99.99% of cases.

api/v0/ingest/schema/envelope.go
```diff
@@ -108,12 +192,68 @@ func (ad *Advertisement) Sign(key crypto.PrivKey) error {
...
 
+// SignWithExtendedProviders signs an advertisement by the main provider as well as by all extended providers if they are present.
+func (ad *Advertisement) SignWithExtendedProviders(key crypto.PrivKey, extendedProviderKeyFetcher func(string) (crypto.PrivKey, error)) error {
+	err := ad.signAd(key)
+	if err != nil {
+		return err
+	}
+
+	if ad.ExtendedProvider == nil {
+		return nil
```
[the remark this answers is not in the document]
ischasny: I did want to make `SignWithExternalProviders` to be able to handle both cases as inside the function it has access to all required keys already.

api/v0/ingest/schema/envelope.go
ischasny: Good point. I will add a logic to deduplicate on the ingestion side. Duplicates aren't prohibited by the spec itself it's more of an optimisation.

api/v0/ingest/schema/schema.ipldsch
ischasny: The indexer's ExtendedProviders list is updated every time when it ingests a new ad with ExtendedProviders. Providers from that list will be returned either for all cids (previous and future ones) if the ad didn't have contextID or only for a specific context id (if the ad had the context ID set). It can also be a union of both but that would require 2 different ads one with contextID one without.

internal/ingest/linksystem.go
ischasny: Good catch, fixed

internal/ingest/linksystem.go
ischasny: Noted, will do

internal/registry/registry.go
ischasny: Done, see previous comment. Marking this one as resolved.

server/finder/handler/finder_handler.go
ischasny: Done as a result of moving `ExtendedProviders` into `ProviderInfo`

server/finder/handler/finder_handler.go
ischasny: Good catch. I revisited `createExtendedProviderResult` and it actually didn't need to return `error` at all (it was always `nil`). Removed the error checking logic here.

server/finder/handler/finder_handler.go
ischasny: Same as above.

internal/registry/registry.go
ischasny: Agreed, that'd be a better solution. Updated the code.

ischasny:
> This looks great so far, and I mostly have a number of questions, There may be some more work needed in the registry (see comments).

Thanks for review @gammazero. I pushed in the changes, please have a look.

internal/registry/registry.go
ischasny: Just one question - is there any specific reason why `ExtendedProviders` has to be a double pointer? Or is this a typo?

gammazero: Typo!!

server/finder/handler/finder_handler.go
gammazero: Awesome! Most efficient code is no code.

gammazero (commented):
All issues from previous review are resolved, looks great.

A couple of other areas that also need modification for extended providers, but does not have to be in this PR:

- This [ProviderInfo data transfer object](https://github.com/filecoin-project/storetheindex/blob/main/api/v0/finder/model/provider_info.go#L12-L18) is used to communicate provider over the external API, for the `/providers/` endpoint as well as [importing providers](https://github.com/filecoin-project/storetheindex/blob/main/server/admin/http/handler.go#L146) via the admin API. This API object needs to be updated.
- The [code for importing providers](https://github.com/filecoin-project/storetheindex/blob/74031627a73f7a8b1a6a02603de65618c22dc558/internal/registry/registry.go#L573) also needs to handle the extended provider data.

ischasny:
> All issues from previous review are resolved, looks great.
> 
> A couple of other areas that also need modification for extended providers, but does not have to be in this PR:
> 
> * This [ProviderInfo data transfer object](https://github.com/filecoin-project/storetheindex/blob/main/api/v0/finder/model/provider_info.go#L12-L18) is used to communicate provider over the external API, for the `/providers/` endpoint as well as [importing providers](https://github.com/filecoin-project/storetheindex/blob/main/server/admin/http/handler.go#L146) via the admin API. This API object needs to be updated.
> * The [code for importing providers](https://github.com/filecoin-project/storetheindex/blob/74031627a73f7a8b1a6a02603de65618c22dc558/internal/registry/registry.go#L573) also needs to handle the extended provider data.

Cool :) Thanks for pointing out. I'll probably add that with a follow up PR then as size of this one gets beyond reviewable.

gammazero (approved):

api/v0/ingest/schema/schema.ipldsch
ischasny: Will add a link to the spec. Otherwise I'll have to copy and paste it's contents.

PR #1004, opened 2022-11-08
Fork go-legs into storetheindex as dagsync package

gammazero:
Storetheindex no longer needs to depend on go-legs. All code from go-legs has been put under the `dagsync` directory/package.

This will allow the go-legs repo to be archived.


willscott (approved):
do we need to archive go-legs? we might simplify it back to it's original 1:1 sync, but i think a place to have general dag synchronization that's less complex than our transports directly is still useful.

PR #1006, opened 2022-11-09
Change default value store to pebble

gammazero:
- Make pebble the default index store
- Remove unused flags
- Reuse flags needed for multiple commands
- Fix admin server when p2p is disabled
- Cancel datatransfer if it takes too long to stop
- Stabilize tests

### DIFF
--- a/config/indexer.go
+++ b/config/indexer.go
@@ -69,7 +69,7 @@ func NewIndexer() Indexer {
 		GCTimeLimit:         Duration(5 * time.Minute),
 		ShutdownTimeout:     0,
 		ValueStoreDir:       "valuestore",
-		ValueStoreType:      "sth",
+		ValueStoreType:      "pebble",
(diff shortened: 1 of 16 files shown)


config/indexer.go
```diff
@@ -69,7 +69,7 @@ func NewIndexer() Indexer {
 		GCTimeLimit:         Duration(5 * time.Minute),
 		ShutdownTimeout:     0,
 		ValueStoreDir:       "valuestore",
-		ValueStoreType:      "sth",
+		ValueStoreType:      "pebble",
```
masih, line 72: This is probably the most significant change the PR.

I recommend rewording the title PR to include it for the benefit of someone looking through commit history and reading titles of it, as well as making the release notes point out what's important when autogenerated from PR merges.

masih (approved):

PR #1016, opened 2022-11-15
Replace `indexer-1` on `dev` with a new node

masih:
Replace `indexer-1` on `dev` environment with a new node called `ber` that uses the `dido` snapshot from `prod` as its initial state.

The replacing node uses the same identity as `indexer-1`, since it is whitelisted on lotus bootrstrap nodes and will receive relayed gossip. As a result, `mya` is configured to now pair with `ber` instead of `indexer-1`.


ischasny (approved):
LGTM

PR #1017, opened 2022-11-15
Remove remaining resources associated with `indexer-1` on `dev`

masih:
Remove the encryped key of `indexer-1` now that it is reused by `ber` and replace it with `ber` in indexstar backend.


ischasny (approved):
Lgtm

PR #1022, opened 2022-11-15
Assigner service

gammazero:
Assigner service was previously called "Coordination service".

The Assigner service is a service implementation that is built and runs as a separate executable from storetheindex. The Assigner receives announce messages via pubsub and HTTP, and assigns the publishers described in these messages to the indexers configured as part of the Assigner's indexer pool.

The assigner queries allowed peers on the indexer to see what publishers are assigned.

- Assigner requires correct indexer policy configuration, defauly block, that has explicit allow list.
- Update policy evaluation in registry, since policy pertains to publisher, not provider.
- New admin command to list explicitly allowed peers.
- Mew admin endpoint to get explicitly allowed peers.
- Removed unused `Registry.Register`, only `Registry.Update` is used now.
- Fixed error in registry when updating publisher addresses (would overwrite with provider address)
- Indexer updates config file when new peer allowed.

TODO in this PR:
- :heavy_check_mark: storetheindex reports assigned publishers immediately after assignment
- :heavy_check_mark: test to verify assigned publisher without sync

TODO in separate PRs (not actually missing from this PR):
- assigner service deployment
- storetheindex needs to enter freeze mode when storage near capacity.
- storetheindex needs to report when frozen
- assigner needs to detect frozen indexer and to handoff
- increase test coverage


masih:
On `TODO`s listed in PR description. Are they to be done in this PR or separately?

My vote is to move them to a separate PR since this one is large enough already?

masih:
Hey Andrew, Here’s a quick braindump on steps to take in order to get the `Assigner` service deployed:

**Context**
* The service is stateless therefore wont need any PVC definitions.
* There should only be a singleton instance of it running to avoid race conditions in assigning responsibilities.
* The service is only internal and need not to be exposed publicly.
* I am assuming you envision building a dedicated Docker image for `assigner`.
  * an alternative here is to add it as a sub command to `storetheindex`, e.g. `storetheindex assigner` similar to `storetheindex daemon` which will simplify the steps by removing the need for creating ECR repo and CI job, since the assigner is highly coupled to the internal admin API of `storetheindex`. Entirely up to you.

**Infra**
The assigner service container needs to be built and pushed to a container registry for K8S to then pull from. We also need to add CI jobs to build and push the image just like we do for storetheindex itself. With that in mind:
* Create a new ECR repo by adding a line [here](https://github.com/filecoin-project/storetheindex/blob/3de220f896597e4aaa496c32731fed68aca0ee1a/deploy/infrastructure/common/ecr.tf#L5), i.e. `storetheindex/assigner`
* Apply the terraform changes by following the steps in `RunBook/How To/Apply changes to Infrastructure manifests`
* Once that’s done, you should be able to see the ECR private repo created in `us-east-2` via AWS Console.
* Define a `Dockerfile` under the `assigner` package.
* Define a CI job that builds and pushes the image, essentially a copy of the [existing job](https://github.com/filecoin-project/storetheindex/blob/main/.github/workflows/ecr-publisher.yml) with [image name](https://github.com/filecoin-project/storetheindex/blob/main/.github/workflows/ecr-publisher.yml#L49) and path to `Dockerfile` in [`build` command ](https://github.com/filecoin-project/storetheindex/blob/main/.github/workflows/ecr-publisher.yml#L50) adjusted.

**Manifests**
The K8S manifests needed to run it would be very similar to the manifests defined for `indexstar` with two differences: 1) We want to only run one instance of the assigner service only, and 2) `assigner` should not be publicly accessible. We would then need to:
* Define a `base` Kustomisation, similar to [`indexstar`](https://github.com/filecoin-project/storetheindex/tree/4c2135b69c3d5101fdb83bf5883e334172d52a24/deploy/manifests/base/indexstar) under `deploy/manifests/base/storetheindex-assigner` directory that contains:
  * `Service` manifest to have a DNS name for the assigner service
  * `Deployment` manifest to run the pod that runs the container
    * Make sure the replica cont is set to 1 like [this](https://github.com/filecoin-project/storetheindex/blob/9cc1d5686bfde91d1a48b371aed0b237f3a787fe/deploy/manifests/base/storetheindex-single/deployment.yaml#L6-L11).
  * `Kustomization` to bring it all together.

  Unlike `indexstar` we won't need `PodDisruptionBudget` since we want to only run a single instance and for one instance PDB doesn't make sense.
* Define an overlay initially in `dev` only for the assigner, again similar to [`indexstar`](https://github.com/filecoin-project/storetheindex/tree/86b8c93a760fb7b1868e4e02cb5870a2286932d2/deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar) under `deploy/manifests/dev/us-east-2/tenant/storetheindex/assigner`
  * I recommend overriding resources like [this](https://github.com/filecoin-project/storetheindex/blob/86b8c93a760fb7b1868e4e02cb5870a2286932d2/deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/deployment.yaml#L20-L26) in the overlay as needed.
  * Because the `assigner` service shouldn't be publically accessible, we should not define any `Ingress` objects.
  * If it does not expose any Prometheus metrics, skip adding `PodMonitor` to the overlay too. Once there are metrics we can add that to the overlay.
* Add `assigner` [here](https://github.com/filecoin-project/storetheindex/blob/2cd561f2bed54efe2a402c71af4542fb0cf9af28/deploy/manifests/dev/us-east-2/tenant/storetheindex/kustomization.yaml#L6) so that it gets added to the collection of manifests applied by the CD pipeline.

That's it. I hope this is helpful. Happy to jump on a quick call if you have any questions.

command/init.go
```diff
@@ -77,31 +77,48 @@ func initCommand(cctx *cli.Context) error {
...
+			_, err := multiaddr.NewMultiaddr(ingestAddr)
+			if err != nil {
+				return fmt.Errorf("bad listen-ingest: %s", err)
+			}
 		}
 		cfg.Addresses.Ingest = ingestAddr
 	}
 
+	p2pAddr := cctx.String("listen-p2p")
+	if p2pAddr != "" {
```
willscott, line 112: should this check be `if cctx.IsSet("listen-p2p")`?

willscott (approved):
this seems like a reasonable new sub-component

assigner/core/assigner.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/filecoin-project/storetheindex/peerutil"
+	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+var log = logging.Logger("assigner/core")
+
+// Core contains the core data and logic for the assigner.
```
masih, line 23: This reads as package doc; move to `doc.go` or to the top of `package`?

assigner/core/assigner.go
```diff
@@ -0,0 +1,364 @@
...
+
+	/*
+		log.Infow("Starting sync on indexer", indexer.adminURL)
+		err = cl.Sync(ctx, amsg.PeerID, amsg.Addrs[0], 0, false)
+		if err != nil {
+			// Do not consider this a failure to assign since allowing the
+			// publisher effectively assigns it.
+			log.Errorw("Error starting sync to new assigned publisher", "err", err, "publisher", amsg.PeerID)
+		}
+	*/
```
masih, line 352: Needed?

command/init.go
gammazero: No, because doing:
```shell
./storetheindex init --listen-p2p=""
```
results in `cctx.IsSet` returning `true`, but the string being empty `""`

assigner/server/server_test.go
```diff
@@ -0,0 +1,294 @@
...
+			Except: []string{trustedID},
+		},
+		PubSubTopic: pubsubTopic,
+	}
+	assigner, err := core.NewAssigner(context.Background(), cfg, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(5 * time.Second)
```
masih, line 216: Can we avoid this using `require.Eventually`?

assigner/core/assigner.go
```diff
@@ -0,0 +1,364 @@
...
+	cl, err := adminclient.New(indexer.adminURL)
+	if err != nil {
+		return err
+	}
+	err = cl.Allow(ctx, amsg.PeerID)
+	if err != nil {
+		return err
+	}
+
+	// Semd announce instead of sync request in case indexer is already syncing
```
masih, line 328:
```suggestion
	// Send announce instead of sync request in case indexer is already syncing
```

assigner/core/assigner.go
```diff
@@ -0,0 +1,364 @@
...
+	// due to receiving announce after immediately allowing the publlisher,
```
masih, line 329:
```suggestion
	// due to receiving announce after immediately allowing the publisher,
```

assigner/core/assigner.go
```diff
@@ -0,0 +1,364 @@
...
+	noticeChan <- a.indexerPool[indexerNum].adminURL
+	close(noticeChan) // signal not more data on channel
+
+	delete(a.waitingNotice, pubID)
+	if len(a.waitingNotice) == 0 {
+		a.waitingNotice = nil
+	}
+}
+
+// watch fetches announce messages from the Reciever.
```
masih, line 211:
```suggestion
// watch fetches announce messages from the Receiver.
```

masih (approved):

api/v0/admin/client/http/client.go
```diff
@@ -189,6 +189,39 @@ func (c *Client) Allow(ctx context.Context, peerID peer.ID) error {
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, httpclient.ReadErrorFrom(resp.StatusCode, resp.Body)
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
gammazero: I believe that is already done 6 lines above on line 205.

api/v0/admin/client/http/client.go
```diff
@@ -189,6 +189,39 @@ func (c *Client) Allow(ctx context.Context, peerID peer.ID) error {
 	return c.ingestRequest(ctx, peerID, "allow", http.MethodPut, nil)
 }
 
+// AllowList gets a list of explicitly allowed peers, if policy is configured
+// to have allow list.
+func (c *Client) AllowList(ctx context.Context) ([]peer.ID, error) {
```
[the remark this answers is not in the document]
gammazero: Definitely! Done.

Also renamed admin cli command to match.

assigner/core/assigner.go
gammazero: no

assigner/server/server_test.go
gammazero: I don't think this is needed at all. It was copied from the e2e test.

api/v0/admin/client/http/client.go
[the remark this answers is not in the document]
masih: That's closing the body. Read to completion doesn't happen if status is not a success right?

PR #1032, opened 2022-11-17
Remove `indexer-0` and `1` from `prod` and enable CD

masih:
Remove redundant manifests for the retired nodes: `indexer-0` and `indexer-1`, along with their ingress point: `prod.cid.contact`.

Reflect the changes in cloudfront configuration and use consistent naming for origins across dev and prod.

Note, CD is not enabled for `kepa` until graceful rollout of updates across multiple nodes is figured out. Otherwise, we will experience downtime on deployment.


ischasny:
> CD is not enabled for kepa until graceful rollout of updates across multiple nodes is figured out

As a temporary solution, maybe we can create two automated PRs for deploying to prod ? You'd merge one, wait until rollout is over and then merge the other one.

masih:
We could indeed have a dedicated PR creation per node. but that gets tedious very quickly. Let's discuss further in Colo.

ischasny (approved):
LGTM

PR #1034, opened 2022-11-17
Save assignments from assigner service in datastore

gammazero:
When the indexer is configured to work with an assigner service, the indexer persists assigned peers in the datastore.

Peers listed in `Discovery.Policy.Except`, in the configuration file, are included in the set of assigned peers. This allows an indexer to be pre-configured with assigned peers.

- Admin command `list-assigned` shows assigned peers configured to use assigner service
- The `/assigned/` admin endpoint gets assigned peers configured to use assigner service
- Admin `block` command can be used to administratively unassign a peer
- New init flag `use-assigner` to create configuration file with settings to use assigner service
- Test to check that assignment is persisted

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -29,7 +29,8 @@ import (
 
 const (
 	// providerKeyPath is where provider info is stored in to indexer repo.
-	providerKeyPath = "/registry/pinfo"
+	providerKeyPath    = "/registry/pinfo"
+	assignmentsKeyPath = "/assignments"
 )
 
 var log = logging.Logger("indexer/registry")
@@ -49,6 +50,7 @@ type Registry struct {
 	discoverWait  sync.WaitGroup
 	discoverTimes map[string]time.Time
 	policy        *policy.Policy
+	useAssigner   bool
 
 	discoveryTimeout time.Duration
 	rediscoverWait   time.Duration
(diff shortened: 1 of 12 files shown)


internal/registry/registry.go
```diff
@@ -432,35 +442,56 @@ func (r *Registry) SetPolicy(policyCfg config.Policy) error {
...
 
-// AllowList returns list of explicitly allowed peer IDs, or false if policy
-// does not have allow list because it allows by default.
-func (r *Registry) AllowList() ([]peer.ID, bool) {
-	return r.policy.AllowList()
+// ListAssignedPeers returns list of allowed peer IDs, when the indexer is
+// configured to work with an assigner service. Otherwise returns error.
+func (r *Registry) ListAssignedPeers() ([]peer.ID, error) {
+	if r.useAssigner {
+		peers, ok := r.policy.ListAllowedPeers()
```
masih, line 489: Thinking out loud:
What is the rationale for introducing the term "ListAssignedPeers" were it is basically identical to "ListAllowedPeers"?

Looking at the admin API holistically, it seems a bit odd that we have allow, block, list allowed and a separate list assigned.

For simplicity, i wonder if we should stick to either one term or the other. 

If the rationale is to make it clear that "hey this node is being managed by assigner" then a simple admin API that returns the value of `useAssigner` say via `/admin/status` would maybe offer a more robust approach where an assigner (and indeed any process with access to admin API) can check if a node is being managed by an assigner or not and program behaviour accordingly.

If any of this makes sense please feel free to leave TODOs for future PRs. 
What's here seems perfectly functional.

masih (approved):

internal/registry/registry.go
gammazero: It is a rename of `ListAllowedPeers` which was only created for this purpose in a previous PR. This rename makes it clear that it is only used when working with an assigner service.

When not working with an assigner service, this is useless, because the config holds this information. So, we will only have a single `ListAssignedPeers` with corresponding API endpoint.

PR #1046, opened 2022-11-19
Use EP metadata for the main provider

ischasny:
If metadata of the main provider's EP record is different from the one in the ad, then such records would be returned in the find resultset as separate entries. This feature is required if the the main provider would like to advertise some new endpoints for a different protocol.

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -226,17 +226,11 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
 
 		// Creating ExtendedProviderInfo record for each of the providers.
 		// Keeping in mind that the provider from the outer ad doesn't need to be included into the extended providers list
-		eProvs := make([]registry.ExtendedProviderInfo, 0, len(ad.ExtendedProvider.Providers)-1)
-		seenEProvs := map[peer.ID]struct{}{}
+		eProvs := make([]registry.ExtendedProviderInfo, 0, len(ad.ExtendedProvider.Providers))
 		for _, ep := range ad.ExtendedProvider.Providers {
 			epID, err := peer.Decode(ep.ID)
 			if err != nil {
 				return adIngestError{adIngestRegisterProviderErr, fmt.Errorf("could not register/update extended provider info: %w", err)}
-			}
-
-			// Skipping the record that is for provider form the main ad or if the provider has already been seen
-			if _, ok := seenEProvs[epID]; ok || ep.ID == ad.Provider {
-				continue
 			}
 
 			eProvs = append(eProvs, registry.ExtendedProviderInfo{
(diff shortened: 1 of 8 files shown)


server/finder/handler/finder_handler.go
```diff
@@ -57,45 +57,56 @@ func (h *FinderHandler) Find(mhashes []multihash.Multihash) (*model.FindResponse
...
+				for _, epInfo := range epRecord.Providers {
+					if epInfo.PeerID != mainProvID {
+						eProvResult := createExtendedProviderResult(epInfo, iVal)
+						provResults = append(provResults, *eProvResult)
+						continue
+					}
+
+					// updating main provider's metadata if it's not empty.
+					if len(epInfo.Metadata) > 0 {
+						(&provResults[0]).Metadata = epInfo.Metadata
```
ischasny, line 90: Addresses aren't included because they should match the ones specified in the AD's body and hence will be updated in the registry.

willscott:
I don't think we want it to over-ride, we want to return 2 results, one with the local value, and one with the chain-level EP one.

In the specific case we're thinking about here, there's transport metadata overall for the provider we always want to be returned that says 'this ID speaks bitswap' there's also a per-advertisement value that's the normal (non EP) metadata in each advertisement that's saying 'this ID speaks graphsync, when you make the graphsync request you should reference this deal'.

ischasny:
Just to confirm my understanding - indexer should return one record for the main provider if metadata and addrs of its EP record match the ones of the main ad. If either metadata or addrs of the main provider's EP record are different from the main AD, then the indexer would return 2 separate records. Is that correct?

internal/ingest/linksystem.go
```diff
@@ -226,19 +226,13 @@ func (ing *Ingester) ingestAd(publisherID peer.ID, adCid cid.Cid, ad schema.Adve
...
-		eProvs := make([]registry.ExtendedProviderInfo, 0, len(ad.ExtendedProvider.Providers)-1)
-		seenEProvs := map[peer.ID]struct{}{}
+		eProvs := make([]registry.ExtendedProviderInfo, 0, len(ad.ExtendedProvider.Providers))
 		for _, ep := range ad.ExtendedProvider.Providers {
 			epID, err := peer.Decode(ep.ID)
 			if err != nil {
 				return adIngestError{adIngestRegisterProviderErr, fmt.Errorf("could not register/update extended provider info: %w", err)}
 			}
 
-			// Skipping the record that is for provider form the main ad or if the provider has already been seen
```
ischasny, line 237: Main provider's EP records used to be skipped before during ingestion.

willscott (approved):
if the direct advertisement has EP and [`Override`](https://github.com/filecoin-project/storetheindex/blob/main/doc/ingest.md?plain=1#L99) set, then maybe we skip the chain-level EP info. that said, this seems like it has the correct semantics from the test, so happy to start here

ischasny:
@willscott updated as per our earlier discussion. Now there *could* be multiple records returned for the main provider: one for the main ad, one from the chain level EPs and one from the context level EPs (if such has been provided of course). 

I also changed it to not include duplicate EP records whos metadata matches the main ad.

PR #1047, opened 2022-11-19
Allow specification of preset publisher assignments

### LINKED ISSUE #1045: Configure pre-assigned indexers in assigner service config
The assigner service should allow indexer assignments to be configured.  Each indexer in the pool should be able to have an optional list of publisher IDs that should get assigned to that indexer.

gammazero:
Each indexer in the pool can have preset publisher ID assignments. This means that when a publisher ID is seen, that publisher is assigned to all indexers that have that ID as a preset assignment. Publishers with preset assignments are assigned to all preset indexers without regard to other existing assignments or to the replication value.

This PR also changes the `OnAssignment` notification to report the indexer number instead of the indexer's admin URL. Also, the `OnAssignment` notification will keep reporting assignments of the specified peer ID until no more assignments of that peer are needed, after which the channel is closed.

Fixes #1045

### DIFF
--- a/assigner/core/assigner.go
+++ b/assigner/core/assigner.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"sort"
 	"sync"
 
 	"github.com/filecoin-project/storetheindex/announce"
@@ -31,6 +32,8 @@ type Assigner struct {
 	p2pHost host.Host
 	// policy decides what publisher to accept announce messages from.
 	policy peerutil.Policy
+	// presets maps publisher ID to pre-assigned indexers.
+	presets map[peer.ID][]int
 	// receiver receives announce messages.
 	receiver *announce.Receiver
 	// replication is the number of indexers to assign a publisher to.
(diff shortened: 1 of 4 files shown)


assigner/core/assigner.go
```diff
@@ -31,22 +32,39 @@ type Assigner struct {
...
 	processing bool
 }
 
+func (asmt *assignment) addIndexer(x int) {
+	i := sort.SearchInts(asmt.indexers, x)
+	if i < len(asmt.indexers) && asmt.indexers[i] == x {
+		return
+	}
+	asmt.indexers = append(asmt.indexers, 0)
+	copy(asmt.indexers[i+1:], asmt.indexers[i:])
```
willscott, line 59: it is non-obvious to me that this is safe, and that it doesn't cause the value at i to be 'smeared' over the rest of the array.

willscott (approved):

assigner/core/assigner.go
gammazero: This is a typical insert into a slice:
https://gist.github.com/GimmyHchs/33bd06e68d72a913a8587b09d41b50d0#insert

PR #1069, opened 2022-11-23
Switch static web UI url to the new repo at IPNI GitHub Org

masih:
`CNAME`s are unique across all GitHub Pages repos. The current one is in use by repo under `willscott` and changing it requires admin access to the repo.

For faster turnaround, switch this in code. This also avoids breaking other DNS set up under `willscott.github.io`.


ischasny (approved):
lgtm

PR #1073, opened 2022-11-24
Add `heyfil` deployment manifests to VCS

masih:
Add the deployment manifests belonging to `heyfil` to `prod` deploy folder of storetheindex so that they are version controlled.

The structure to be revisited on move to IPNI org.


ischasny (approved):
lgtm

Issue #982, opened 2022-10-31
Revise CID count estimation


masih:
The current estimate of number of CIDs ingested by storetheindex is backing store dependant. It is made up of the size of index [divided by a constant ](https://github.com/filecoin-project/storetheindex/blob/main/server/finder/handler/finder_handler.go#L140) as a reasonable estimate of a single CID size. This approach can result in high discrepancy if applied to nodes that use different backing stores which take advantage of things like compression.

Retiring old nodes with larger index size will cause the CID count estimate to fluctuate, which may be confusing for the users. At the same time, iterating over the existing indices just to make an accurate count of the number of CIDs ingested seems cumbersome with little reward. Problem remains that we do need to make a better estimate of the number of CIDs ingested reported on cid.contact.

Ways forward:
1. take the current number as the offset, and add a counter to the nodes to report count going forward. Then report the number of CIDs as offset plus current accurate count.
   * Doesn't require much engineering and won't cause dramatic fluctuation in the current value reported but will result in a less accurate estimate of total number of indexed records.
2. Take a snapshot of the most up-to-date node state, attach it on a counter implementation that iterates over the records and returns an accurate count. At the same time implement a counter that starts counting going forward. Once the counter on snapshot finishes, add the resulting value to the new count and we then should have an accurate number.
   * This will give us an accurate number of indexed records but will come with potential fluctuation (most likely decrease) of the value reported today.

My current thought is to proceed with  option No. 1 considering the other work at hand.

Cc @honghaoq

honghaoq:
Thanks Masih, I like #1, it is useful to have a more consistent (less jumpy) overall CID count estimation, but probably not worth it to spend a ton of engineering effort just to get a super accurate count.

honghaoq:
@masih do we end up agreeing to go with option 1 here, is there an ETA for the update? From our call last time I think in Q4 - End of Dec?

masih:
Based on further discussion we are going to add a delta calculation metric that reports the number of currently indexed hashes instead of an absolute "total ingested". That is a more useful metric that would go up and down as indices are added and removed

LaurenSpiegel:
Would it be possible to do option 1 plus a low lift accuracy adjustment to the current count? Basically adjust the constant for each indexer by what we see the actual size increase for new CID's added to that indexer. So, if we see from the new counter that 100 CID's are added and storage space increased by X, we average that on that indexer, each index takes up X/100. Feel free to tell me this is a bad idea.

[...]

PR #1074, opened 2022-11-24
Refined total CIDs indexed estimation

masih:
Integrate `/stats` API with the new `indexer.Stats` values when the backing store supports it. This allows the API to return a more accurate estimation of the total CIDs indexed.

The logic is cached and refreshed hourly. This is because, such estimation may take time and can be expensive. Using caching we can avoid stressing the backing store if `/stats` is called in short succession.

Relates to #982


server/finder/handler/finder_handler.go
```diff
@@ -247,16 +250,19 @@ func (h *FinderHandler) GetProvider(providerID peer.ID) ([]byte, error) {
...
 		return nil, err
+	case stats.EntriesEstimate == 0:
+		return nil, nil
+	default:
+		return model.MarshalStats(&stats)
 	}
+}
 
-	s := model.Stats{
-		EntriesEstimate: size / avg_mh_size,
```
willscott, line 256: should we add this as a new field - so we keep the current `EntriesEstimate` but also add a `EntriesCount` with the more precise stats from the engine?

masih: Both are estimates. The existing one is very off and backing store-dependant.
I am curious what value would that deliver?

willscott: it would allow a period where we could report both metrics and say what the correction factor we're applying is in switching to the new one.

gammazero:
One way to get a more accurate estimate of the number of indexes stored in a given amount of storage, is to look at the change in storage size compared to the change in the [actual number of indexes](https://github.com/filecoin-project/storetheindex/blob/main/internal/metrics/server.go#L38) stored. As these numbers get larger, the approximation of indexes per storage gets more accurate.

server/finder/handler/finder_handler.go
masih: That's fine; i'll add that as a separate commit so that we can revert it easily when we do the comparison study.

masih:
measurements based on size growth are fundamentally flawed since they entirely depend on the store implementation.  At "storetheindex" level of abstraction drawing such conclusions based on size of the index is bound to be unprodictibly inaccurate.

Instead we should delegate such stats to the store to provide. This is what [this](https://github.com/filecoin-project/go-indexer-core/pull/115) PR introduced. In the case of Pebble, I think getting the count from SST file number of entries would be pretty accurate.

server/finder/handler/finder_handler.go
masih: See: https://github.com/filecoin-project/storetheindex/pull/1074/commits/9ec2de45935dc01a68a09eaea2022cbde9e5f50c

willscott (approved):

PR #1084, opened 2022-11-30
Provide items for Dockerfile ENTRYPOINT and CMD in deployment.yaml

gammazero:
Since `command` is used in deployment.yaml, this overrides both `ENTRYPOINT` and `CMD` in Dockerfile.

### DIFF
--- a/deploy/manifests/base/assigner/deployment.yaml
+++ b/deploy/manifests/base/assigner/deployment.yaml
@@ -35,7 +35,6 @@ spec:
               name: http
             - containerPort: 8081
               name: metrics
-          volumeMounts:
           readinessProbe:
             httpGet:
               port: http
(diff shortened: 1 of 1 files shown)


masih:
Do not merge this; verifying.

deploy/manifests/base/assigner/deployment.yaml
```diff
@@ -20,6 +20,7 @@ spec:
         - name: assigner
           image: storetheindex
           command: ["/sbin/tini", "--", "/usr/local/bin/start_assigner"]
+          args: ["daemon"]
```
masih, line 23: Though valid YAML, why use JSON syntax to specify arrays when one can use `-` which is consistent with how arrays are specified across the manifest?

gammazero: fixed

masih (approved):

PR #1085, opened 2022-11-30
Configure dev indexers to use assigner service

gammazero:
- Use latest image for dev indexers
- Configure ber and cali to use assigner service


masih (commented):
👍

PR #1086, opened 2022-12-01
Fix empty providers listing by `indexstar`

masih:
Deploy latest to fix empty providers listing by `/providers` in indexstar.


ischasny (approved):
lgtm

PR #1109, opened 2022-12-07
Configure Cloudfront to route announce messages to assigner node

gammazero:


gammazero:
@masih The assigner node is using the identity from the old indexer-0 node. Does this need to be swapped with the identity of the kepa node?

masih:
It doesn't. Kepa is using the identity of indexer-1.

masih (approved):

PR #1111, opened 2022-12-07
Create configuration to create indexer snapshots

### LINKED ISSUE #1105: Snapshot `dido` and `oden` before enabling assigner on prod
The last snapshot taken was from `dido` on 2nd of Nov. 22. As a backup, take snapshot of both `dido` and `oden`. 

Note that `oden` is only whitelisted to nft.storage provider. 

For snapshotting itself, this is managed by K8S controllers and is done by defining relevant YAML manifests. See existing example [here](https://github.com/ipni/storetheindex/blob/main/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/kepa/dido-snapshot.yaml). 

Please feel free to add snapshot manifests in a dedicated subdirectory with a dedicated kustomization file.

gammazero:
Fixes #1105


masih (changes requested):
Looks great. We just need a fresh manifest for dido and shouldn't modify or move the existing manifest for that node.

masih (approved):

PR #1116, opened 2022-12-09
Assigner waits to get existing assignments from indexers

### LINKED ISSUE #1118: Do not assign publishers when pool assignments are unknown
If an indexer is offline when the assigner service starts, the assigner cannot read the existing assignments or preferences from that indexer. The assigner will not know what publishers are assigned to that indexer, resulting in assigning those publishers to other indexers as announcements from them arrive. That can result in over-assigning publishers to the indexer pool.

This is further complicated by the assigner not reading preferred assignments (peers that an indexer has previously stored content for but are not assigned to it) when the indexer comes back online. When another indexer has those same preferences, all the assignments are given to the indexer whose preferences are know, even though both indexers are now online.

This is particularly problematic when indexers are newly configured to use an assigner, because most of the assignments happen in a short amount of time as new announce messages arrive. This is what happened on the prod indexers.

To fix this, two things are needed:
- Reset the assignments on the indexers
- Update the assigner to not assign publishers when pool assignments are unknown.

gammazero:
## Context
If an indexer is offline when the assigner service starts, the assigner cannot read the existing assignments or preferences from that indexer. The assigner will not know what publishers are assigned to that indexer, resulting in assigning those publishers to other indexers as announcements from them arrive.

The assigner service needs to retry getting indexer assignments before assigning a publisher to an indexer, since that publisher may already be assigned to an offline indexer. This prevents over-assigning publishers to the indexer pool.

Fixes #1118 

## Proposed Changes
If the assigner starts without getting assignment information for all indexers in the pool, then it will recheck each time an announce message is received from an unassigned peer. If there are still indexers offline, the assigner assumes that the peer may be assigned to an offline indexer and reduces the required number of assignments by the number of indexers offline.

This may mean that an unassigned peer will not get assigned until indexer(s) come back online. It is necessary to prevent over-assigning publishers to the indexer pool.

## Tests
Tests include starting with an indexer offline and testing that:
- Assignment information is initially unread
- Publisher is not assigned when indexers assignments are unknown
- After indexer is back online, all assignment info is all read when unassigned publisher seen
 
## Revert Strategy
Reverting will return to the previous situation and may require resetting indexer assignments.

### DIFF
--- a/assigner/core/assigner.go
+++ b/assigner/core/assigner.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
@@ -28,6 +29,8 @@ type Assigner struct {
 	assigned map[peer.ID]*assignment
 	// indexerPool is the set of indexers to assign publishers to.
 	indexerPool []indexerInfo
+	// initDone is true when assignments have been read from all indexers.
+	initDone bool
 	// mutex protects assigned.
 	mutex   sync.Mutex
 	p2pHost host.Host
@@ -78,6 +81,7 @@ type indexerInfo struct {
 	adminURL  string
 	ingestURL string
 	assigned  int32
+	initDone  bool
 }
 
 // assignedCount returns the number of publishers assigned to this indexer.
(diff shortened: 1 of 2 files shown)


masih (approved):
Should we test this on dev first?

PR #1132, opened 2023-01-03
Indexer frozen mode and handoff

gammazero:
## Context
Implementation of indexer freeze mode and assigner coordination for handoff of assigned publishers to other indexers.

### Indexer Freeze

The indexer monitors the remaining storage capacity of the filesystem that the value store is kept on. When the percent of space used reaches the configured threshold, the indexer enters "frozen" mode where it does not store any more index content. It still continues to process ads and handles removal ads. The indexer records the last ad ingested before becoming frozen, and makes this available as part of the providers information.

### Indexer Handles Handoff when Assigner Tells it to

The assigner service tells an indexer to take over ingesting ads from a publisher that was assigned to a frozen indexer. The indexer handles receives the publisher being handed off and the ID and address of the frozen indexer previously handling the publisher. The indexer then queries the frozen indexer for provider records, finds the provider that has the handed-off publisher, sets the last processed ad to the ad that the provider was frozen at on the frozen indexer, and syncs with handed-off publisher. The sync only goes as deep as the last advertisement the frozen indexer processed before entering frozen mode.

### Assigner Service Detects Frozen Indexer and Initiates Handoff

The Assigner polls indexers to detect when they have become frozen. On finding a frozen indexer the Assigner selects a candidate indexer for each assigned publisher. The Assigner then tells the candidate indexer to do handoff giving it the publisher and the frozen indexer info. If handoff is successful, the Assigner changes the assignment from the frozen indexer to the new indexer that the publisher was handed off to, and then the Assigner does the same for the next publisher to handoff. If handoff is incomplete, the Assigner will attempt to handoff the remaining publishers at the next poll.

### Indexer can be unfrozen
The indexer can be configured to revert back to unfrozen mode at startup. This reverts the latest processed advertisements to what they where when the indexer was frozen, only preserving the any updates to the publisher and provider addresses.

## Tests
- Indexer freezes when disk usage reaches configured threshold.
- Indexer handles handoff of publisher and resumes indexing provider from frozen ad.
- Assigner service detects frozen indexers and reassigns publishers from frozen indexer
- Test handoff of publisher that is no longer in use on indexer
- Test status API
- Test that /providers API returns frozen information and optional extended provider info.
- Test conversion between internal ProviderInfo and API ProviderInfo
- New e2e tests for `admin freeze`, `status`, and `providers` commands.
 - Test registry unfreeze
 - 
# Command changes
- New `admin status` command that reports indexer ID, frozen status, and percent usage.
- New `admin freeze` command to immediately freeze indexer.
- The `providers` command shows frozen info if indexer is frozen.

## Revert Strategy

- Revert indexer(s) first while Assigner is still running.
- After indexers reverted and running, revert Assigner service. 

## Note to reviewers

The review can be broken in separate parts, in line with the sections under Proposed Changes. If preferable, this PR can also be broken into a series of smaller PRs, but without the ability to have an end-to-end test until those PRs are merged.


api/v0/finder/model/provider_info.go
```diff
@@ -17,36 +14,19 @@ type ProviderInfo struct {
 	Publisher             *peer.AddrInfo     `json:",omitempty"`
 	IndexCount            uint64             `json:",omitempty"`
 	ExtendedProviders     *ExtendedProviders `json:",omitempty"`
+	FrozenAt              cid.Cid            `json:",omitempty"`
+	FrozenAtTime          string             `json:",omitempty"`
```
masih, line 18: This info is going to end up at the public facing `/providers` endpoint correct? 

I am curious what is the rationale for exposing these publicly?

config/indexer.go
```diff
@@ -16,6 +16,10 @@ type Indexer struct {
 	// individual multihashes within a Put. A value of 1 means no concurrency,
 	// and zero uses the default.
 	CorePutConcurrency int
+	// FreezeAtPercent is the percent used, of the file system that
+	// ValueStoreDir is on, at which to trigger the indexer to enter frozen
+	// mode. A zero value uses the default. A negative value disables freezing.
+	FreezeAtPercent float64
```
masih, line 22: Should there also be a grace period after which the publishers assigned to a frozen node are reassigned? 

The idea is to give a window to an admininstrator to potentially expand disk and avoid the cost of reassignment. 

In the case of cid.contact deployment we should almost never need to use the freeze feature considering the current disk growth rate, the alert at 20% utilisation which is monitored at the infra level, and the on call rota with runbook entries to expand disk.

masih (approved):

api/v0/finder/model/provider_info.go
gammazero: The reason for making this public is so that people know why that particular indexer does not return results for queries for content beyond the `FrozenAt` advertisement. For general provider information, as presented by indexstar, this would be filtered out.

config/indexer.go
gammazero: That could be something configurable in the assigner service, but it seems unnecessary. Each indexer logs its disk usage at every check
```
Disk usage OK   {"usage": "0.13%", "freezeAt": "90%"}
```
When it is within 10% of reaching the freeze threshold, a warning is logged:
```
Disk usage alert     {"usage": "81.77%", "freezeAt": "90%"}
```
When it is within 2% of reaching the freeze threshold, a critical warning is logged:
```
Disk usage critical  {"usage": "88.45%", "freezeAt": "90%"}
```
The usage checks and log messages become more frequent as usage approaches the freeze point.

masih: That makes sense thanks Andrew

PR #1133, opened 2023-01-10
[DO NOT MERGE] Add reader privacy

ischasny:
This PR adds Reader Privacy to storetheindex.

### DIFF
--- a/api/v0/finder/client/http/client.go
+++ b/api/v0/finder/client/http/client.go
@@ -13,12 +13,11 @@ import (
 	"github.com/ipni/storetheindex/api/v0/finder/model"
 	"github.com/ipni/storetheindex/api/v0/httpclient"
 	"github.com/libp2p/go-libp2p/core/peer"
-	b58 "github.com/mr-tron/base58/base58"
 	"github.com/multiformats/go-multihash"
 )
 
 const (
-	finderPath    = "/private/multihash"
+	finderPath    = "/multihash"
 	providersPath = "/providers"
 	statsPath     = "/stats"
 )
@@ -30,6 +29,7 @@ type Client struct {
 	finderURL    string
 	providersURL string
 	statsURL     string
+	useEncApi    bool
 }
 
 // New creates a new finder HTTP client.
@@ -38,6 +38,11 @@ func New(baseURL string, options ...httpclient.Option) (*Client, error) {
 	if err != nil {
 		return nil, err
 	}
+	var cfg httpclient.ClientConfig
+	if err := cfg.Apply(options...); err != nil {
+		return nil, err
+	}
+
 	baseURL = u.String()
 	return &Client{
 		c:            c,
@@ -45,12 +50,30 @@ func New(baseURL string, options ...httpclient.Option) (*Client, error) {
 		finderURL:    baseURL + finderPath,
 		providersURL: baseURL + providersPath,
 		statsURL:     baseURL + statsPath,
+		useEncApi:    cfg.UseEncApi,
 	}, nil
 }
 
 // Find queries indexer entries for a multihash
 func (c *Client) Find(ctx context.Context, mh multihash.Multihash) (*model.FindResponse, error) {
+	if c.useEncApi {
+		return c.encFind(ctx, mh)
+	} else {
+		return c.find(ctx, mh)
+	}
+}
 
+func (c *Client) find(ctx context.Context, mh multihash.Multihash) (*model.FindResponse, error) {
+	u := fmt.Sprint(c.finderURL, "/", mh.B58String())
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.sendRequest(req)
+}
+
+func (c *Client) encFind(ctx context.Context, mh multihash.Multihash) (*model.FindResponse, error) {
 	// query value keys from indexer
 	smh, err := dhash.SecondMultihash(mh)
 	if err != nil {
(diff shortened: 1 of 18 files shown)


api/v0/finder/client/http/client.go
```diff
@@ -8,21 +8,25 @@ import (
...
 	"github.com/ipni/storetheindex/api/v0/finder/model"
 	"github.com/ipni/storetheindex/api/v0/httpclient"
 	"github.com/libp2p/go-libp2p/core/peer"
+	b58 "github.com/mr-tron/base58/base58"
 	"github.com/multiformats/go-multihash"
 )
 
 const (
-	finderPath    = "/multihash"
+	finderPath    = "/private/multihash"
```
ischasny, line 21: Client has been changed to serve private requests only. We can have multiple flavours of the client to server both legacy and new lookups.

masih: This is a bad idea; we should infer this from the multihash code of the multihash.

ischasny: What do you mean? Client always takes the original multihash (not the second hash).

ischasny: Server will return different responses depending on whether it's a regular / double hashed lookup. So we can't possibly have a single endpoint serving both. Can you elaborate please?

masih: There is a [dedivated multicodec](https://github.com/multiformats/go-multicodec/blob/master/code_table.go#L124) for double hashed multihashes. Such multihashes are used for private lookup. We should infer the type of query by parting the multihash and based on the code respond appropriately. 

> So we can't possibly have a single endpoint serving both

Why?

ischasny: They are structurally different

masih: Let's circle back on this at colo.

api/v0/finder/client/http/client.go
```diff
@@ -37,37 +41,161 @@ func New(baseURL string, options ...httpclient.Option) (*Client, error) {
...
+				return nil, err
+			}
+			vk, err := dhash.DecryptValueKey(evkBytes, mh)
+			if err != nil {
+				return nil, err
+			}
+
+			pid, _ := c.keyer.SplitKey(vk)
+
+			pinfo, err := c.GetProvider(ctx, pid)
```
gammazero, line 183: This should probably use a different API, since this was intended to get general information about a provider, and not specifically what is needed for a find query response.

api/v0/finder/client/http/client.go
```diff
@@ -37,37 +41,161 @@ func New(baseURL string, options ...httpclient.Option) (*Client, error) {
...
+		}
+
+		if mh == nil {
+			break
+		}
+
+		mhr := model.MultihashResult{
+			Multihash: mh,
+		}
+		for _, evk := range pmhr.ValueKeys {
```
gammazero, line 171: Instead of doing a separate request for each value key, could these be combined into a single batch request?

api/v0/finder/client/http/client.go
```diff
@@ -37,37 +41,161 @@ func New(baseURL string, options ...httpclient.Option) (*Client, error) {
...
+			evkBytes, err := b58.Decode(evk)
+			if err != nil {
+				return nil, err
+			}
+			vk, err := dhash.DecryptValueKey(evkBytes, mh)
+			if err != nil {
+				return nil, err
+			}
+
+			pid, _ := c.keyer.SplitKey(vk)
```
gammazero, line 181: Many of the value keys are likely to have the same provider, so only one provider record may be needed for many value keys.

api/v0/finder/client/http/client.go
```diff
@@ -37,37 +41,161 @@ func New(baseURL string, options ...httpclient.Option) (*Client, error) {
...
+	req.Header.Add("Accept", "application/json")
+
+	resp, err := c.c.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
```
gammazero, line 73: Return here does not close `resp.Body`. Suggest adding `defer resp.Body.Close()` above, after getting the response.

api/v0/finder/client/http/client.go
```diff
@@ -37,37 +41,161 @@ func New(baseURL string, options ...httpclient.Option) (*Client, error) {
...
+	}
+
+	resBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, err
 	}
-	return c.sendRequest(req)
+	_ = resp.Body.Close()
```
gammazero, line 125: Remove this and instead do `defer resp.Body.Close()` after getting `resp`.

api/v0/finder/client/http/client.go
masih: Discussed in colo: 

We agreeded that 
* `/mh/<multihash>` would support both and change response based on the code in the given multihash.
* if the code is double-hashed, then the returned JSON object would include `EncryptedMultihashResults` which maps to an array of base64 encoded bytes instead of the regular `MultihashResults`.
* if the code is not double-hashed then no change.

This also facilitates response to POST requests that may contain a mix of double hashed and regular multihashes.

gammazero: So, it seems that the question here is which implementation do we want:
1.  The client queries the same endpoint on the server, sending either the plain or the hashed MH to the server, and the server responds with the appropriate response based on what type of MH was given.
2. The client queries a separate endpoint on the server, depending on whether it is sending a plain or a hashed MH. Those different endpoints respond with the respective response types.

gammazero: Will there ever be a case where there is a mixed set of plain and hashed MHs?

Having separate endpoints for plain and hashed MH queries has the advantage of being able to disable one or the other to support only one. Having separate endpoints also allows routing to different servers, which may be useful if double-hashing is supported on one and not another.

PR #1141, opened 2023-01-13
Configure core to use DHStore

gammazero:
Optionally configure core to use DHStore service.

### DIFF
--- a/config/indexer.go
+++ b/config/indexer.go
@@ -16,6 +16,13 @@ type Indexer struct {
 	// individual multihashes within a Put. A value of 1 means no concurrency,
 	// and zero uses the default.
 	CorePutConcurrency int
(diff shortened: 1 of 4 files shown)


masih (approved):

config/indexer.go
```diff
@@ -16,6 +16,9 @@ type Indexer struct {
 	// individual multihashes within a Put. A value of 1 means no concurrency,
 	// and zero uses the default.
 	CorePutConcurrency int
+	// DHStoreURL is the base URL for the DHStore service. This option value
+	// tells the indexer core to use a DHStore service, if configured.
+	DHStoreURL string
```
ischasny, line 21: Can we also add an ability to disable storing to the local store and use only dhstore?

PR #1155, opened 2023-01-18
Deploy latest assigner to prod

gammazero:


deploy/manifests/prod/us-east-2/tenant/storetheindex/assigner/kustomization.yaml
```diff
@@ -25,4 +25,4 @@ configMapGenerator:
 images:
 - name: storetheindex
   newName: 407967248065.dkr.ecr.us-east-2.amazonaws.com/storetheindex/storetheindex
-  newTag: 20221210085347-5e44409b6c8868df9da298d977bf9eb75ae03948
+  newTag: 20230118131811-68223393bdbc269aad5b31c470eba6fabb406885
```
gammazero, line 28: @masih Can this be set to `0.5.5` as it is with prod storetheindex?

masih: indeed; that should be totally fine 👍

PR #1158, opened 2023-01-18
Use release tag for production assigner deployment

gammazero:


deploy/manifests/base/assigner/deployment.yaml
```diff
@@ -8,11 +8,14 @@ spec:
   selector:
     matchLabels:
       app: assigner
+  strategy:
```
masih, line 11: Minor comment: Document rationale? something like:
```suggestion
  # Terminate previous assigner before rolling out new ones to avoid conflicts between assignments.
  strategy:
```

masih (approved):

PR #1160, opened 2023-01-19
Remove p2p peer ID from publisher multiaddr given to sync

### LINKED ISSUE #865: Manual sync from previously unknown peers fail with `no good addresses`

gammazero:
Sync does not work if the publisher multiaddr contains a p2p peer ID. This PR detects if there is one and truncates the multiaddress at the start of the /p2p/<peerID> part of the multiaddr.

This allows the sync command to work whether or not the specified publisher multiaddress has a p2p peer ID.

Fixes #865

### DIFF
--- a/dagsync/subscriber.go
+++ b/dagsync/subscriber.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -406,7 +407,7 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid, 
 
 	var peerAddrs []multiaddr.Multiaddr
 	if peerAddr != nil {
-		peerAddrs = []multiaddr.Multiaddr{peerAddr}
+		peerAddrs = []multiaddr.Multiaddr{rmPeerIDFromMultiaddr(peerAddr)}
 	}
 	syncer, isHttp, err := s.makeSyncer(peerID, peerAddrs, tempAddrTTL, opts.rateLimiter)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ischasny:
@gammazero the issue surfaced up when querying root CID, i.e. here: https://github.com/ipni/storetheindex/blob/main/dagsync/p2p/protocol/head/head.go#L85. Would we need to strip out p2p ID in multiple places then?

masih:
For reference I have captured [this](https://github.com/libp2p/go-libp2p/issues/2001) issue to ask libp2p folks if they think this is expected behaviour.

dagsync/subscriber.go
```diff
@@ -406,7 +407,7 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid,
 
 	var peerAddrs []multiaddr.Multiaddr
 	if peerAddr != nil {
-		peerAddrs = []multiaddr.Multiaddr{peerAddr}
+		peerAddrs = []multiaddr.Multiaddr{rmPeerIDFromMultiaddr(peerAddr)}
```
masih, line 410: The code would be much less verbose if we use `peer.SplitAddr`; e.g.:

```suggestion
		transport, _ := peer.SplitAddr(peerAddr)
		peerAddrs = []multiaddr.Multiaddr{rmPeerIDFromMultiaddr(transport)}
```

masih (changes requested):
LGTM, thank you for figuring this out Andrew 👍 

Left a comment to use `peer.SplitAddr` instead which should relax the need for string manipulation.

masih:
Fixes #865

dagsync/subscriber.go
gammazero: Agree, will update.

gammazero:
@ischasny Potentially we would need to do this in any place that the indexer dials another host.  Other than CLI use, This is really only done during sync, so I think we are OK. This should all go away completely when there is a fix for the [libp2p issue](https://github.com/libp2p/go-libp2p/issues/2001) that was filed.

PR #1161, opened 2023-01-19
Deploy `dhstore` to `dev` on minimal resources

masih:
Deploy `dhstore` as an internally accessible service only on `dev` with minimal resource demand and a 10Ti `io2` PVC volume.

The internal service configures an internal DNS name for the service, resolvable only within the VPC pointing at an internal NLB.


ischasny (approved):
BOOM!

PR #1172, opened 2023-01-19
Revert assigner deployment strategy

gammazero:
It appears to be causing a deployment problem: Deployment/storetheindex/assigner dry-run failed, reason: Invalid, error: Deployment.apps "assigner" is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'...


masih:
The changes reverted seem correct; I am going to give this a shot on dev only to see if I can reproduce the issue.

PR #1188, opened 2023-01-23
Increase ago memory limit

ischasny:
Increate ago memory limit as it has been OOMKilled multiple times over the weekend.

See https://github.com/ipni/storetheindex/issues/1185

### DIFF
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/ago/deployment.yaml
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/ago/deployment.yaml
@@ -16,10 +16,23 @@ spec:
           resources:
             limits:
               cpu: "3"
-              memory: 10Gi
+              memory: 28Gi
             requests:
               cpu: "3"
-              memory: 10Gi
+              memory: 28Gi
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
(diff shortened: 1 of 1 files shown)


deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/ago/deployment.yaml
```diff
@@ -16,10 +16,10 @@ spec:
           resources:
             limits:
               cpu: "3"
-              memory: 10Gi
+              memory: 20Gi
```
masih, line 19: the node currently running on [`m4.xl`](https://github.com/ipni/storetheindex/blob/main/deploy/infrastructure/dev/us-east-2/eks.tf#L28) with maximum of 16GiB memory. This is a relatively cheap instance. Can I ask you to take a look through [instance types](https://aws.amazon.com/ec2/instance-types/) and find one with sufficient memory for this node that is also cheap?  

Otherwise, merging this PR will trigger r5b scale up of nodes that are much more expensive.

masih: `r6g.xlarge` is [$`0.0016` more expensive](https://aws.amazon.com/ec2/pricing/on-demand/) that `m4.xlarge` and gives us 32GiB` of memory. Let's create a new node-group, in dev with that instance type, then set the node selectors for `ago` to always gets deployed on that node type.

masih (approved):

PR #1201, opened 2023-01-27
Increase dhstore RAM in dev

ischasny:
Increase DHStore RAM in DEV as it has been OOMKilled recently:
```
    Last State:     Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Fri, 27 Jan 2023 10:53:27 +0000
      Finished:     Fri, 27 Jan 2023 10:59:44 +0000
    Ready:          True
    Restart Count:  97
```

### DIFF
--- a/deploy/infrastructure/dev/us-east-2/eks.tf
+++ b/deploy/infrastructure/dev/us-east-2/eks.tf
@@ -29,6 +29,14 @@ module "eks" {
       instance_types = ["r6a.xlarge"]
       subnet_ids     = [data.aws_subnet.ue2a2.id]
     }
+    # Node group used by dhstore nodes
+    dev-ue2c-r6a-xl = {
+      min_size       = 0
+      max_size       = 3
+      desired_size   = 1
+      instance_types = ["r6a.xlarge"]
+      subnet_ids     = [data.aws_subnet.ue2c1.id]
+    }
     dev-ue2-m4-xl-2 = {
       min_size       = 3
       max_size       = 7
--- a/deploy/infrastructure/dev/us-east-2/eks.tf
+++ b/deploy/infrastructure/dev/us-east-2/eks.tf
@@ -35,7 +35,7 @@ module "eks" {
       max_size       = 3
       desired_size   = 1
       instance_types = ["r6a.xlarge"]
-      subnet_ids     = [data.aws_subnet.ue2c1.id]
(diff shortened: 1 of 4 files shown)


willscott (approved):

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore/deployment.yaml
```diff
@@ -19,10 +19,10 @@ spec:
           resources:
             limits:
               cpu: "3"
-              memory: 10Gi
+              memory: 20Gi
             requests:
               cpu: "3"
-              memory: 10Gi
+              memory: 20Gi
```
masih, line 25: As is the chances are this manifest is going to get scheduled on expensive node types or not get scheduled at all.

We have created a memory optimised worker group to run `ago` before. 

I recommend configuring node affinity for this deployment to use that.

Check that the AZ for the node group matches the AZ of PVC used by dhstore. If not create a new worker node group with the same instance type as the ones used by `ago` and use that.

masih (changes requested):

deploy/infrastructure/dev/us-east-2/eks.tf
```diff
@@ -29,6 +29,14 @@ module "eks" {
       instance_types = ["r6a.xlarge"]
       subnet_ids     = [data.aws_subnet.ue2a2.id]
     }
+    # Node group used by dhstore nodes
+    dev-ue2c-r6a-xl = {
+      min_size       = 0
+      max_size       = 3
+      desired_size   = 1
+      instance_types = ["r6a.xlarge"]
+      subnet_ids     = [data.aws_subnet.ue2c1.id]
```
masih, line 38: Let's use c2. if there is no `data` resource filter for it let's create one.

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore/deployment.yaml
```diff
@@ -19,10 +19,23 @@ spec:
           resources:
             limits:
               cpu: "3"
-              memory: 10Gi
+              memory: 20Gi
```
masih, line 22: Let's increase both memories to 28Gi considering the worker node type offers up to 32Gi.

PR #1203, opened 2023-01-28
Keep entries when keeping ads

gammazero:
When advertisements are not deleted, for keeping a local copy, the entries chunks and HAMTs also need to be kept.


willscott (approved):

masih:
Do we need to reingest from the beginning on ago after this is deployed?

PR #1210, opened 2023-02-01
Save advertisements as CAR files

gammazero:
When configured to save advertisements, save advertisements and entries from the datastore as CAR files in the configured directory. The advertisement and its entries are removed from the datastore after they are saved as a CAR file.

If there are existing advertisements saved in the datastore, on startup, save these as CAR files and remove them from the datastore.

Advertisements are not saved when the indexer is frozen.

- Fetching advertisements and entries from another indexer, or mirror location, will be implemented in a subsequent PR.
- Saving advertisement CAR files to S3 (or other locations) will be done in a subsequent PR.


config/ingest.go
```diff
@@ -41,9 +41,9 @@ type Ingest struct {
 	// IngestWorkerCount sets how many ingest worker goroutines to spawn. This
 	// controls how many concurrent ingest from different providers we can handle.
 	IngestWorkerCount int
-	// KeepAdvertisements, when true, prevents advertisements from being
-	// deleted after they are ingested.
-	KeepAdvertisements bool
+	// KeepAdvertisementsCarDir specifies a directory to write CAR files
+	// containing advertisements that have been ingested.
+	KeepAdvertisementsCarDir string
```
gammazero, line 46: - Should this be changed to `CarSaveLocation` or `MirrorLocation` and take the form of a URI?
- Do we want to save to multiple locations, like directory and S3?
- Do we want the indexer to be responsible for moving the CAR files to the final mirror location (e.g. S3) or should a separate process do that?

internal/ingest/ingest.go
```diff
@@ -1121,9 +1146,18 @@ func (ing *Ingester) ingestWorkerLogic(ctx context.Context, provider peer.ID) {
 			return
 		}
 
-		if markErr := ing.markAdProcessed(assignment.publisher, ai.cid, frozen); markErr != nil {
+		keep := ing.adToCar != nil
+		if markErr := ing.markAdProcessed(assignment.publisher, ai.cid, frozen, keep); markErr != nil {
 			log.Errorw("Failed to mark ad as processed", "err", markErr)
 		}
+		if !frozen && keep {
+			_, err = ing.adToCar.Write(ai.cid, false)
```
gammazero, line 1154: This could be done asynchronously by queuing the ad CIDs and having a limited number of goroutines write the CAR files. It would be at the cost of more memory and risk of building up a backlog of ads to save as CAR files. The synchronous solution was chosen because it is simpler and is already done in the publisher's worker goroutine so other publishers are not blocked,

config/ingest.go
masih: Ideally, we want to _stream_ the car files directly to S3. That may involve temporary storage of files on local disk. 

Instead of string location, the config should allow "CarWriter" config or something along those lines with 2 implementations: one to write on local disk that asks for a destination path, and one that uploads to S3 directly.

Please leave the implementation of the latter writer out for a separate PR.

gammazero: Streaming directly to S3 has two disadvantages:
1. Requires a separate detached index for the CAR file.
2. Cannot use a separate process to send files to S3 or another destination.

If we care about either of these then it may be worth writing CAR files to the local filesystem temporarily.

willscott: I'm less concerned about S3 specifically as the long-term goal. If we can get to car files in local memory/disk so that we can turn them into filecoin deals or similar that should be just about as good

masih: > Requires a separate detached index for the CAR file.

Fine to temporarily write them to disk. This should be encapsulated in the exporter implementation. Even without it, the deal making and upload to FileCoin without a doubt needs to be a separate service/thing from storetheindex and it is easy enough to generate index there if needed, assuming we cannot upload carv1s straight up.

> Cannot use a separate process to send files to S3 or another destination.

Off the top of my head we can have a separate process or a side car container that watches a directory and uploads. What am I missing?

gammazero: Missing nothing. We can still export CARs to local filesystem and then write to S3 using a separate process if we want.

config/ingest.go
```diff
@@ -76,6 +76,29 @@ type Ingest struct {
...
+	// Path to filesystem directory where files are stored.
+	BasePath string
+}
+
+type S3FileStore struct {
+	BasePath   string
+	BucketName string
+	Region     string
+	AccessKey  string
+	SecretKey  string
```
masih, line 99: `Region` will typically be set via environment variables that are well known to the AWS sdk, standardised across all languages.

`AccessKey` and `Secret(Access)Key` are typically not specified in code at all for security reasons. Instead, the authentication is done by assuming a role, for which we do not need to write any code at all. This is handled by the authentication chain standardised and built into AWS SDK. We will configure the pod in k8s-level to assume a role that would then allows it to implicitly access the configured bucket.

Following that train of thought, for S3 we then only need a bucket name, and "Key",i.e. the path to which the data is uploaded, which i suspect is the same as `BasePath`? I recommend using terminology that is consistent with that of S3, i.e. Key instead of base path. [Here](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/gov2/s3/actions/bucket_basics.go#L103) are some examples.

internal/carwriter/carwriter.go
```diff
@@ -0,0 +1,413 @@
...
+
+	log.Infow("Wrote advertisement CAR", "adFile", adCarFile.Path, "adFileSize", adCarFile.Size,
+		"entriesFile", entCarFile.Path, "entriesFileSize", entCarFile.Size)
+
+	return adCarFile.Path, entCarFile.Path, nil
+}
+
+// writeAdToCar wites the advertisement data to a CAR file.
+func (cw *CarWriter) writeAdToCar(ctx context.Context, adCid cid.Cid, data []byte) (*filestore.File, error) {
+	fileName := adCid.String() + "_adv.car"
```
masih, line 159: I am curious why include `_adv` and `_mhs` suffix? Would CID alone not be enough? 

If we add such prefixes, a client that only has a CID and does not know what node it contains would need to hit two endpoints.

If you think we need this, would it make sense to use path instead of file suffix? In the case of s3 that gives us advantages in terms of controlling things like retention per node type.

masih (approved):

internal/carwriter/carwriter.go
gammazero: All good points; agree.

config/ingest.go
gammazero: My thinking was to allow the _option_ of specifying these in indexer config instead of setting it elsewhere in the infrastructure. So, generally, a `EnvProvider` or `SharedCredentialsProvider` or `EC2RoleProvider` will be used in the credentials chain, but shouldn't there be an option to have a `StaticProvider`?

masih: I don't think it's needed considering the built in auth chain in vanilla AWS sdk.

PR #1218, opened 2023-02-01
Remove optional metadata query parameter

ischasny:
Remove metadata query parameter that changed behaviour of whether metadata has been included into extended provider results. Including metadata is required so that the client can correctly assemble extended provider results.

### DIFF
--- a/api/v0/finder/client/interface.go
+++ b/api/v0/finder/client/interface.go
@@ -20,7 +20,7 @@ type Finder interface {
 	// ListPrividers gets information about all providers known to the indexer.
 	// The withExtMetadata requests that extended provider metedata be
 	// included in the response.
-	ListProviders(ctx context.Context, withExtMetadata bool) ([]*model.ProviderInfo, error)
+	ListProviders(ctx context.Context) ([]*model.ProviderInfo, error)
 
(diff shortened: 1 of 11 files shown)


gammazero (changes requested):
Also need to remove parameter from clients:
- https://github.com/ipni/storetheindex/blob/main/api/v0/finder/client/http/client.go#L74
- https://github.com/ipni/storetheindex/blob/main/api/v0/finder/client/libp2p/client.go#L98-L107
- https://github.com/ipni/storetheindex/blob/main/api/v0/finder/client/interface.go#L23
- https://github.com/ipni/storetheindex/blob/main/command/providers.go#L62

ischasny:
@gammazero good catch, thanks. Removed - please re-review.

api/v0/finder/client/interface.go
```diff
@@ -20,7 +20,7 @@ type Finder interface {
 	// ListPrividers gets information about all providers known to the indexer.
 	// The withExtMetadata requests that extended provider metedata be
 	// included in the response.
```
gammazero, line 22:
```suggestion
```

gammazero (approved):
One suggestion.

PR #1228, opened 2023-02-02
Add ExtendedProviders to the double hashed client

ischasny:
* Add extended providers to double hashed client;
* Maintain `ProviderInfo`s cache at the client level instead of per-request. This is done to save up some time on roundtrips as well as to index `ContextualExtendedProviders` by `ContextID` for faster lookup times.

### DIFF
--- a/api/v0/finder/client/http/dhash_client.go
+++ b/api/v0/finder/client/http/dhash_client.go
@@ -1,11 +1,13 @@
 package finderhttpclient
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
 	"github.com/ipni/dhstore"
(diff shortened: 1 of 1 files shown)


willscott (commented):
is it worth adding prometheus metrics to the providercache so we have a sense of how big it is in memory and how much of a hit rate it's giving us?

ischasny:
> is it worth adding prometheus metrics to the providercache so we have a sense of how big it is in memory and how much of a hit rate it's giving us?

Bear in mind - the library is going to sit on the client side (i.e. it's not a server side cache). Do we report into prometheus from the client libraries anywhere already?

willscott (approved):
don't need to report always, but worth having the metrics, and maybe a way to configure max cache sizing?

ischasny:
> don't need to report always, but worth having the metrics, and maybe a way to configure max cache sizing?

Makes sense - will add GH issues for both.

PR #1253, opened 2023-02-09
Add client reconciler

ischasny:
Add reconciliation script between regular and double hashed clients

### DIFF
--- a/scripts/reconcile_dh_index/main.go
+++ b/scripts/reconcile_dh_index/main.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"sort"
 	"strings"
+	"time"
 
 	finderhttpclient "github.com/ipni/storetheindex/api/v0/finder/client/http"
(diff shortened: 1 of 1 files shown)


scripts/reconcile_dh_index/main.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/libp2p/go-libp2p/core/peer"
+	b58 "github.com/mr-tron/base58/base58"
+	"github.com/multiformats/go-multihash"
+)
+
+// This program can be used to reconcile results from regular and double hashed clients.
+// It works as following:
+// * fetch list of providers from source and target indexers
+// * find providers that exist in both indexers and that have the same last ad cid
+// * for each provider fetch last add via provider CLI
```
masih, line 27: It would be really nice if we can make this script to not depend on `provider` binary for ease of use, specially considering storetheindex offers go libraries to do what the provider CLI would do and this script lives in storetheindex repo.

PR #1260, opened 2023-02-11
Use std go http package instead of gorilla

### LINKED ISSUE #1212: Consider moving away from `gorilla/mux` for request handing
The HTTP server implementations in this repo use `gorilla/mux` framework, which [has been archived since late 2022](https://github.com/gorilla#gorilla-toolkit).  Consider moving away from the framework.

I recommend using the vanilla go `http` package considering the APIs offered by storetheindex are simple enough.

gammazero:
Should use the standard go `http` package unless there is a specific reason not to, which there is not.

Fixes #1212


willscott (approved):

masih (approved):
Thank you for getting this done @gammazero 👍

PR #1265, opened 2023-02-14
Add dhstore prometheus pod monitor

ischasny:


deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore/deployment.yaml
```diff
@@ -23,6 +23,9 @@ spec:
             requests:
               cpu: "3"
               memory: 28Gi
+          ports:
+            - containerPort: 40081
+              name: metrics
```
masih, line 28: It's fine to add this here. Eventually we should also update the manifests in `dhstore` since the code expoes that port now.

masih (approved):

PR #1276, opened 2023-02-16
Use multiaddr built-in filtering to reduce code

gammazero:


mautil/mautil.go
```diff
@@ -13,45 +13,18 @@ import (
...
-			}
-			continue
-		}
-		newAddrs = append(newAddrs, maddr)
+func notPrivateAddr(a multiaddr.Multiaddr) bool {
+	if a == nil {
+		return true
 	}
-	return newAddrs
+	return !manet.IsPrivateAddr(a)
```
masih, line 27: Unfortunately this won't check unroutable Addrs.

mautil/mautil.go
```diff
@@ -13,45 +13,18 @@ import (
...
-
-	// If no private addrs, return original slice.
-	if len(pvt) == 0 {
-		return maddrs
-	}
-
-	// If no non-private addrs, return nothing.
-	notPrivLen := len(maddrs) - len(pvt)
-	if notPrivLen == 0 {
+	filtered := multiaddr.FilterAddrs(maddrs, notPrivateAddr)
```
masih, line 16: Similar to the approach I have implemented [here](https://github.com/ipni/caskadht/pull/21), I think something like this should suffice:

```suggestion
	filtered := multiaddr.FilterAddrs(maddrs, func(target multiaddr.Multiaddr) bool {
		c, _ := multiaddr.SplitFirst(target)
		if c == nil {
			return false
		}
		switch c.Protocol().Code {
		case multiaddr.P_IP4, multiaddr.P_IP6, multiaddr.P_IP6ZONE, multiaddr.P_IPCIDR:
			return manet.IsPublicAddr(target)
		case multiaddr.P_DNS, multiaddr.P_DNS4, multiaddr.P_DNS6, multiaddr.P_DNSADDR:
			return c.Value() != "localhost"
		default:
			return true
		}
	})
```

masih (approved):

PR #1277, opened 2023-02-16
Write advertisement and entries into one CAR file

gammazero:
An advertisement and its multihash entries blocks are all stored in a single CAR file. The car file has two roots, the advertisement CID and the entries CID. The entries CID root may be omitted if the advertisement has not entries.

### DIFF
--- a/carwriter/carwriter.go
+++ b/carwriter/carwriter.go
@@ -40,12 +40,12 @@ func (e WriteError) Unwrap() error {
 }
 
 // CarWriter writes advertisements and entries, that are stored in a datastore,
-// to CAR files. For each advertisemtns, two separate CAR files are writen. One
-// for the advertisement data and one for all of the multihash entries chunks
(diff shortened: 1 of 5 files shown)


carwriter/carwriter.go
```diff
@@ -40,12 +40,12 @@ func (e WriteError) Unwrap() error {
 }
 
 // CarWriter writes advertisements and entries, that are stored in a datastore,
-// to CAR files. For each advertisemtns, two separate CAR files are writen. One
-// for the advertisement data and one for all of the multihash entries chunks
-// or HAMT. The original chunks or HAMT are preserved for the purpose of being
-// able to verify links from advertisements and verify the advertisement
-// signature. Such verification may be necessary when fetching car files from a
-// location that is not completely trusted.
+// to CAR files. Each advertisemtns and its associated multihash entries
```
willscott, line 43:
```suggestion
// to CAR files. Each advertisement and its associated multihash entries
```

willscott (approved):

PR #1286, opened 2023-02-20
Move dhstore to c5n.2xlarge instance

ischasny:
Move dhstore to a compute optimised instance that have more CPUs available

### DIFF
--- a/deploy/infrastructure/dev/us-east-2/eks.tf
+++ b/deploy/infrastructure/dev/us-east-2/eks.tf
@@ -30,11 +30,11 @@ module "eks" {
       instance_types = ["r6a.xlarge"]
       subnet_ids     = [data.aws_subnet.ue2a2.id]
     }
-    dev-ue2c-r6a-xl = {
+    dev-ue2c-c5n-2xl = {
       min_size       = 0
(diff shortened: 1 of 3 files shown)


deploy/infrastructure/dev/us-east-2/eks.tf
```diff
@@ -30,11 +30,11 @@ module "eks" {
       instance_types = ["r6a.xlarge"]
       subnet_ids     = [data.aws_subnet.ue2a2.id]
     }
-    dev-ue2c-r6a-xl = {
+    dev-ue2c-c5n-2xl = {
       min_size       = 0
       max_size       = 3
       desired_size   = 1
-      instance_types = ["r6a.xlarge"]
+      instance_types = ["c5n.2xlarge"]
```
masih, line 37: * Do not modify existing node groups unless no pods are running on it. Pods should be drained off the node group after a new one is created. Then the old node group can be safely destroyed :) 

* Can you add a short description to the PR to rationalise the choice of this specific instance type please?

ischasny: Sounds good, will do

PR #1292, opened 2023-02-21
Deploy dhstore to larger instance

ischasny:
* Currently dhstore runs on `r6a.xlarge` at close to 100% CPU load
* `r5n.2xlarge` will give us 8 vCPUs (vs 4 currently) as well as 64GiB RAM (vs 32 currently)
* Compute optimised instances with 8 vCPUs have less than 32GiB RAM. Compute optimised instances with 16 vCPUs and 32 GiB RAM are at least as expensive as `r5n.2xlarge`.

### DIFF
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore/deployment.yaml
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore/deployment.yaml
@@ -35,7 +35,7 @@ spec:
                   - key: node.kubernetes.io/instance-type
                     operator: In
                     values:
-                      - r6a.xlarge
+                      - r5n.2xlarge
                   - key: topology.kubernetes.io/zone
                     operator: In
                     values:


deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore/deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
                   - key: node.kubernetes.io/instance-type
                     operator: In
                     values:
-                      - r6a.xlarge
+                      - r5n.2xlarge
```
masih, line 38: Two questions:
* Are the existing node groups for this instance type in the right AZ?
* What is the max number of instances in the corresponding node group and does it have headroom to grow in order to fit the new node?

ischasny: * new node group is in [us-east-2c](https://github.com/ipni/storetheindex/blob/main/deploy/infrastructure/dev/us-east-2/eks.tf#L91), same to what was before
* trhe max size is 3 and it's currently running 2 instances - ber and cali.

ischasny: Actually ber runs in (2b)[https://github.com/ipni/storetheindex/blob/main/deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/ber/deployment.yaml#L33]. So it's just one node there - cali. Do you still want to bump to 5?

masih: Good not needed. 

The thing to point out here is that rolling deployment starts up a pod, wait for it to be ready before tearing down the old one. That's why extra headroom is needed in ASG so that new worker nodes can be started to run the deployment. 

It all cleans itself up back to minimal number of nodes.

masih (approved):

PR #1295, opened 2023-02-21
Bump up dhstore cpu limits

ischasny:


deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore/deployment.yaml
```diff
@@ -19,10 +19,10 @@ spec:
               mountPath: /data
           resources:
             limits:
-              cpu: "3"
+              cpu: "6"
               memory: 28Gi
```
masih, line 23: While at it bump the memory too now that dhstore runs on dedicated instance?

ischasny: makes sense - bumped

masih (approved):

PR #1330, opened 2023-02-27
Write advertisements and publisher heads from datastore to carstore

gammazero:
At indexer startup, the latest processed advertisement CID for each publisher is written to the car store as a head file. Then all existing advertisements are read from the datastore, and in the background, these are written to the car store as CAR files.

During normal operations, when the latest advertisement is written to a CAR file, a head file is written for the publisher as well. This head file is later used by other indexers when importing CAR files, as a place to stop importing directly from the publisher to switch to importing advertisements from the carstore.

This PR is in preparation for a following PR that imports advertisement data from a carstore.


carstore/carwriter_test.go
```diff
@@ -7,15 +7,15 @@ import (
 	"math/rand"
 	"testing"
 
-	"github.com/ipfs/go-cid"
+	//"github.com/ipfs/go-cid"
```
masih, line 10: Needed?

masih (approved):

PR #1331, opened 2023-02-27
Show inactive providers when listing or getting provider info

gammazero:
An indication is only shown when the provider is inactive. When active, no indication is given.


ischasny (approved):

ischasny:
@gammazero is this one good to be merged?

gammazero:
@ischasny Yes, was trying to see why some tests are failing intermittently.

PR #1358, opened 2023-03-06
Disable accelerated DHT on `caskadht` instances

masih:
Spikes of p95th lookup latency at cid.contact correlate to CPU spikes of DHT lookup cascades. The service is already given 30 CPU cores which is a lot. To maintain DHT lookup cascading as a feature we need a more efficient implementation of DHT client mode. Before any further work on optimising the accelerated DHT client, however, we should understand how the non-accelerated version behaves.

Disable accelerated DHT on `dev` and `prod` for `caskadht` instances in order to understand:
* resource consumption between the two
* lookup performance between the two


masih:
Deferred until the streaming candidate finder rollout on Rhea to limit moving parts.

PR #1359, opened 2023-03-06
Resolve gogo proto deprecation

masih:
Move away from gogo proto to resolve CI failures related to its deprecation.


gammazero (approved):
Thank you for taking care of this. LGTM.

PR #1377, opened 2023-03-08
Wrap the data transfer datastore to remove the need for FSM migration

masih:
Wrap the data store passed to data transfer manager with a namespace, such that it starts fresh and remove the need for migrating the old FSM to v2.

Relates to:
 - https://github.com/ipni/storetheindex/issues/1371

### DIFF
--- a/dagsync/subscriber.go
+++ b/dagsync/subscriber.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/ipld/go-ipld-prime"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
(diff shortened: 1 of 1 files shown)


dagsync/subscriber.go
```diff
@@ -169,6 +170,7 @@ func NewSubscriber(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem,
 		}
 		dtSync, err = dtsync.NewSyncWithDT(host, opts.dtManager, opts.graphExchange, &lsys, blockHook)
 	} else {
+		ds := namespace.Wrap(ds, datastore.NewKey("data-transfer-v2"))
```
gammazero, line 173: Doing this will have the effect of adding the prefix `"data-transfer-v2"` to all new keys stored in the datastore, and will make all the existing keys inaccessible.  All of the other items stored in the datastore, that we care about keeping, already have their own prefixes.  Perhaps it would be better to delete all the keys that do not have one of said prefixes, to avoid migration. ... or am I missing something about what this change actually does?

masih: >to all new keys stored in the datastore 

_For data transfer manager only_. Note that DAGs are stored in the linksystem that's passed in. 

As for other keys, that are storetheindex specific, they should continye to be accessibl. Unless I have missed something.

gammazero: If we are OK with leaving FSM state garbage behind, then this will get our nodes working again.

gammazero (approved):

dagsync/subscriber.go
masih: https://github.com/ipni/storetheindex/issues/1380

PR #1383, opened 2023-03-08
Print commands tree

gammazero:
Provide a new `commands` command that prints a tree view of the available commands. This helps users see what commands are available and where to find them.  The tree is static text and can be seen in the diffs for this PR.


command/tree.txt
```diff
@@ -0,0 +1,33 @@
+|-- admin           Perform admin activities with an indexer
```
masih, line 1: We should generate this from urfav command objects if we really want it.

gammazero: Agree, but did not find any built-in way to do it.

PR #1384, opened 2023-03-08
No separate ads datastore

gammazero:
## Context
Before the content advertisement mirrors were available, the indexer could be made to keep advertisements and entries data in a separate datastore from the one that other indexer state data is kept in. This datastore could then be copied to other indexers that were starting up to quickly bootstrap indexing.

Now that advertisement mirrors are available as an alternate source of index data, building up advertisements in a datastore is no longer necessary. Any existing ad datastores are no longer needed and can be discarded.

## Proposed changes
Do not use a separate datastore for temporary ad data

- Use the same datastore instance for temporary data and indexer state
- Remove configuration for separate ad datastore

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -107,14 +107,11 @@ func daemonAction(cctx *cli.Context) error {
 	log.Info("Valuestore initialized")
 
 	// Create datastore
-	dstore, dstoreAds, err := createDatastore(cfg.Datastore)
+	dstore, err := createDatastore(cfg.Datastore)
 	if err != nil {
 		return err
 	}
 	defer dstore.Close()
-	if dstoreAds != nil {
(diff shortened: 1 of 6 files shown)


ischasny (approved):
We need to be careful with rolling that out, in particular - are any of the current indexers using separate ad datastore?

gammazero:
The only indexer using a seperate ads datastore is ago.  Even if there were other, there is no need to keep that datastore. It stores only what is usually the temporary ad and entries data from already ingested ads. None of the indexer state data is kept in there.

gammazero:
@masih Is there any reason that we would want a separate leveldb instance for the temporary advertisement and entry data?

masih (approved):

PR #1398, opened 2023-03-14
Add get pending syncs command

ischasny:
Currently there is no functionality for checking whether there are any syncs in progress. That complicates troubleshooting, specifically when the indexer seems to be busy with doing something without writing anything into the datastore. This PR adds:
* a new admin endpoint and respective CLI functionality for fetching a sorted list of currently pending syncs
* it also prevents a sync for the same peer from being submitted while there is one already in-progress.

### DIFF
--- a/command/admin.go
+++ b/command/admin.go
@@ -29,7 +29,7 @@ var AdminCmd = &cli.Command{
 
 var syncCmd = &cli.Command{
 	Name:   "sync",
-	Usage:  "Sync indexer with provider",
+	Usage:  "Sync indexer with provider. If no flags are provided returns a list of currently pending syncs.",
 	Flags:  adminSyncFlags,
 	Action: syncAction,
 }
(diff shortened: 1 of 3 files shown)


command/admin.go
```diff
@@ -29,18 +29,17 @@ var AdminCmd = &cli.Command{
 
 var syncCmd = &cli.Command{
 	Name:   "sync",
-	Usage:  "Sync indexer with provider",
+	Usage:  "Sync indexer with provider. If no flags are provided returns a list of currently pending syncs.",
```
masih, line 32: Would it make sense to separate this into `sync list-pending` or something along those lines? Mixing it with `sync` such that it works without publisher ID _and_ does not sync instead returns list of pending syncs is something that I would find surprising as a user: There is no way I'd guess that what it would do.

ischasny: @masih done

masih (approved):

PR #1403, opened 2023-03-14
Fix missing addr handling

### LINKED ISSUE #1402: Cannot process ad chain if initial ad missing routable addrs
If a publisher publishes an ad chain, and the first advertisement does not have a routable _provider_ address, the indexer will stop ingesting the ad chain.  This error will happen every time there is an attempt to ingest the.  Once the indexer has a routable address for the provider, subsequent ads that are missing that data will use the previous value.

This error is indicated by the log message:
```json
{
"level":"error",
"ts":"2023-03-14T19:02:46.692Z",
"logger":"indexer/ingest",
"caller":"ingest/ingest.go:1196",
"msg":"Error while ingesting ad. Bailing early, not ingesting later ads.",
"publisher":"redated",
"adCid":"redacted",
"err":"registerErr: could not register/update provider info: missing provider address",
"adsLeftToProcess":10226
}
```
A solution to this is to treat the "missing provider address" error as a permanent error and skip leading advertisements in the chain that are missing routable provider addresses. It would be better no to do this, however, because a later advertisement may be able to provide a usable provider address that would allow all the earlier ads with missing provider addresses to be ingested.

A fix for this issue should use the latest valid provider address when updating provider info for all previous ads. Do so will also have the added benefit that the latest correct provider information is served by the indexer, even if the indexer has not yet ingested the whole ad chain.

### LINKED ISSUE #1404: Some types of ingestion errors are handled incorrectly
When ingesting entry chunks, any error during the ingestion of chunks is treated as a permanent error. Permanent errors cause the advertisement sync to be ended and the ad marked as processed, with the assumption that it is not usable. When there is an error reading or writing to the datastore or valuestore during entry chunk ingestion, this also gets marked as a permanent error, which is wrong because there is nothing wrong with the data and it should be retried in the future. Instead, such errors should be designated as a non-permanent type of error.

### LINKED ISSUE #1405: Remove unused `MhStoreNanoseconds` metric
This metric is not used anywhere, and is not particularly useful given that the view is only of the last value.  It does generate two syscalls per multihash entries chunk, so removing it would be nice.

gammazero:
## Context
Fixes #1402
Fixes #1404
Fixes #1405 

See the above issues for context.

Cannot process advertisement chain if the first advertisement is missing provider addresses. This is preventing us from ingesting some chains where the first advertisement(s) do not have routable addresses, even if later advertisements do provide routable addresses.

Some non-permanent errors are being incorrectly handled as permanent errors and causing advertisements to be discarded.

## Proposed Changes
This PR fixes #1402 by trying to update using the provider addresses from the head advertisement when ingesting ads deeper in the ad chain.

This PR fixes #1404 by correctly designating errors during ingestion of entry chunks.
 
## Tests
Included a test of an ad chain with the first ad missing the provider addresses. Fails without this PR.

## Revert Strategy
Change is safe to revert.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -956,28 +956,28 @@ func (ing *Ingester) runIngestStep(syncFinishedEvent dagsync.SyncFinished) {
 
 	// 2. For each provider put the ad stack to the worker msg channel. Each ad
 	// stack contains ads for a single provider, from a single publisher.
-	for p, adInfos := range adsGroupedByProvider {
+	for providerID, adInfos := range adsGroupedByProvider {
 		ing.providersBeingProcessedMu.Lock()
-		if _, ok := ing.providersBeingProcessed[p]; !ok {
-			ing.providersBeingProcessed[p] = make(chan struct{}, 1)
+		if _, ok := ing.providersBeingProcessed[providerID]; !ok {
+			ing.providersBeingProcessed[providerID] = make(chan struct{}, 1)
 		}
-		wa, ok := ing.providerAdChainStaging[p]
+		wa, ok := ing.providerAdChainStaging[providerID]
 		if !ok {
 			wa = &atomic.Value{}
-			ing.providerAdChainStaging[p] = wa
+			ing.providerAdChainStaging[providerID] = wa
 		}
 		ing.providersBeingProcessedMu.Unlock()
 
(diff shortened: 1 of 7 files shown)


masih (approved):

ischasny (approved):

ischasny:
@gammazero FYI - merging it and deploying to dev

PR #1411, opened 2023-03-16
Remove explicitly defined sservice account

ischasny:
Fixes k8s error: 

```
0s          Warning   FailedCreate                 replicaset/inga-indexer-59bccc7945                              Error creating: pods "inga-indexer-59bccc7945-" is forbidden: error looking up service account storetheindex/storetheindex: serviceaccount "storetheindex" not found
```

Explicit service account isn't needed as `inga` doesn't need to write to s3.


masih (approved):
Context: The reason this was added to `dev` is to grant node access to s3 so that it can write data to it via role assumption.

No S3 writes are set up on prod yet and that can be done in a separate PR as long as the new node is not configured to write to s3. 

Cc @gammazero  as we need to set this up in prod eventually.

PR #1412, opened 2023-03-16
Hook up dhstore to indexstar in prod

ischasny:


masih:
Let's hold off on this until the latency in prod is figured out

ischasny:
> Let's hold off on this until the latency in prod is figured out

Seems that it has calmed down since yesterday. @masih wdyt?

masih (approved):

ischasny:
FYI I'll also add dhstore to it too

PR #1420, opened 2023-03-17
Update dhfind service configuration

ischasny:


masih (changes requested):
service.yaml is a patch. The changes here make no difference. Look at service.yaml in base layer

PR #1447, opened 2023-03-21
Temporarily increase pebble block cache size

ischasny:
* This is to test whether bigger block cache will help to make read latencies smoother.
* Will be deployed only to dido in prod where spiky latencies are observed.

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -586,7 +586,7 @@ func createValueStore(ctx context.Context, cfgIndexer config.Indexer) (indexer.I
 			l.EnsureDefaults()
 		}
 		pebbleOpts.Levels[numLevels-1].FilterPolicy = nil
-		pebbleOpts.Cache = pbl.NewCache(1 << 30) // 1 GiB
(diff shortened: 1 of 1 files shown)


command/daemon.go
```diff
@@ -586,7 +586,7 @@ func createValueStore(ctx context.Context, cfgIndexer config.Indexer) (indexer.I
 			l.EnsureDefaults()
 		}
 		pebbleOpts.Levels[numLevels-1].FilterPolicy = nil
-		pebbleOpts.Cache = pbl.NewCache(1 << 30) // 1 GiB
+		pebbleOpts.Cache = pbl.NewCache(2 << 30) // 2 GiB
```
masih, line 589: Let's parameterise this as a config. As is it would change the setting for every node that  deploys the built image

gammazero:
Is this PR replaced by #1448?

ischasny:
> Is this PR replaced by #1448?

Yes, thanks @gammazero - closing this PR

PR #1453, opened 2023-03-22
Deploy latest dhstore to dev and prod

ischasny:
* Deploy the latest version that supports configurable block cache size to dev and prod
* Set block cache size to 2Gi in prod as that where we see latency spikes


willscott (approved):

willscott:
do you want to do dev before prod to confirm it works?

ischasny:
> do you want to do dev before prod to confirm it works?

Fair enough - removed the prod part. Will deploy in a follow up PR.

PR #1463, opened 2023-03-24
Sync now accepts multiple addresses

### LINKED ISSUE #321: Handle multiple publisher addresses
Announce messages may have multiple publisher addresses.  Sync should accept multiple addresses and use them instead of only using a single address.

`dagsync/dtsync` and `dagsync/httpsync` should both make use of multiple publisher addresses.

If there are multiple publisher addresses for a provider, use the addresses for auto-sync until a working address is found.

gammazero:
## Context
Sync should make use of multiple publisher addresses. Announce messages may contain multiple publisher addresses, and the indexer is only using a single one of these. All should be used. 

Fixes #321

Will also fix this: https://github.com/ipni/index-provider/blob/main/mirror/mirror.go#L236-L238

## Proposed Changes
This changes APIs in a few places to take a peer.AddrInfo instead of a peer.ID and a multiaddr.Multiaddr.

- Updated `ingester.Sync` `dagsync.Subscriber.Sync`, APIs, and `httpsync` and `dtsync` packages to accept multiple publisher addresses

## Tests
- Updated tests to use updated APIs

## Revert Strategy
Has breaking API changes, so reverting should be avoided after other modules depend on this. Otherwise, revert is OK.

### DIFF
--- a/dagsync/httpsync/sync.go
+++ b/dagsync/httpsync/sync.go
@@ -52,16 +52,21 @@ func NewSync(lsys ipld.LinkSystem, client *http.Client, blockHook func(peer.ID, 
 }
 
 // NewSyncer creates a new Syncer to use for a single sync operation against a peer.
-func (s *Sync) NewSyncer(peerID peer.ID, peerAddr multiaddr.Multiaddr, rateLimiter *rate.Limiter) (*Syncer, error) {
-	rootURL, err := maconv.ToURL(peerAddr)
-	if err != nil {
-		return nil, err
+func (s *Sync) NewSyncer(peerID peer.ID, peerAddrs []multiaddr.Multiaddr, rateLimiter *rate.Limiter) (*Syncer, error) {
+	urls := make([]*url.URL, len(peerAddrs))
+	for i := range peerAddrs {
+		var err error
+		urls[i], err = maconv.ToURL(peerAddrs[i])
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return &Syncer{
 		peerID:      peerID,
 		rateLimiter: rateLimiter,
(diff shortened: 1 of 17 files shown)


dagsync/httpsync/sync.go
```diff
@@ -52,16 +52,21 @@ func NewSync(lsys ipld.LinkSystem, client *http.Client, blockHook func(peer.ID,
...
-	if err != nil {
-		return nil, err
+func (s *Sync) NewSyncer(peerID peer.ID, peerAddrs []multiaddr.Multiaddr, rateLimiter *rate.Limiter) (*Syncer, error) {
+	urls := make([]*url.URL, len(peerAddrs))
+	for i := range peerAddrs {
+		var err error
+		urls[i], err = maconv.ToURL(peerAddrs[i])
+		if err != nil {
+			return nil, err
+		}
```
willscott, line 62: do we care about a publisher with both an HTTP and a graphsync multiaddr? if one of the addrs isn't http this will fail for all, as written

gammazero: If a publisher has _any_ HTTP addresses, then _only_ the HTTP addresses are selected and the sync is done with the HTTP publisher using the HTTP addresses.  So, an HTTP sync will never use a non-HTTP address.

willscott: we should make sure that if someone sets an http address, and then reverts back, we don't remember the old http address and fail to revert back to graphsync pulls.

gammazero: So we will remember, and prefer the HTTP address, but if it is not being put back into the http peerstore by any new announce messages or manual syncs, then it will age out and we will go back to using whatever has been given to us since.

PR #1465, opened 2023-03-24
Increase inga dhstore timeout to 60s

ischasny:
Increase timeout to dhstore to see whether it'll provide some breathing room to the database and reduce the number of errors


gammazero (approved):
Feels wrong for any HTTP timeout to be this long, so really this is masking a bigger problem.

PR #1467, opened 2023-03-24
Update to latest core and update version

gammazero:
- Use new core with new dhstore
- Update version for new release


gammazero:
Ivan did the same thing

PR #1471, opened 2023-03-27
S3 terraform configuration for prod

gammazero:
- Configure terraform to create S3 bucker, bucket ACL, iam policy, iam assumable role, and service account.
- Configure inga pod to use new service account for accessing S3 bucket.


masih (approved):
LGTM, please remember to apply terraform before merge 👍

PR #1475, opened 2023-03-28
Deploy latest dhfind to prod and hook it up to indexstar

ischasny:
* Deploy the latest dhfind to prod with the `/cid/` endpoint enabled
* Hook it up to indexer in simulation mode


ischasny:
> Looks good to me though we should also [remove subnet](https://github.com/ipni/storetheindex/blob/main/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhfind/service.yaml#L14) mapping in internal load balancer.

Good spot - removed

PR #1482, opened 2023-03-29
Scale out dhfind in prod and hook it to indexstar

ischasny:
* Scale out dhfind in prod to verify whether that would help to reduce timeouts
* Hook dhfind to prod indexstar through loadbalancer


deploy/manifests/prod/us-east-2/tenant/storetheindex/dhfind/deployment.yaml
```diff
@@ -3,6 +3,7 @@ kind: Deployment
 metadata:
   name: dhfind
 spec:
+  replicas: 5
```
masih, line 6: Reduce to 3 and up CPU+memory?

masih (approved):

PR #1488, opened 2023-03-29
Reduce dhfind replicas down to 4

ischasny:
As we now have much better latency situation and no "context deadline exceeded" errors, reducing number of dhfind replicas down to see whether it's going to be enough to maintain the same service level.


masih (approved):

ischasny:
Close as outdated

PR #1493, opened 2023-03-30
Fix bad substitution error

ischasny:
[Context](https://github.com/ipni/storetheindex/actions/runs/4562801631/jobs/8050477342?pr=1492)
 
image

### DIFF
--- a/.github/workflows/ecr-publisher-release.yml
+++ b/.github/workflows/ecr-publisher-release.yml
@@ -25,7 +25,8 @@ jobs:
       - id: version
         name: Determine Container Tag
         run: |
-          IMAGE_TAG="$(jq -r .version version.json)"
+          VERSION="$(jq -r .version version.json)"
+          IMAGE_TAG="${VERSION#v}"
           echo "Using image tag: ${IMAGE_TAG}"
           echo "IMAGE_TAG=${IMAGE_TAG}" >> $GITHUB_ENV
       - name: AWS Login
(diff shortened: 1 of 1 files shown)


.github/workflows/ecr-publisher-release.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - id: version
         name: Determine Container Tag
         run: |
-          IMAGE_TAG="${$(jq -r .version version.json)#v}"
```
masih, line 28: This line strips `v` from the beginning of the value associated to `version` key in JSON object contained by `version.json` file.

The cause of build failure must be something else, as testing this locally on my bash session it works; Try running: ` IMAGE_TAG="${$(jq -r .version version.json)#v}" && echo $IMAGE_TAG` in root of `storetheindex`

masih (changes requested):

.github/workflows/ecr-publisher-release.yml
ischasny: Try creating a file with the following contents:
```
#!/bin/bash

IMAGE_TAG="${$(jq -r .version version.json)#v}"
echo "Using image tag: ${IMAGE_TAG}"
```

If you execute it you'd get the bad substitution error. Removing surrounding`${` resolves it. There must be some other way to remove `v` then

ischasny: Ok changed it to strip out `v` in two steps seems to be working now

masih:
Yup looks like we did it at the same time. 

This PR also fixes the build trigger to only run on merge to main: https://github.com/ipni/storetheindex/pull/1495

Feel free to either merge that  or move the changes to this PR. Whichever is easier.

PR #1503, opened 2023-03-31
Reduce Inga ingest worker count

ischasny:
Reduce Inga ingest worker count as dhstore got an inverted LSM in prod

image

image

image

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/config.json
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/config.json
@@ -85,7 +85,7 @@
     "HttpSyncRetryWaitMax": "30s",
     "HttpSyncRetryWaitMin": "1s",
     "HttpSyncTimeout": "10s",
-    "IngestWorkerCount": 20,
+    "IngestWorkerCount": 10,
     "KeepAdvertisements":true,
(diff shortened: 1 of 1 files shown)


deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/config.json
```diff
@@ -85,7 +85,7 @@
     "HttpSyncRetryWaitMax": "30s",
     "HttpSyncRetryWaitMin": "1s",
     "HttpSyncTimeout": "10s",
-    "IngestWorkerCount": 20,
+    "IngestWorkerCount": 10,
```
masih, line 88: This is a double edge sword because it impacts both fetching advertisements _and_ processing them.

Reducing it means less write parallelism but also potentially slower ad chain download.

We should engineer a tuning knob to offer individual control for each, considering one is disk IO bound and the other is network bound.

Cc @gammazero

masih (approved):

deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/config.json
gammazero: see https://github.com/ipni/storetheindex/issues/1504#issuecomment-1493220011

PR #1525, opened 2023-04-03
Revert "Re-enable dhfind in prod"

ischasny:
Reverts ipni/storetheindex#1522. Will re-enable it tomorrow as it might cause prod alerts during the night otherwise.


masih (approved):
Thanks Ivan

We should look into canary config in nginx to test this in prod without impacting upstream latency

PR #1533, opened 2023-04-05
Remove now redundant ad data PVC on `inga` and `ago`

### LINKED ISSUE #1520: Remove addata pvc from inga and ago
Inga and ago do not need to keep ad data in a separate datastore, since doing so was only done because there was not yet a way to keep this data in a car mirror.  Now that a car mirror is available there is no need for a separate ad datastore. The latest version of storetheindex does not even support that configuration.

So, the addata pvc should be removed from [inga](https://github.com/ipni/storetheindex/blob/main/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/pvc_addata.yaml) and [ago](https://github.com/ipni/storetheindex/blob/main/deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/ago/pvc_addata.yaml).

masih:
Ad data is on longer stored on a separate datastore and therefore the separate volume is no longer needed.

Fixes #1520


ischasny (approved):
Thank you for removing these

PR #1548, opened 2023-04-08
Improve ad chain ingestion concurrency

### LINKED ISSUE #1547: Improve ad chain ingestion concurrency
Advertisement chain ingestion suffers from two problems:
1. A significant bottleneck is caused by serially processing new chains, to split them up into provider work assignments. When a long chain arrives, no other chains will be processed until the current one is finished being traversed and a work assignment is created. This also puts backpressure on dagsync and may prevent it from syncing new chains that are announced.
2. Workers are stuck waiting idle when they are waiting for another worker to finish working on an assignment for the same provider. Workers should not wait for other workers to finish and should be able to work on existing work assignments for other providers.

gammazero:
## Context
See #1547

Advertisement chain ingestion suffers from two problems:
1. A significant bottleneck is caused by [serially processing](https://github.com/ipni/storetheindex/blob/main/internal/ingest/ingest.go#L940) new chains, to split them up into provider work assignments.
2. Workers are [stuck waiting](https://github.com/ipni/storetheindex/blob/main/internal/ingest/ingest.go#L1047) idle when they are waiting for another worker to finish working on an assignment for the same provider.

Fixes #1547

## Proposed Changes
The first problem, the bottleneck, is solved by putting the new advertisement chains on a queue and having the workers process the queued new chains. This queuing removes backpressure from dagsync, and having the workers do the processing allows new advertisement chains to be turned into work assignments concurrently

The second problem is solved by starting a separate goroutine that waits for any previous work assignment processing to finish, before notifying the workers about a provider's new work assignment. This way the worker itself is not waiting for other workers to finish.

### Other Fixes
This PR also removes an entire round of iteration of each advertisement chain when the work assignment is handled. This is accomplished by moving the logic to mark skipped ads into the initial new chain processing, where each ad chain is iterated to group ads by provider.

Log noise is reduced by not logging each advertisement received, but instead logging the length of the advertisement when it finishes syncing. Logging received ads can still be done by setting log level to debug.

## Tests
Uses existing tests since functionality is basically the same, but improved.

## Revert Strategy
Change is safe to revert.


gammazero:
@masih Checking this in to deploy on dev. Please review.

PR #1589, opened 2023-04-19
Workers read SyncFinishedEvents directly

gammazero:
Remove the unneeded goroutine that reads `SyncFinishedEvent`s from dagsync and puts them in another channel that the workers read. Let the workers read the events directly.

Clarified some comments and variable names to document their purpose.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -128,10 +128,6 @@ type Ingester struct {
 
 	autoSyncDone      chan struct{}
 	closePendingSyncs chan struct{}
-	cancelWorkers     context.CancelFunc
-	workersCtx        context.Context
-
-	cancelOnSyncFinished context.CancelFunc
 
 	// A map of providers currently being processed. A worker holds the lock of
 	// a provider while ingesting ads for that provider.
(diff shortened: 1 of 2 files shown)


internal/ingest/ingest.go
```diff
@@ -943,30 +941,21 @@ func (ing *Ingester) RunWorkers(n int) {
...
-
-// ingestWorker processes events from the chainQueue to create provider work
-// assignments from a raw advertisement chain, and processes those provider
-// work assignments.
-func (ing *Ingester) ingestWorker(ctx context.Context) {
+// ingestWorker processes events from dagsync.Subscriber, signaling that an
+// advertisement chain has been synced. Provider work assignments are created
+// from raw advertisement chains, and processed by ingestWorker. Work
+// assignments are processed preferentially over new advertisement chains.
+func (ing *Ingester) ingestWorker(ctx context.Context, syncFinishedEvents <-chan dagsync.SyncFinished) {
```
gammazero, line 958: Note: `syncFinishedEvents` is passed in as an argument so that if a test needs to replace `ing.syncFinishedEvents` , it does not have to wait for all the workers to start and then stop first.

masih (approved):

PR #1601, opened 2023-04-24
Add canary indexstar in dev

ischasny:
Add canary indexstar in dev to test that setup before rolling it out to prod. The rationale is - we would like to route just a portion of traffic to dhfind in the beginning.


deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/canary-deployment.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: apps/v1
```
ischasny, line 1: This file contains a merge of definitions from `base` too

deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/canary-deployment.yaml
```diff
@@ -0,0 +1,74 @@
...
+kind: Deployment
```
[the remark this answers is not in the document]
ischasny: We'd like to send only 20% of the read traffic to `dhfind` / `dhstore`. The idea is to create a new single-instance deployment of `indexstar-canary` and route these 20% of traffic to it. Regular `indexstar` won't be hooked up to `dhstore` / `dhfind` and hence won't be affected if there are any issues.

ischasny:
> Can I ask you to move canary config into a sub folder inside `indexstar` with their own `kustomization.yaml` and drop the `canary-` prefix in the manifest names plase?

Good point - moved

deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/canary-deployment.yaml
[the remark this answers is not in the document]
masih: Looks like "canary" here does A/B testing which confused me. please ignore me

deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/canary/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - deployment.yaml
+  - pdb.yaml
+  - service.yaml
+
+replicas:
+  - name: indexstar-canary
+    count: 1
+
+images:
+  - name: indexstar
```
masih, line 21: * name should match the name in `Deployment` object, which is set to `indexstar-canary`
* I recommend moving this to the parent `kustomization` so that all images are overridden in a single kustomization.
* If they should always run the same version then i would use the same image `name` for both and specify it once.

ischasny: Moved to parent

deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/canary/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+namespace: storetheindex
+
+resources:
+  - ../../../../../base/indexstar
+  - ingress.yaml
+  - pod-monitor.yaml
+
+patchesStrategicMerge:
+  - deployment.yaml
```
masih, line 12: The deployment is added as a patch but changes the name of the deployment. This is invalid in yaml patching; try building the canary by running `kustomize build`. I suspect it will fail.

Instead of renaming you can use `namePrefix` (or `nameSuffix`) like [this](https://github.com/ipni/storetheindex/blob/068b04ffbaa3a890642605698f856223e29d3a2b/deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/ago/kustomization.yaml#L11-L10). You will then need to remove all the `-canary` suffixes added manually.

masih (changes requested):

PR #1604, opened 2023-04-25
Add canary indexstar in prod

ischasny:
Enable indexstar-canary in prod that will take 33% of the traffic and includes dhfind as one of the backends


masih (approved):
🚀

PR #1612, opened 2023-04-26
Overwrite CAR in mirror on resync

gammazero:
If resyncing, then overwrite any existing CAR file in the mirror for the advertisements that are resynced.

### DIFF
--- a/carstore/carwriter.go
+++ b/carstore/carwriter.go
@@ -405,14 +405,6 @@ func decodeIPLDNode(r io.Reader, codec uint64, prototype ipld.NodePrototype) (ip
 	return nb.Build(), nil
 }
 
-// isAdvertisement checks if an IPLD node is an advertisement, by looking to
-// see if it has a "Signature" field. Additional checks may be needed if the
-// schema is extended with new types that are traversable.
-func isAdvertisement(node ipld.Node) bool {
-	indexID, _ := node.LookupByString("Signature")
-	return indexID != nil
-}
-
 // isHAMT checks if the given IPLD node is a HAMT root node by looking for a
 // field named "hamt".
 //
--- a/carstore/carwriter.go
+++ b/carstore/carwriter.go
@@ -116,6 +116,8 @@ func (cw *CarWriter) write(ctx context.Context, adCid cid.Cid, ad schema.Adverti
 		}
 		// OK, car file does not exist.
 	} else if !overWrite {
+		// If overWrite is false then only do datastore cleanup without
+		// overwriting car file.
 		if err = cw.removeAdData(roots); err != nil {
 			log.Errorw("Cannot remove advertisement data from datastore", "err", err)
 		}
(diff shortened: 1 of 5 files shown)


willscott:
is it cheaper to stat and skip the write if it would be exactly the same? is there a way to know if it would be exactly the same before writing?

masih:
I am curious if we can do something with S3 `ETag`s, where somewhere we store a mapping of ad/entry CID to its S3 digest. That would then allow us to infer whether the block associated to a CID is already there in its complete form or not? 
This would be S3 specific, though. 

It would be nice to be able to do this agnostic of the mirror backing store.

gammazero:
@willscott Resync rarely happens, and is a manual administrative action. It is done to redo a sync of ads and their data, presumably because we think that some things will be different.  For example, maybe some bug caused the indexer to not index the last multihash or record it in a car file, and we want to rewrite those things. Unfortunately, there is no way to tell if the car file we are going to write is the exact same as what was written, since we do not keep hashes of the files themselves, and reading+hashing will not be much better than just writing a new one.

@masih Relying on S3 tags starts tying us to S3 functionality that goes beyond reading and writing files, and for a resync (which AFAIK is almost never done) it is just not worth trying to optimize this.

The only reason I even made this PR, was because I was concerned that something had been indexed something incorrectly and I wanted to be able to correct it if that was the case.  Turns out it was not, but having the ability to correct the car mirror when resyncing seemed like the correct behavior.

masih:
Thanks Andrew for explaining. 
Rewriting it every time would certainly work. It would just be nice to avoid doing that if there is a way to infer that the bytes stored are identical to the ones we are going to redownload. 

It seems likely that each mirror backing store is going to have its own efficient way of doing this and maybe is a concern that the backing store implementation can encapsulate? i.e. atomically write a given CAR file.

carstore/carwriter.go
```diff
@@ -115,7 +115,7 @@ func (cw *CarWriter) write(ctx context.Context, adCid cid.Cid, ad schema.Adverti
 			return nil, err
 		}
 		// OK, car file does not exist.
-	} else {
+	} else if !overWrite {
```
ischasny, line 118: Should the flag be the other way around? I.e. if `overwrite` is `true` then remove ad data?

gammazero: No. The call to `removeAdData` removes the ad data from the data store, that was put there by datatransfer/graphsync, instead of writing the car file and then removing that data after writing the car file.

In otherwords, if overWrite is false then only do datastore cleanup without overwriting car file.

gammazero: Added comment to clarify.

gammazero:
@masih Whether this is up to the backing store or not depends on what we mean by "the same data". If two CAR files contain the same data, but are compressed differently (suppose a different compression level) or maybe have an embedded timestamp, then these will appear different to the backing store (after compression) and the same to the indexer (before compression). Since there may be a reason we want to overwrite the existing file, beyond what the indexer knows, this needs to be done by the writer to the backing store. 

Regardless of whether we overwrite identical data or not, this PR is still needed to tell the car store to overwrite an existing file, diff check aside. Otherwise, the wwrite request is ignored if the underlying file exists at all.

ischasny (approved):

PR #1615, opened 2023-04-26
Expose dhfind for external access in prod

ischasny:
Exposing in prod only where it runs in non-simulation mode to verify metrics.


masih (approved):
Do the same for `dev` to keep deployments consistent?

PR #1622, opened 2023-04-26
Handle trailing slash with find batch request

gammazero:
The `findBatch` handler is called when there is no slash following the `/multihash` endpoint. If there is a slash, then the `findMultihash` handler is called and expects a multihash in the URL path. Some clients may add a trailing slash, even when they mean to call the `findBatch` handler. Fix this by calling `findBatch` if `findMultihash` is called and the request method is `POST`.

### DIFF
--- a/server/find/http/server.go
+++ b/server/find/http/server.go
@@ -170,6 +170,10 @@ func (s *Server) findCid(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *Server) findMultihash(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodPost {
+		s.findBatch(w, r)
+		return
+	}
 	enableCors(w)
 
 	if !httpserver.MethodOK(w, r, http.MethodGet) {


willscott:
should it be if it's post *and* there's no multihash in the url?

server/find/http/server.go
```diff
@@ -170,6 +170,10 @@ func (s *Server) findCid(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *Server) findMultihash(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodPost {
```
masih, line 173: This handler is associated with `/multihash/` pattern, i.e. `/mutlihash` subtree. 

This should immediately return `MethodNotAllowed` IMHO, because the presence of a tailing `/` means the client is in subtree teritory which does not support `POST` method and only supports `GET`.

masih:
Would be good to check for `MethodNotAllowed` elsewhere in the code. It can be returned [here](https://github.com/ipni/storetheindex/blob/ecb230aeb36da9a7df3fea8f76b4facf0ddc11c1/server/find/http/server.go#L200) for example.

gammazero:
@masih The [`MethodOK`](https://github.com/ipni/storetheindex/blob/ecb230aeb36da9a7df3fea8f76b4facf0ddc11c1/server/find/http/server.go#L199) method already does the correct thing, [here](https://github.com/ipni/storetheindex/blob/main/internal/httpserver/util.go#L20), by returning `StatusMethodNotAllowed`

gammazero:
Closing, because adding the trailing slash when sending a POST request to `/multihash` is a client error.

masih:
Ah i see; I missed methodok thanks Andrew

PR #1624, opened 2023-04-27
Increase canary traffic to 20%

ischasny:
Encrypted store has been running ok against 10% for the last 24 hours. Increasing that number to 20%.


willscott (approved):
🚀

PR #1626, opened 2023-04-27
Add dhstore-helga

ischasny:
Add a new instance `dhstore-helga` in prod. It's going to be used for writes while the existing one will serve reads only.

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
@@ -46,7 +46,7 @@ spec:
                   - key: topology.kubernetes.io/zone
                     operator: In
                     values:
-                      - us-east-2c    
+                      - us-east-2a
       volumes:
         - name: data
           persistentVolumeClaim:
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
@@ -1,0 +1,59 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: dhstore
+spec:
+  selector:
+    matchLabels:
+      app: dhstore-helga
+  template:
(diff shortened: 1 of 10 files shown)


deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: node.kubernetes.io/instance-type
+                    operator: In
+                    values:
+                      - r5n.2xlarge
+                  - key: topology.kubernetes.io/zone
+                    operator: In
+                    values:
+                      - us-east-2c
```
masih, line 49: Checking this is different from the current AZ used by the existing dhstore?

deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/pvc.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: dhstore-data
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 20Ti
```
masih, line 10: Max is 16 right?

Regardless, let's start with a small number and increase as needed.

deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
ischasny: I think it's the same. Should it be different?

deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/pvc.yaml
ischasny: Good point - will decrease.

ischasny: Done

masih (approved):

PR #1631, opened 2023-04-27
Revert "Revert "Increase prod dhstore PVC to 20Ti""

ischasny:
Setting the volume size back to 20Ti to fix reconciliation

Reverts ipni/storetheindex#1630


masih (approved):
revert ^2 💣

PR #1636, opened 2023-04-27
Reduce canary traffic

ischasny:
Reduce canary traffic as indexers struggle to respond onto provider requests. Needs to be bumped back up when provider cache is implemented.


masih (approved):

ischasny:
Close as not needed

PR #1665, opened 2023-05-02
[do not merge] Indexer configurable to process removal advertisements only

gammazero:
Requires that the config setting `config.Ingest.RmOnly` be set to `true`.


gammazero:
We are not going to do this afterall.

PR #1695, opened 2023-05-06
Store temporary ad data in separate datastore

### LINKED ISSUE #1380: Remove old FSM state from datastore after namespace wrapping
[This](https://github.com/ipni/storetheindex/pull/1377) PR wraps the datastore passed to data transfer in a namespace to avid FSM migration of data transfer 2 and get the crashing nodes to come back up.

As a result, the old FSM data remains at the root of datastore that needs to be cherry picked and deleted depening on storetheindex specific keys that are also stored in the same datastore.

gammazero:
- Configure indexers to use separate datastore for ad and fsm data.
- Remove old previous version data-transfer fsm records
- Remove current version data-transfer fsm records
- Remove advertisement and entries data blocks
- Record new datastore version string

Fixes #1380

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -752,6 +752,9 @@ func createDatastore(ctx context.Context, dir, dsType string) (datastore.Batchin
 }
 
 func updateDatastore(ctx context.Context, ds datastore.Batching) error {
+	// updateBatchSize is the number of records to update at a time.
+	const updateBatchSize = 500000
+
 	dsVerKey := datastore.NewKey(dsInfoPrefix + dsVersionKey)
 	curVerData, err := ds.Get(ctx, dsVerKey)
 	if err != nil && !errors.Is(err, datastore.ErrNotFound) {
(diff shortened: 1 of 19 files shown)


willscott:
why would we want a separate key namespace in the same datastore vs a separate datastore?
a different datastore means you could tune the temporary ad data in a memory cache or otherwise give it different iops peroperties than the main datastore, which seems generally desirable.

gammazero:
@willscott That is a good point as the temporary block data may have very different access patterns.

command/daemon.go
```diff
@@ -729,3 +752,83 @@ func createDatastore(cfg config.Datastore) (datastore.Batching, error) {
...
+			if err = batch.Commit(ctx); err != nil {
+				return fmt.Errorf("cannot commit to datastore: %w", err)
+			}
+			log.Infow("Datastore update removed data", "dtSessions", dtKeyCount, "cids", cidCount)
+		}
+		if result.Error != nil {
+			return fmt.Errorf("cannot read query result from datastore: %w", result.Error)
+		}
+		ent := result.Entry
+		key := ent.Key[1:]
```
willscott, line 802: is it possible for `ent.Key` to be empty / for this to panic?

gammazero: It is impossible, at least with leveldb:
https://github.com/ipfs/go-ds-leveldb/blob/master/datastore.go#L135-L136

Will handle this case anyway just in case there is some implementation that does have same behavior.

command/daemon.go
```diff
@@ -729,3 +756,83 @@ func createDatastore(cfg config.Datastore) (datastore.Batching, error) {
...
+	defer results.Close()
+
+	batch, err := ds.Batch(ctx)
+	if err != nil {
+		return fmt.Errorf("cannot create datastore batch: %w", err)
+	}
+
+	var cidCount, dtKeyCount, writeCount int
+	for result := range results.Next() {
+		if writeCount >= 500000 {
```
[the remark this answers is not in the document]
gammazero: Done

command/daemon.go
gammazero: Done

willscott (approved):

PR #1712, opened 2023-05-10
Fix poll override causing provider deactivation

gammazero:
There are really two defects here:
1. A default for `deactivateAfter` is not being set correctly when there is an override.
2. Once an override is encountered, the poll keeps the value of the override for all following providers. It needs to be restored to the previous value for the next provider.

This PR fixes both of these issues.


masih (approved):
Nice catch thanks @gammazero 👍

ischasny (approved):

PR #1743, opened 2023-05-16
Automate image update for `lassie-event-recorder`

masih:
Set image repo and selection policy for lassie event recorder in order to automate the image image update and deployment.

This change will eliminate the need to manually restart the event recorder pods in order to reflect the changes of the latest built image.

### DIFF
--- a/deploy/manifests/dev/us-east-2/cluster/autoretrieve/event-recorder-cd.yaml
+++ b/deploy/manifests/dev/us-east-2/cluster/autoretrieve/event-recorder-cd.yaml
@@ -1,0 +1,22 @@
+apiVersion: image.toolkit.fluxcd.io/v1beta1
+kind: ImageRepository
+metadata:
+  name: lassie-event-recorder
+spec:
+  interval: 5m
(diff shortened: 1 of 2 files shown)


willscott:
and this needs to live here vesus in https://github.com/filecoin-project/autoretrieve-deploy because this is where we have the image CD stuff happening?

masih:
They live here because this repo contains the "cluster-level" manifests, i.e. admin stuff; the manifests in the "cluster" folder affect the entire cluster and need to be administered separately.

The "tenant" manifests then live in a separate repo , which is `autoretrieve-deploy`, hence the setter changes [here](https://github.com/filecoin-project/autoretrieve-deploy/pull/65).

`lassie-event-recorder` is using the `autioretrieve` tenancy to run its stuff ( which was the fastest way to get it running at a time due to the cost issues of its previous version). Once `autoretrieve` is no longer needed, my plan is to simply rename this repo to be called `lassie-event-recorder`.

willscott (approved):

PR #1748, opened 2023-05-17
Define initial dhstore cluster in `dev`

masih:
Define manifests that configures a cluster of dhstore instances fronted by envoy front proxy. The envoy front proxy is configured to distribute the requests using Maglev consistent hashing algorithm

### DIFF
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/base/dhstore/deployment.yaml
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/base/dhstore/deployment.yaml
@@ -1,0 +1,46 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: dhstore
+spec:
+  template:
+    spec:
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: topology.kubernetes.io/zone
+          whenUnsatisfiable: ScheduleAnyway
+      containers:
+        - name: dhstore
+          args:
+            - '--storePath=/data'
+          volumeMounts:
+            - name: data
+              mountPath: /data
+          resources:
+            limits:
+              cpu: "2"
+              memory: 30Gi
+            requests:
+              cpu: "2"
+              memory: 30Gi
+          ports:
+            - containerPort: 40081
+              name: metrics
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: node.kubernetes.io/instance-type
+                    operator: In
+                    values:
+                      - r5n.2xlarge
+      volumes:
+        - name: data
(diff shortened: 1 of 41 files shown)


deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/base/dhstore/deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+    spec:
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: topology.kubernetes.io/zone
+          whenUnsatisfiable: ScheduleAnyway
+      containers:
+        - name: dhstore
+          args:
+            - '--storePath=/data'
+            - '--disableWAL'
```
ischasny, line 16: Do we want to keep WAL on? We disabled it originally to improve performance but that might not be the case for the cluster anymore.

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/base/dhstore/deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+          volumeMounts:
+            - name: data
+              mountPath: /data
+          resources:
+            limits:
+              cpu: "6"
+              memory: 60Gi
```
ischasny, line 23: Do we want to lower down the spec for dhstores?

masih: what instance type do you suggest instead of the current one?

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/base/dhstore/pvc.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: dhstore-data
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Ti
```
ischasny, line 10: 1Ti might be not enough (even though we do have 10 instances). Right now we are already at 18(ish)Ti  in prod. Shall we bump that up to 2Ti at least?

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/base/dhstore/deployment.yaml
ischasny: I'd probably go down to 2 cpus initially. `dhstore-helga` now runs at 75% CPU utilisation.

image

WRT RAM - I *think* we can try lowering it down to 10-20Gi, given that instances are going to receive just a 1/10th of the traffic.

ischasny (approved):

PR #1751, opened 2023-05-18
Change ExtendedProviders behaviour to 'replace' instead of 'add'

ischasny:
Removed ExtendedProviders additive behaviour. Each ExtendedProviders advertisement will have to have a full set of ExtendedProviders from now on (no deltas). Additive behaviour makes updates and removals harder and is generally counterintuitive.

Context: https://filecoinproject.slack.com/archives/C02T827T9N0/p1684402768677939

### DIFF
--- a/internal/ingest/ingest_test.go
+++ b/internal/ingest/ingest_test.go
@@ -904,7 +904,7 @@ func testSyncWithExtendedProviders(t *testing.T,
 	testFunc(privKey, pubKey, providerID, reg, lsys, pubHost, ingester, pub)
 }
 
-func TestSyncWithExtendedProviders(t *testing.T) {
+func TestSyncWithContextualExtendedProviders(t *testing.T) {
 	testSyncWithExtendedProviders(t, func(privKey crypto.PrivKey,
 		pubKey crypto.PubKey,
 		providerID peer.ID,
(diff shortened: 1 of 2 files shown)


willscott:
Should we do this in the provider first before changing the semantics on the server?

ischasny:
> Should we do this in the provider first before changing the semantics on the server?

What do you mean in the provider? In index provider it's stateless, i.e. whatever you put into extended providers there it'd be sent in ad.

willscott:
as in make sure the boost-index-provider interaction is using replace rather than add semantics.

I _think_ they may already be using replace semantics, for what it's worth

ischasny:
> as in make sure the boost-index-provider interaction is using replace rather than add semantics.
> 
> 
> 
> I _think_ they may already be using replace semantics, for what it's worth

Good point - asked the already

ischasny:
@willscott confirmed - all is good on the boost side

willscott:
I don't really understand how this change in the ingest link system changes the behavior from additive to replacing - i don't see any `append` or other thing that would have concatenated before that's being removed.

ischasny:
> I don't really understand how this change in the ingest link system changes the behavior from additive to replacing - i don't see any `append` or other thing that would have concatenated before that's being removed.

Before it used to [load](https://github.com/ipni/storetheindex/pull/1751/files#diff-9224e87eb2202ee2db2e74a111cf0cb420d9a05619897c77d8dadd57cbaa3af3L213) existing extended providers from the registry and append to those, but now it creates an empty extended providers list every time.

willscott (approved):

PR #1752, opened 2023-05-18
Temporarily pause inga

ischasny:
To allow dhstore to catch up with compaction


masih (commented):
As a learning exercise let's bump the CPU of `dhstore` in prod to 7 before pausing writes. I suspect the root cause of this (aside from write path optimistion) is CPU contention.

ischasny:
> As a learning exercise let's bump the CPU of `dhstore` in prod to 7 before pausing writes. I suspect the root cause of this (aside from write path optimistion) is CPU contention.

See https://github.com/ipni/storetheindex/pull/1753

gammazero (approved):

ischasny:
Closing as not relevant anymore.

PR #1753, opened 2023-05-18
Bump dhstore CPU to 7 in prod

ischasny:


deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
```diff
@@ -26,10 +26,10 @@ spec:
               mountPath: /data
           resources:
             limits:
-              cpu: "6"
+              cpu: "7"
               memory: 60Gi
             requests:
-              cpu: "6"
+              cpu: "7"
```
masih, line 32: Please do not increase `helga`'s CPU until we confirm that this is a reasonable change. We may need to up workers to beefier CPU optimised types.

masih (approved):

PR #1774, opened 2023-05-19
Gracefully skip not found content from HTTP publishers

### LINKED ISSUE #1771: Inconsistent error handling between HTTP publisher and DT publishers
The ingestion safely skips over ads the entries for which is not found by checking for specific error messages. The error messages work fine for DT publishers since the original error message from provider side bubbles up to the indexers.

The same is not true for HTTP publishers, however. Two issues:
1. The [error](https://github.com/ipni/go-libipni/blob/098f9147b2d29c27649425209b2a1a49f6bd8ad9/dagsync/httpsync/publisher.go#L264) returned by the HTTP publisher does not include the original error message.
2. storetheindex uses a HTTP retry client which does not expose the original error content, and instead returns an error like ` Get \"http://[::]:24008/baguqeeqqbiezezlpw2nvevijrd7ez7tv4q\": GET http://[::]:24008/baguqeeqqbiezezlpw2nvevijrd7ez7tv4q giving up after 5 attempt(s)` which will not be interpreted as a skippable error by the ingestion pipeline.

To fix this:
* bubble up the error message from client side up to the server side so that not found entries are skipped safely
* Use 404 as a code in HTTP side instead to interpret content not found.

masih:
Bubble up the fix in HTTP syncer so that 404s from HTTP publisher are surfaced appropriately. Improve error checking logic to break out early if the error is of kind `ipld.ErrNotExists`.

Relates to:
 - https://github.com/ipni/go-libipni/pull/39

Fixes #1771

### DIFF
--- a/internal/ingest/linksystem.go
+++ b/internal/ingest/linksystem.go
@@ -368,6 +368,7 @@ func (ing *Ingester) ingestAd(ctx context.Context, publisherID peer.ID, adCid ci
 		msg := err.Error()
 		switch {
 		case
+			errors.Is(err, ipld.ErrNotExists{}),
 			strings.Contains(msg, "content not found"),
 			strings.Contains(msg, "graphsync request failed to complete: skip"):
 			return adIngestError{adIngestContentNotFound, wrappedErr}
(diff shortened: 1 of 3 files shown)


internal/ingest/linksystem.go
```diff
@@ -566,7 +567,7 @@ func (ing *Ingester) ingestEntriesFromPublisher(ctx context.Context, ad schema.A
 				return mhCount, err
 			}
 			wrappedErr := fmt.Errorf("failed to sync entries: %w", err)
-			if strings.Contains(err.Error(), "content not found") {
+			if errors.Is(err, ipld.ErrNotExists{}) || strings.Contains(err.Error(), "content not found") {
```
gammazero, line 570: It might be worth converting the "content not found" error from graphsync into an `ipld.ErrNotExists{}` error in go-libipni.

gammazero: This should convert the "content not found" error: https://github.com/ipni/go-libipni/pull/40

gammazero (approved):

PR #1775, opened 2023-05-19
Add pod monitor configs for envoy cluster

gammazero:


deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/envoy/pod-monitor.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: envoy
+  namespaceSelector:
+    matchNames:
+      - storetheindex
+  podMetricsEndpoints:
+    - path: /metrics
+      port: metrics
```
masih, line 16: envoy port is called `admin` i think

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/envoy/pod-monitor.yaml
```diff
@@ -0,0 +1,16 @@
...
+    app: envoy
+spec:
+  selector:
+    matchLabels:
+      app: envoy
+  namespaceSelector:
+    matchNames:
+      - storetheindex
+  podMetricsEndpoints:
+    - path: /metrics
```
masih, line 15: The path at which metrics are exposed in envoy is also different. This value should be the same as the one in `ServiceMonitor`

gammazero: I did copy this from [service-monitor.yaml](https://github.com/ipni/storetheindex/blob/main/deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/instances/service-monitor.yaml), I think you mean I need to use what is in [envoy-monitor.yaml](https://github.com/ipni/storetheindex/blob/main/deploy/manifests/dev/us-east-2/tenant/storetheindex/dhstore-cluster/envoy/envoy-monitor.yaml)

gammazero: Done

masih (commented):
spec looks right thanks Andrew 👍 They also need to be added to their `kustomization.yaml` under `resources` to be included in the generation.

masih (approved):

gammazero:
Not using dhstore cluster

PR #1800, opened 2023-05-24
Take snapshots of read-only dhstores

ischasny:
Snapshots will be used to set up volumes of smaller capacity, as `dhstore` and `dhstore-helga` are 60%(ish) utilised after compaction


masih:
Snapshot can't be applied to volumes of smaller size I don't think.

ischasny:
> Snapshot can't be applied to volumes of smaller size I don't think.

That's a bummer. Ok closing then.

PR #1802, opened 2023-05-24
Add new overflow dhstore and dhfind for ago

gammazero:
New dhstore, dhstore-ago2, to handle overflow from first dhstore. Also, deploying dhfind-ago2 to go with it.


deploy/manifests/dev/us-east-2/tenant/storetheindex/dhfind-ago2/service.yaml
```diff
@@ -0,0 +1,20 @@
...
+# See: https://github.com/ipni/dhfind
+kind: Service
+apiVersion: v1
+metadata:
+  name: dhfind
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-internal: "true"
+    service.beta.kubernetes.io/aws-load-balancer-scheme: internal
+    service.beta.kubernetes.io/aws-load-balancer-type: nlb
+    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
```
gammazero, line 13: Do I need `aws-load-balancer-subnets` like prod does, [here](https://github.com/ipni/storetheindex/blob/main/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-porvy/internal-service.yaml#L14)

ischasny (approved):
Thanks Andrew

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhfind-ago2/deployment.yaml
```diff
@@ -0,0 +1,30 @@
...
+    spec:
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: topology.kubernetes.io/zone
+          whenUnsatisfiable: ScheduleAnyway
+      containers:
+        - name: dhfind
+          args:
+            - '--dhstoreAddr=http://dhstore-ago2.internal.dev.cid.contact/'
+            - '--stiAddr=http://ago-indexer:3000/'
```
gammazero, line 23: Would it be better for this to go to an indexer or to indexstar?

deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/deployment.yaml
```diff
@@ -19,8 +19,8 @@ spec:
             - '--backends=http://ber-indexer:3000/'
             - '--backends=http://cali-indexer:3000/'
             - '--backends=http://ago-indexer:3000/'
```
gammazero, line 21: If ago is not running a valuestore (only writing to dhstore), then I think I should remove it from the indexstar config. Is that correct?

masih: Right. I think this should be the corresponding dhfind endpoint if not included already.

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhfind-ago2/deployment.yaml
masih: this is used for fetching providers info right? So pointing that to indexstar would result in circular query across instances?

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhfind-ago2/service.yaml
masih: nope

deploy/manifests/dev/us-east-2/tenant/storetheindex/dhfind-ago2/deployment.yaml
gammazero: The `--stiAddr` is only used for fetching provider info. So, I think it makes most sense to point this at the indexer instance that is associated with this dhfind instance.

ischasny: Btw, this can be pointed at `indexstar`. There won't be a circular dependency as the `stiAddr`  endpoint is used for provider lookups only. Pointing that directly at the indexer instance saves up some latency on provider lookups at the cost of not being able to do lookups at all when this indexer instance is down. However such performance boost is negligible because provider records are cached.

So the bottom line is - I recommend pointing `stiAddr` to `indexstar` next time.

PR #1803, opened 2023-05-24
Remove support for pogreb as valuestore

gammazero:
Pogreb is not used by anyone so support is being removed.


gammazero:
@ischasny Error returned here: https://github.com/ipni/storetheindex/blob/main/command/daemon.go#L591
then here https://github.com/ipni/storetheindex/blob/main/command/daemon.go#L105
And is printed.

PR #1805, opened 2023-05-25
New dev indexer-dhstore-dhfind set

gammazero:
Create a new set of indexers in dev. Each indexer has a dedicated dhstore and a dedicated dhfind. Their names are matched, for example: `alva-indexer`, `dhstore-alva`, `dhfind-alva`

The assigner configuration has changed to use the new dev indexers as its indexer pool.


deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/cora/pvc_data.yaml
```diff
@@ -0,0 +1,15 @@
...
+    app: indexer-single
+    app.kubernetes.io/managed-by: kustomization
+    app.kubernetes.io/part-of: storetheindex
+  name: data
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Ti
```
masih, line 14: This volume is used for provider info only right? 

Would it make sense to look at its utilisation in other nodes and potentially reduce this? We can always grow it later but shrinking it is non trivial and would need data migration.

masih (approved):
@gammazero the infra ninja 🥷

ischasny (approved):

deploy/manifests/dev/us-east-2/tenant/storetheindex/instances/cora/pvc_data.yaml
gammazero: The data stores do grow with recording all the ads ingested as well as keeping the mh counts for all contextIDs. ... and of course all that wonderful data-transfer FSM state.  On ago this is currently 24G... so would 100G be good?

masih: `100Gi` sounds good 👍

gammazero:
Closing in favor of new indexers deployed for use with FDB.  If we want a set indexer-dhstore-dhfind nodes later, that can be done in a new PR.

PR #1806, opened 2023-05-25
Gen ident utilty

gammazero:
Add a utility to make it easy to create indexer identity files.  Here is the usage message:
```
NAME:
    gen-identity

USAGE:
    go run gen-identity [command options]

DESCRIPTION:
    Generate an encrypted identity file, identity.key.encrypted, and output the associated peer ID.
    Run gen-identity from a directory with a .sops.yaml file to generate the encrypted identity.

    Example:
        cd deploy/manifests/dev/us-east-2/tenant/storetheindex
        go run ../../../../../../scripts/gen-identity
        mv identity.key.encrypted instances/my-indexer/

OPTIONS:
    -help, -h       Show help
```


scripts/gen-identity.go
```diff
@@ -0,0 +1,124 @@
+package main
```
masih, line 1: we probably want to move this to its own package e.g. `scripts/gen-identity/main.go` because otherwise we won't be able to have more than one `main` func in `scripts` folder.

masih (approved):

scripts/gen-identity.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	pkFileName, pid, err := genPrivateKeyFile()
+	if err != nil {
+		return err
+	}
+	defer os.Remove(pkFileName)
+
+	// sops -e identity.key > identity.key.encrypted
+	cmd := exec.Command("sops", "-e", pkFileName)
```
masih, line 66: I advise against this TBH. It is easy enough to pipe it to sops and having it here leaks infra choices. So my vote is for something like [this](https://github.com/ipni/storetheindex/pull/341).

scripts/gen-identity.go
gammazero: > won't be able to have more than one main func in scripts folder
I am not sure I understand.  I can have any number of scripts in the same folder, all with a `func main() { }` in them. When I use `go run script1.go` it does not care that `script2.go` and `script3.go` in the same directory also have a `func main() { }`.

This would only be a problem if we were trying to compile multiple files in the same directory that had a main, and I do not think we were doing this. These should be left as `.go` files and run as scripts using `go run`.

masih: I usually run go run ./scripts/... from project root. But that's just me

gammazero: Yes, that is what I do as well.  I still do not understand what you mean by:
> won't be able to have more than one main func in scripts folder

If I have multiple `.go` files in the `scripts` directory, all with their own `func main() { }`, I can run them just fine. Do this from the root directory:
```shell
$ cp scripts/peer_id_from_priv_key/main.go scripts/peer-id-from-priv-key.go
$ go run scripts/gen-identity.go | go run scripts/peer-id-from-priv-key.go 
Peer ID for private key: 12D3KooWBh6KsVDPu4s67gVWfDNRY7oMCxLTeV8ompce8Cya3V6k
12D3KooWBh6KsVDPu4s67gVWfDNRY7oMCxLTeV8ompce8Cya3V6k
```

masih: right you are; i recall hitting an issue with having multiple `main` functions in a single `main` package and I cannot recall what it was. Please ignore me. Thanks Andrew

PR #1815, opened 2023-05-26
Add dhstores as indexstar backends

gammazero:
Needed to scatter dh queries over dhstore instances.


deploy/manifests/dev/us-east-2/tenant/storetheindex/indexstar/deployment.yaml
```diff
@@ -18,8 +18,11 @@ spec:
             # TLS handshake overhead.
             - '--backends=http://ber-indexer:3000/'
             - '--backends=http://cali-indexer:3000/'
+            - '--backends=http://ago-indexer:3000/'
             - '--backends=http://dhfind.internal.dev.cid.contact/'
+            - '--backends=http://dhstore.internal.dev.cid.contact/' 
             - '--backends=http://dhfind-ago2.internal.dev.cid.contact/'
+            - '--backends=http://dhstore-ago2.internal.dev.cid.contact/'
```
masih, line 25: Let's check that indexstar does not forward PUTs and POSTs to the backends

Aside from that I think dhstore returns bad request on non-double hashed requests, and should be fronted with dhfind usually. Time to combine the two?

gammazero:
Already done in another PR.

PR #1832, opened 2023-05-30
Bump up dhstore-helga CPU

ischasny:
Bump up dhstore-helga CPU as it seems to be the largest contributor to high read latencies

image

image

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
@@ -26,10 +26,10 @@ spec:
               mountPath: /data
           resources:
             limits:
-              cpu: "3"
+              cpu: "4"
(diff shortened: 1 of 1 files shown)


ischasny:
> How did they do on memory utilisation?

Pebble always consumes 100% regardless of how much you give it. hard to say. I think it should be ok as no writes are coming in, only reads. Lets bump up CPUs and see

deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-helga/deployment.yaml
```diff
@@ -26,10 +26,10 @@ spec:
               mountPath: /data
           resources:
             limits:
-              cpu: "2"
+              cpu: "4"
```
masih, line 29: Let's bump these to 7. dhstore runs on `r5n.2xlarge` and those nodes have 8 CPUs.

masih (approved):

PR #1837, opened 2023-05-30
Disconnect old indexers

ischasny:
Also connect `inga` to indexestar so that it can serve the static page and `/providers` endpoint


gammazero (approved):
Or would it be better to make inga the stiAddr for dhfind instances?

ischasny:
> Or would it be better to make inga the stiAddr for dhfind instances?

We need an indexer connected to indexstar anyway - so that it can serve the static page and `/providers` requests

Issue #1842, opened 2023-05-31
Remove libp2p find service


gammazero:
No client is currently using the libp2p to access an indexer's find service. Remove support for the libp2p find service to have less code to update and maintain.

gammazero:
Fixed by #1838

PR #1843, opened 2023-05-31
Remove libp2p find server

gammazero:
Removed unused libp2p find server.

Fixes issue #1842

This PR also fixes a problem of not responding correctly to HTTP method OPTIONS requests.


masih (approved):
Looks good apart from minor checks failing

PR #1868, opened 2023-06-01
Deploy new nodes and point to stateless dhstore

masih:
Deploy the new nodes on dev, spread across AZs and point them to the FDB-backed dhstore.


gammazero (approved):
You may want to rebase on main and update image to `20230601233715-9d2f7ccbd13dd304c0a3843969b9ddb8210ae6f1` first. That way they will start running with the new datastore configuration. If not, which is OK, then they will need to go through a datastore update later.  Might slow startup by a minute or so.

ischasny (approved):

PR #1876, opened 2023-06-02
Connect dhstores to indexstar in dev

ischasny:


masih:
what is the rationale for this? Does this not already happen via dhfind backends corresponding to those dhstores?

ischasny:
> what is the rationale for this? Does this not already happen via dhfind backends corresponding to those dhstores?

DHstores used to be connected to indexstar but got removed probably by mistake in one of the latest PRs. DHFinds don't serve EncryptedValueKeys.

masih:
I see thanks

masih (approved):
looking forward to combining dhstore and dhfind

PR #1877, opened 2023-06-02
Deploy new assigner for FDB-backed indexers

masih:
Deploy a new assigner for the FoundationDB backed indexer nodes, that is daisy-chained to the front-facing assigner to propagate announcements through.

### DIFF
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/assigner-fdb-backed/config.json
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/assigner-fdb-backed/config.json
@@ -8,14 +8,19 @@
     "FilterIPs": false,
     "IndexerPool": [
       {
-        "AdminURL": "http://ber-indexer:3002",
-        "FindURL": "http://ber-indexer:3000",
-        "IngestURL": "http://ber-indexer:3001"
+        "AdminURL": "http://alva-indexer:3002",
+        "FindURL": "http://alva-indexer:3000",
+        "IngestURL": "http://alva-indexer:3001"
       },
       {
-        "AdminURL": "http://cali-indexer:3002",
-        "FindURL": "http://cali-indexer:3000",
-        "IngestURL": "http://cali-indexer:3001"
+        "AdminURL": "http://bria-indexer:3002",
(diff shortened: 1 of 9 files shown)


ischasny (approved):

gammazero:
I do not think we need a new assigner.

If we reconfigure the current assigner to only have the new indexers in its pool, the old indexers will keep their current assignments but will not get any new ones. The assigner will continue to republish direct announcements to all the indexers over pubsub. We will not be able to do handoff of a frozen indexer if ber or cali become frozen, but we do not care about ever doing that.

Since the assigner is stateless, we can always revert its config without any problem. Then we can create this new assigner if we think it is necessary, also without any problems.

gammazero (commented):
I think we can use the current assigner instead and do not need a new one.

If we reconfigure the current assigner to only have the new indexers in its pool, the old indexers will keep their current assignments but will not get any new ones. The assigner will continue to republish direct announcements to all the indexers over pubsub. We will not be able to do handoff of a frozen indexer if ber or cali become frozen, but we do not care about ever doing that.

Since the assigner is stateless, we can always revert its config without any problem. Then we can create this new assigner if we think it is necessary, also without any problems.

masih:
> I think we can use the current assigner 

That would halt assignments to the curren unencrypted nodes, which means their state becomes out of date while the new cluster is catching up. This then means lookups on dev would start returning incorrect results until the new cluster is caught. Right?

gammazero:
It would halt the assignment of new publishers, but would not stop indexing with the current set of assigned publishers.
If the intent is to remove ber and cali as soon, then no need for a new assigner. Aside from the added deployment item to watch, there is no problem with this.

gammazero (approved):

masih:
> My bigger concern was that direct http announcements will now need to go to 2 assigners.

That should be covered since I peered the new assigner with the old one. But you pointed out a great point: old nodes continue to ingest from current assignments. It's just that they won't pick up new providers.

And I think that is fine, so going to remove the new assigner. Thanks, Andrew.

masih:
Closing in favour of #1879

PR #1879, opened 2023-06-02
Config assigner to use new indexers for FDB valuestore

gammazero:
**Not needed if a second assigner #1877 is preferred.**

This configures the existing assigner to assign to only the new indexers.

Notice that the old indexers are still set as peers to assigner. This ensures that rebublisher direct announcements are sent to them.


masih (approved):
Thanks Andrew, let's get this rolling 👍

masih (approved):
Thanks Andrew, let's get this rolling 👍

masih (approved):
Thanks Andrew, let's get this rolling 👍

PR #1889, opened 2023-06-06
Create FDB CRDs in prod

ischasny:
In preparation for FDB deployment


deploy/manifests/prod/us-east-2/cluster/kustomization.yaml
```diff
@@ -13,3 +13,5 @@ resources:
   - aws-ebs-csi-driver
   - promtail
   - external-snapshotter
+  - ../../../base/k6-operator
```
masih, line 16: why k6?

ischasny: Yes, removed

PR #1891, opened 2023-06-06
Expose dev assigner p2p port for peering

gammazero:
Need to peer prod assigner dev assigner to forward direct announce messages.


deploy/manifests/dev/us-east-2/tenant/storetheindex/assigner/service-external.yaml
```diff
@@ -0,0 +1,18 @@
+# External facing assigner libp2p endpoint, used for peering.
+kind: Service
```
masih, line 2: LGTM 👍 Only needs to be added to its corresponding `kusomization` resources. Otherwise it will not be picked up.

masih (approved):

PR #1894, opened 2023-06-06
Remove support for storethehash valuestore

gammazero:
We are no longer using this valuestore, and will not be in the future. Also, there is no need to support it for previous installations since re-indexing is needed anyway.


ischasny (approved):
Thanks Andrew

PR #1906, opened 2023-06-07
New indexers for indexing with FDB

gammazero:
Create new indexers for FDB indexing.

- arya: us-east-2a
- bala: us-east-2b
- cera: us-east-2c

These each use a `r6a.xlarge` instance with 
- cpu: "4"
- memory: 28Gi


masih (commented):
Overall looks good thanks @gammazero 

The only issue is that we need to define node workgroups for `r6a.2xlarge` since they do not currently exist in K8S Cluster. 

I'll get that sorted and Cc you for visibility. After that we can get the shiny new indexers up :rocket:

masih (approved):
New work groups are up. We are good to go

PR #1912, opened 2023-06-08
Disconnect dido and kepa from reads in prod

ischasny:
Leaving oden connected until inga catches up with dag.house. The [provider](https://inga.prod.cid.contact/providers/QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6ZmC) to watch. 

Also connect inga to not to break `providers` endpoint.


gammazero (approved):
Do you want to set the replica count to 0 for those indexers as well?

ischasny:
> Do you want to set the replica count to 0 for those indexers as well?

I'd leave them running in the background until we are certain that everything is working smoothly. Just in case we need to roll back.

PR #1921, opened 2023-06-09
Change dev assigner service hostname

gammazero:
Libp2p server is not exposed because the subdomain [assigner.dev](http://assigner.dev/).cid.contact is already in use by ingress for taking in direct announces via HTTP PUT.

The assigner has two different services (HTTP ingest and libp2p). They are handled with two different instances. One uses nginx which handles http, the other uses nlb directly to service. Those two create their own load balancer in aws. Therefore cannot use the same name server record.

This PR gives a new subdomain for the libp2p assigner service, i.e. change the annotation.

### DIFF
--- a/deploy/manifests/dev/us-east-2/tenant/storetheindex/assigner/service-external.yaml
+++ b/deploy/manifests/dev/us-east-2/tenant/storetheindex/assigner/service-external.yaml
@@ -6,7 +6,7 @@ metadata:
   annotations:
     service.beta.kubernetes.io/aws-load-balancer-type: nlb
     service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
-    external-dns.alpha.kubernetes.io/hostname: assigner.dev.cid.contact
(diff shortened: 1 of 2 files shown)


deploy/manifests/dev/us-east-2/tenant/storetheindex/assigner/service-external.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   annotations:
     service.beta.kubernetes.io/aws-load-balancer-type: nlb
     service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
-    external-dns.alpha.kubernetes.io/hostname: assigner.dev.cid.contact
+    external-dns.alpha.kubernetes.io/hostname: assigner-peer.dev.cid.contact
```
masih, line 9: also update peering in `prod` assigner?

gammazero: Done

masih (approved):

PR #1943, opened 2023-06-12
Increase number of CPUs on inga

ischasny:
Inga has higher concurrency factor now and has been running at 100% CPU lately that has been failing readiness probes.


masih (approved):
I am in the middle of node group updates. It might take 10 minutes or so before you see the effect of the changes here.

PR #1950, opened 2023-06-12
Deploy FDB and stateless dhstore on `prod`

masih:
Set up FDB cluster and dhstore on prod. Use 5 storage servers with increased rate keeper limits tuned for write-heavy workload.

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/fdb-cluster/cluster.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/fdb-cluster/cluster.yaml
@@ -1,0 +1,70 @@
+apiVersion: apps.foundationdb.org/v1beta2
+kind: FoundationDBCluster
+metadata:
+  name: fdb-prod-cluster
+spec:
+  version: 7.1.33
+  storageServersPerPod: 3
+  automationOptions:
+    replacements:
+      enabled: true
+  faultDomain:
+    key: kubernetes.io/hostname
+    valueFrom: spec.nodeName
+  databaseConfiguration:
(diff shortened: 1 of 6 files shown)


deploy/manifests/prod/us-east-2/tenant/storetheindex/fdb-cluster/cluster.yaml
```diff
@@ -0,0 +1,70 @@
...
+      foundationdb.org/fdb-cluster-name: fdb-prod-cluster
+    processClassLabels:
+      - foundationdb.org/fdb-process-class
+    processGroupIDLabels:
+      - foundationdb.org/fdb-process-group-id
+  minimumUptimeSecondsForBounce: 60
+  processes:
+    general:
+      customParameters:
+        - memory=20GiB
```
gammazero, line 31: Should this parameter be tuned to take up most or all of memory allocated to container?

gammazero (approved):
Would it be better to have a separate dhstore-stateless per indexer?

deploy/manifests/prod/us-east-2/tenant/storetheindex/fdb-cluster/cluster.yaml
masih: `storageServersPerPod` are set to 3 which leaves 20 GiB ish per process. Right?

masih:
> Would it be better to have a separate dhstore-stateless per indexer?

I don't think so. there is simply no need for it. Because we can independently scale dhstores this way since they have become stateless. This also means better utilsiation: if there is not much writes then we can scale down dhstores. 

Whereas if they are run one per indexer we cannot do this.

PR #1975, opened 2023-06-14
Fix landing in dev

ischasny:


masih (changes requested):
No longer needed now that we serve iframe from indexstar. we should also rip out fall back mechanism.

PR #1985, opened 2023-06-15
Deploy latest dhfind

ischasny:


masih (changes requested):
We should not revert the check on multihash. instead we should fix it here https://github.com/ipni/dhfind/pull/28

Because otherwise it would always double hash the multihashes and would cause redundant reads on dhstores.

ischasny:
> We should not revert the check on multihash. instead we should fix it here [ipni/dhfind#28](https://github.com/ipni/dhfind/pull/28)
> 
> Because otherwise it would always double hash the multihashes and would cause redundant reads on dhstores.

It wouldn't because indexstar won't route double hashed queries to it. But as discussed - closing it.

PR #1989, opened 2023-06-16
Update core and has dhstore as a valuestore

gammazero:
## Context
The new core has dhstore as a valuestore implementation, and not as an option to the engine.

## Proposed Changes
- Use latest core with dhstore valuestore instance.
- Make value store required.
- If valuestore is "" or "none", and DHStoreURL is configured, then use dhstore as valuestore. This makes upgrade compatible with previous configs.

## Tests
- Update e2e test to have 1st indexer write to pebble and 2nd write to dhstore.

## Revert Strategy
Change is safe to revert.

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -20,6 +20,7 @@ import (
 	"github.com/ipni/go-indexer-core/cache"
 	"github.com/ipni/go-indexer-core/cache/radixcache"
 	"github.com/ipni/go-indexer-core/engine"
+	"github.com/ipni/go-indexer-core/store/dhstore"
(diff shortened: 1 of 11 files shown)


ischasny:
Hi Andrew, thanks for the PR. What would be the migration plan for existing indexers? I.e. how would they move their value stores over to pebble?

gammazero:
@ischasny There is no migration plan.  This is completely compatible with their current configurations.  All indexers that are writing to dhstore have no local pebble valuestore.

ischasny (approved):

PR #2030, opened 2023-06-27
Make rewriting the mirror when resyncing configurable

gammazero:
When resyncing, sometimes we want to rewrite the mirror, usually we do not.

We may want to rewrite the mirror to get rid of the deleted multihash data that is not needed or to re-encode or compress differently.  Usually, we will only want to read existing CAR files from the mirror when resyncing.

### DIFF
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -55,6 +55,8 @@ type Ingest struct {
 	// MinimumKeyLengt causes any multihash, that has a digest length less than
 	// this, to be ignored.
 	MinimumKeyLength int
+	// OverwriteMirrorOnResync overwrites the advertisement when resyncing.
+	OverwriteMirrorOnResync bool
 	// PubSubTopic sets the topic name to which to subscribe for ingestion
 	// announcements.
 	PubSubTopic string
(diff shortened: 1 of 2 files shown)


config/ingest.go
```diff
@@ -55,6 +55,8 @@ type Ingest struct {
 	// MinimumKeyLengt causes any multihash, that has a digest length less than
 	// this, to be ignored.
 	MinimumKeyLength int
+	// OverwriteMirrorOnResync overwrites the advertisement when resyncing.
+	OverwriteMirrorOnResync bool
```
masih, line 59: At some point we probably this configurable per `admin sync --resync` command as well?

masih (approved):
Thank you for getting this sorted in such short time 🍻

config/ingest.go
gammazero: Agree, but that will slightly more complicated because of how the resync flag is passed through. I can do that in a separate PR. https://github.com/ipni/storetheindex/issues/2031

masih: Totally; separate all PRs ✊

PR #2033, opened 2023-06-28
Update the github certificate thumbprints

gammazero:
Looks like github changed certs, so we need a new fingerprint. Otherwise, ECR builds are failing with error:

"OpenIDConnect provider's HTTPS certificate doesn't match configured thumbprint"

I manually added the new fingerprint in the AWS console (IAM > identity providers > token.actions.githubusercontent.com) to see that it fixed the build problem, so now adding it here. The old fingerprint is still there in case this is a temporary situation and github changes back, or some servers still present the old certs.

### DIFF
--- a/deploy/infrastructure/common/github_actions.tf
+++ b/deploy/infrastructure/common/github_actions.tf
@@ -6,7 +6,8 @@ resource "aws_iam_openid_connect_provider" "github" {
   ]
 
   thumbprint_list = [
-    "6938fd4d98bab03faadb97b34396831e3780aea1"
+    "6938fd4d98bab03faadb97b34396831e3780aea1",
+    "a3b59e5fe884ee1f34d98eef858e3fb662ac104a"
   ]
   url = "https://token.actions.githubusercontent.com"
 }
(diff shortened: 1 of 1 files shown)


masih (approved):

willscott (approved):

masih (changes requested):
The new thumbprint does not match the one adverised by github?

https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/

deploy/infrastructure/common/github_actions.tf
```diff
@@ -6,7 +6,8 @@ resource "aws_iam_openid_connect_provider" "github" {
   ]
 
   thumbprint_list = [
-    "6938fd4d98bab03faadb97b34396831e3780aea1"
+    "6938fd4d98bab03faadb97b34396831e3780aea1",
+    "a3b59e5fe884ee1f34d98eef858e3fb662ac104a"
```
masih, line 10: Should this be `1c58a3a8518e8759bf075b76b750d4f2df264fcd` ?

gammazero: Maybe.  I grabbed "a3b59e5fe884ee1f34d98eef858e3fb662ac104a" from the github.com certificate that I saw, and I tested that using it fixed the problem. If there are other valid github certs then we should include their fingerprints as well.

gammazero:
I did a `terraform apply` with the `1c58a3a8518e8759bf075b76b750d4f2df264fcd` fingerprint. If that does not work, then we can add the other.

PR #2037, opened 2023-06-28
Clear LastError on good sync

gammazero:
- Do not set LastError on skipped advertisements
- Do not set LastError if db write failure
- Use new go-libipni that does not block SyncFinished events.

### DIFF
--- a/.github/workflows/ecr-publisher.yml
+++ b/.github/workflows/ecr-publisher.yml
@@ -14,7 +14,6 @@ on:
       - 'doc/**'
     branches:
       - main
-      - update-last-error
 
 jobs:
   publisher:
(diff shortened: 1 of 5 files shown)


.github/workflows/ecr-publisher.yml
```diff
@@ -14,6 +14,7 @@ on:
       - 'doc/**'
     branches:
       - main
+      - update-last-error
```
masih, line 17: is this for testing?

masih (approved):

.github/workflows/ecr-publisher.yml
gammazero: Yes, will remove that.

Issue #2062, opened 2023-06-30
High CPU usage to the extent where the node becomes unresponsive


masih:
Observed on Inga on 30th of June between 0930-1130 UTC on `inga`.

Check grafana metrics, logs and investigate root cause. For now, we should not roll out 0.7.5 to prod.

masih:
Updated the issue title as this occurs on 0.7.3 too. 

PProf/profile on the node seems to suggest locking contention with leveldb lookups
<img width="1256" alt="Screenshot 2023-06-30 at 14 14 34" src="https://github.com/ipni/storetheindex/assets/301855/384cc288-4e07-4a3d-93a5-def5ac46d371">

Which points to index counter? Investigating.

masih:
Update: Excluding index counter from find server seems to have resolved the CPU contention. Working on a fix in #2067

gammazero:
Fixed by #2101 and by integration of new provider cache into dhstore and indexstar.

PR #2070, opened 2023-06-30
Make index counts configurable and disable by default

masih:
Index count is enabled by default across both ingestion and find server. While we look into how to do it without extensive lock contention make the feature configurable and disable it by default on both ingest and find server.

Relates to #2062

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -171,7 +171,7 @@ func daemonAction(cctx *cli.Context) error {
 	indexerCore := engine.New(valueStore, engine.WithCache(resultCache))
 
 	var indexCounts *counter.IndexCounts
-	if cfg.Ingest.IndexCountsEnabled || cfg.Finder.IndexCountsEnabled {
+	if cfg.Indexer.IndexCountEnabled {
 		indexCounts = counter.NewIndexCounts(dstore)
 		indexCounts.SetTotalAddend(cfg.Indexer.IndexCountTotalAddend)
 	}
@@ -195,17 +195,14 @@ func daemonAction(cctx *cli.Context) error {
 		if err != nil {
 			return fmt.Errorf("bad find address %s: %s", findAddr, err)
 		}
-		opts := []httpfind.Option{
+		findSvr, err = httpfind.New(findNetAddr.String(), indexerCore, reg,
 			httpfind.WithReadTimeout(time.Duration(cfg.Finder.ApiReadTimeout)),
 			httpfind.WithWriteTimeout(time.Duration(cfg.Finder.ApiWriteTimeout)),
 			httpfind.WithMaxConnections(cfg.Finder.MaxConnections),
 			httpfind.WithHomepage(cfg.Finder.Webpage),
+			httpfind.WithIndexCounts(indexCounts),
 			httpfind.WithVersion(cctx.App.Version),
-		}
-		if cfg.Finder.IndexCountsEnabled {
-			opts = append(opts, httpfind.WithIndexCounts(indexCounts))
-		}
-		findSvr, err = httpfind.New(findNetAddr.String(), indexerCore, reg, opts...)
+		)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


willscott (approved):

command/daemon.go
```diff
@@ -247,9 +253,12 @@ func daemonAction(cctx *cli.Context) error {
 			cfg.Ingest.ResendDirectAnnounce = false
 		}
 
+		var ingestOpts []ingest.Option
+		if cfg.Ingest.IndexCountsEnabled {
+			ingestOpts = append(ingestOpts, ingest.WithIndexCounts(indexCounts))
+		}
```
gammazero, line 259: This is not needed, because passing in a nil `indexCounts` is the same as not passing the option at all:
https://github.com/ipni/storetheindex/blob/main/internal/ingest/option.go#L31

command/daemon.go
```diff
@@ -170,8 +170,11 @@ func daemonAction(cctx *cli.Context) error {
 	// Create indexer core
 	indexerCore := engine.New(valueStore, engine.WithCache(resultCache))
 
-	indexCounts := counter.NewIndexCounts(dstore)
-	indexCounts.SetTotalAddend(cfg.Indexer.IndexCountTotalAddend)
+	var indexCounts *counter.IndexCounts
+	if cfg.Ingest.IndexCountsEnabled || cfg.Finder.IndexCountsEnabled {
```
gammazero, line 174: Is there a reason to be able to configure the option in other places? I think there should be an enable option in one place that controls the creation of the counter.  If the counter is created, then it gets passed to both ingester and find server, otherwise, nill get passed to both.

gammazero:
Maybe we just get rid of these counters, if we are going to support unique multihash counts in database.

command/daemon.go
masih: This if condition is needed because index counter is instantiated if finder flag **or** ingest flag is set.

command/daemon.go
masih: The rationale for separate flags was that the load pattern on each of those flags is very different. One is pure writes to index counts and the other is pure read. 

Considering explicit options exist already in ingest and finder config, it seems reasonable to also have separate config for them.

command/daemon.go
```diff
@@ -192,14 +195,17 @@ func daemonAction(cctx *cli.Context) error {
 		if err != nil {
 			return fmt.Errorf("bad find address %s: %s", findAddr, err)
 		}
-		findSvr, err = httpfind.New(findNetAddr.String(), indexerCore, reg,
+		opts := []httpfind.Option{
```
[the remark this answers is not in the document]
masih: See https://github.com/ipni/storetheindex/pull/2070#discussion_r1248099248

masih:
> Maybe we just get rid of these counters, if we are going to support unique multihash counts in database. Persisting this is already expensive in a local datastore, because we keep counts per context ID.


I think from visibility perspective we want both, i.e. globally unique CID count, and unique index record count. The former tells us the number of destinations and the latter tells us how many ways there are to get there. Having said that, if we have to pick the former is perhaps more important to know.

Considering the work on measuring the latter, it's worthwhile keeping this capability but optimising it such that it has minimal impact on ingest rate I think.

command/daemon.go
gammazero: Consider: If write is not needed, then read cannot be done.  If read is not needed, then there is no reason to write.

command/daemon.go
gammazero: What I am saying is that since either flag instantiates the index counter, always pass `WithIndexCounter(indexCounts)`, and the fact that indexCounts is `nil` or not handles making it optional.

gammazero:
One thing that we can do to reduce read contention is to move this out of the info reported providers info, and make it require a separate request or only use the counters to update metrics.

gammazero (approved):
Lets consider removing the counters from the provider info to reduce read contention. Or, at least remove from the single provider read which is used to populate find results.

PR #2082, opened 2023-07-03
Replace prod dhfind with new dhstore

gammazero:
- Update indexstar backend to replace dhfind with dhstore
- Update dhstore to new image and specify providers URL
- Set dhfind replica count to 0

Safe to revert


masih (commented):
LGTM thanks @gammazero  🍻 

To avoid momentary outage on prod however, would it make sense to break this PR into two PRs, where:
* First `dhstore` changes are rolled out, including additional `--backend` endpoints that point to `dhstore`, then
* `dhfind` is spun down with the removal of corresponding `--backends` in indexstar.

Rolling the changes this way should assure that a superset of lookup endpoints are running before the (then becoming) redundant `dhfind` endpoints are spun down. 

WDYT? Does that make sense?

gammazero:
The outage should be tolerable, but to be on the safe side I will split into two.

PR #2083, opened 2023-07-03
Update to new dhstore

gammazero:
- Update to new dhstore that supports dhfind
- Configure each dhstore with providers URL


masih (approved):
Also add dhstore endpoints as `--backend` in indexstar? Without doing this no non-dboulehashed lookups would go to the new dhstores.

PR #2084, opened 2023-07-03
Replace dhfind with new dhstore

gammazero:
- Set dhfind replica count to 0
- Replace dhfind with dhstore in indexstar config

To be deployed after #2083


masih (approved):
LGTM to merge after #2083 deployment is all good 👍

PR #2101, opened 2023-07-08
Remove index counter functionality

gammazero:
This counter is very expensive to maintain and ends up offering little value.

- Remove index logic
- Cleanup datastore to remove index count records


willscott:
Does this remove the ability to know overall size of the index?
If so we should consider if that remains a thing product want to have a clear sense of and if there's a cheaper way to get a decent approximation of index size before removing it.

gammazero:
> Does this remove the ability to know overall size of the index?

It removes the ability to know the number of mappings multihash=>(providerID+contextID) that have been stored by particular indexer.  So it does not give an indication of the overall size of the index, unless all indexers are taken into consideration and there is no overlap in what any indexer has stored.

It seems like this index count is a useful metric, but as far as I can tell it has only been a source of questions about what it means and has never provided an answer to anything that anyone wanted to know. It seems that what we really want is the number of _unique_ multihashes indexed, and how many of those are retrievable. ... which this is not.

masih:
I am curious to hear @TorfinnOlsen 's take on this. 

Here is how it occurs for me:
I think this is an important metric that we want to provide going forward. Based on previous conversations, it is perhaps not as important as the unique CID count but it is useful nevertheless. Where the unique CID count tells us about the "number of destinations", the index count tells us "how many ways there are to get to them". From product perspective, the number of ways to get to CID would help reason about content distribution across providers. So overall I think we want to retain the ability to measure it, _but_ not at the cost of hindering ingest rate. 

The work is already done to disable the current measurement mechanism by default. In terms of code removal, we probably wan to retain the hooks that are already in place to increment counters, since whatever the implementation the hooks in code will be needed.

In terms of improving efficiency, the counter should ideally limit the impact per provider and avoid global locks. One can think of other more elaborate designs, e.g. roll up sums, to further reduce the footprint of synchronization when incrementing counters. But right now this does not seem to be a priority.

TorfinnOlsen:
I think the index count is valuable to know, but I think the problem with our present index count is that it counts everything and we don't have an easy way to isolate non duplicated or inactive CIDs. 

The way I see it the 'count' is currently made up of a strata of value pairs:

1. CIDs that active storage providers who are still providing, have advertised to us, and are retrievable at the CID:ProviderID advertised. Can be retrieved, from the location implied by this key value pair.

2. CIDs active storage providers are no longer providing. ProviderID has not changed in anyway but the advertised CID:ProviderID pair is no longer relevant. CID:ProviderID remains advertised via IPNI. Cannot be retrieved.

3. CIDs advertised to IPNI by inactive ProviderIDs. ProviderID meets our inactivity criteria(2 weeks?). Nothing can be retrieved from this ProviderID all CID's associated with it are dead value pairs.

4. Duplications of the above incurred by kicking off ingestion(on cid.contact). 

I think the count function would be much more valuable if we could project this stratification in the count. So right now we have a bar chart which simply states the count every week. There's no nuance however into what the makeup of that ingestion is though so when we have big spikes in ingestion we know it's because we are building up a new index, but to folks outside the team they perceive that as growth of the index which isn't really accurate. The count has grown but the index itself has not.

Right now the index count has surpassed 2 trillion key value pairs counted, but we know that the number of unique CID's is in the ~200 billion range I believe. At least having a consistent unique count to contrast the ingestion count would help people understand that the count is not really representative of network growth. I believe that's how it's widely perceived now in spite of our explanations that this is the case people naturally bias towards understanding our count in this way.
 
**Summary:** The count is valuable, but is to vague in it's current implementation is not high enough resolution and creates a potential for misunderstanding outside of our team. Unique CID counts are also a valuable representation of how many CID's are on the network, but what people really want to know is "how big the index is" which means to me the number of unique CID:ProviderID's which are associated with active providers. 

CID's which no longer exist, duplicates, and CID's associated with inactive providers are basically historical records of what has happened. As far as I can tell the only utility they serve is possibly reconstructing prior states of the network. While this may be an important function, including it in the counts we share creates a false impression of how big the "index" is.

gammazero:
We should at least remove this counter from the information returned from the `/providers` API, and instead make it either a metric or make it separately queryable.

TorfinnOlsen:
@gammazero I can't think of any reason why that would be a loss to us. I assume our grafana panel here: https://protocollabs.grafana.net/d/pPROWOD7z/providers?orgId=1&refresh=5s&from=1688446116224&to=1689050916224&viewPanel=23

will be broken but we can change up our reporting for the time being. 

I think if we manage to sort out the distance of providers, and sometime soon thereafter the unique cid count this puts us in a position of higher visibility than where we are at present.

I'm curious this is probably a somewhat ignorant question, but considering the index is a key value store can't we just query it for unique CID's? I know this is greatly simplifying the problem but someone asked me this question and I found myself without a good answer.

masih:
> can't we just query it for unique CID's
No. The scale at which the key value store operates can't simply return the number of unique keys.

masih (approved):

PR #2116, opened 2023-07-11
Use new FDB indexers as providers backends.

gammazero:
This is necessary since there are 17 providers known to the fdb indexers that are not known to inga.

It is no longer necessary to remove these for FDB load testing, because the indexstar provider cache removes any significant query load.

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/indexstar/deployment.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/indexstar/deployment.yaml
@@ -16,10 +16,9 @@ spec:
             - '--providersBackends=http://inga-indexer:3000/'
             
             # New FDB-backed new indexers
-            # Excluded while load testing FoundationDB
-            # - '--providersBackends=http://arya-indexer:3000/'
-            # - '--providersBackends=http://bala-indexer:3000/'
-            # - '--providersBackends=http://cera-indexer:3000/'
+            - '--providersBackends=http://arya-indexer:3000/'
+            - '--providersBackends=http://bala-indexer:3000/'
+            - '--providersBackends=http://cera-indexer:3000/'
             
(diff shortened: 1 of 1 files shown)


masih:
> This is necessary since there are 17 providers known to the fdb indexers that are not known to inga.

I am curious why? direct HTTP announces are propagated through gossipsub and inga ins peered with assigner.

> because the indexstar provider cache removes any significant query load

They were not included to avoid having inconsistent provider info since the new indexers are still catching up, and less so for load. Does the new chaching mechism determine which provider info is the latest one and how?

gammazero:
@masih
> I am curious why? direct HTTP announces are propagated through gossipsub and inga ins peered with assigner.
I am wondering the same thing. It might have happened that an announce was received when inga was not available, or inga was unable to communicate with the publisher and one of the fdb indexer go through.

> They were not included to avoid having inconsistent provider info since the new indexers are still catching up, and less so for load. Does the new chaching mechism determine which provider info is the latest one and how?

Yes, it does determine which provider info is the latest. It does this by comparing the LastAdvertisementTime. This is not perfect becaue one indexer could have seen a later advertisement at a later time, but works most of them time. The provider address data returned in responses does not change very often, so even in the event of choosing the provider info that is not the latest, it is very unlikely to provide wrong data, and should correct itself quickly if it does. 

The place this does not work is where one indexer is caught up and is not getting additional updates, while another indexer is far behind and is getting caught up, and the most recent ad the behind indexer has started reading is not far enough in the chain to have the new address.  That last part is neecessary also, because even a very far behind indexer always applies the provider data from the latest advertisement in the chain, even if it has to process many previous advertisements.

masih:
> but works most of them time. 

Let's monitor toe 200s to 404s ratio. I think we have observed a few alerts since this was merged which may or may not be related.

PR #2122, opened 2023-07-13
Add FDB 7.2 and 7.3 binaries to controller in prep for upgrade

masih:
In preparation for upgrading foundation db, add the binaries for two higher versions with the latest patch.


gammazero (approved):
I assume that after upgrade we can get rid of the multiple versions.

masih:
> I assume that after upgrade we can get rid of the multiple versions.

Indeed. For now we want to support older versions for server side experimentation, and graceful client side upgrades.

PR #2141, opened 2023-07-19
Increase sensitivity to read amplification relative to compaction

masih:
Reduce the threshold at which compaction is triggered and stop writes
more aggressively if compaction is not keeping up.

Set the config for both `helga` and `qiu`.

These changes need to be experimented with to see impact on ingest rate
first.

**Do not merge until the current compaction debt and provider lag is
caught up.**


gammazero:
I think the values should be set to whatever they are on dhstore-helga, since we know that worked well for this instance type. If we need to in the future the values can be adjusted.

gammazero (approved):

gammazero:
We can keep this as a draft PR, so it is here in case we see read amplification going up and we need to apply these settings.

masih:
No longer needed as Pebble settings are stabilised though increase in volume throughput.

PR #2148, opened 2023-07-24
Allocate more storage on dhstore-qiu

gammazero:


deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-qiu/pvc.yaml
```diff
@@ -7,5 +7,5 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 4Ti
+      storage: 15Ti
```
masih, line 10: 16?

PR #2155, opened 2023-07-26
Fix panic in bootstrap

gammazero:
If the bootstrap config function LoadTempBootstrapPeers is not defined, then bootstrapping panics.

This happens here, because a nil function is called: https://github.com/ipfs/kubo/blob/master/core/bootstrap/bootstrap.go#L248

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


gammazero:
Force merging, since this fixes a panic in already-merged code.

PR #2166, opened 2023-07-27
Shut down FDB on `dev`

masih:
Shut down FDB cluster in dev now that it is not used for the write path due to memory issues. Remove dependant services lik its dhstore. Remove corresponding indexers from the assigner.

Assigner will have no indexers in its pool but it will continue to propagate requests over gossipsub which get picked up by ago since ago is peered to the assigner.


willscott (approved):

gammazero (approved):
Sad to see this go :-( Hopefully it will be back soon.

masih:
> Sad to see this go :-( Hopefully it will be back soon.

I know.. we can revert this PR to bring it all back. No point keeping dead code in the repo

PR #2167, opened 2023-07-27
Include telemetry service in dev deploy manifests

masih:
The new telemetry service was added but not included in the kustomization that gets applied by flux CD. Add it to trigger deployment.


gammazero (approved):
Doh!

PR #2170, opened 2023-07-27
Add missing deployment config for telemetry

gammazero:


masih (changes requested):
There seem to be some issues in the telemetry dev overlay which I can't comment on in this PR since that code is already merged. Here is the list:
* No need to specify port [here](https://github.com/ipni/storetheindex/blob/1eff8ab4476255b0072bf71486c10ddd8fab9197/deploy/manifests/dev/us-east-2/tenant/storetheindex/telemetry/deployment.yaml#L20-L22) as it is already specified in base.
* [This ](https://github.com/ipni/storetheindex/blob/1eff8ab4476255b0072bf71486c10ddd8fab9197/deploy/manifests/dev/us-east-2/tenant/storetheindex/telemetry/deployment.yaml#L23-L27) will fail the deployment. It is not needed since telemetry has no identity nor config, nor the manifest specify such volumes to then be mounted.
* [3](https://github.com/ipni/storetheindex/blob/1eff8ab4476255b0072bf71486c10ddd8fab9197/deploy/manifests/dev/us-east-2/tenant/storetheindex/telemetry/deployment.yaml#L30) CPUs seems a tad too high?
* There is no [service-external](https://github.com/ipni/storetheindex/blob/1eff8ab4476255b0072bf71486c10ddd8fab9197/deploy/manifests/dev/us-east-2/tenant/storetheindex/telemetry/kustomization.yaml#L8C6-L8C6). This line should be removed.

masih (approved):

PR #2175, opened 2023-07-27
Remove all non-encrypted indexers and their records

masih:
The non-encrypted indexers were shut down a few weeks ago. But their manifests was left in code base in case we needed to reinstantiate them. Enough time has passed for us to be confident that they are no
 longer needed.

Remove their data and corresponding manifests. Clean up snapshots of the unencrypted data. Long live reader privacy.


willscott (approved):
🎉🔐

gammazero (approved):

PR #2216, opened 2023-08-10
Update to go-libipni without selectors

gammazero:
Remove selectors from go-libipni API, and do not use selectors to control traversal of advertisement chains. This is a refactor toward using the new dagsync with ipnisync.


masih (approved):
PR title seems off?

gammazero:
Lol oops... ipbp2p -> libipni

PR #2242, opened 2023-09-12
feat: publicly expose e2e TestIndexerRunner

rvagg:
Ref: https://github.com/ipld/frisbii/pull/36

This doesn't introduce anything new to the test, but exposes pieces in a way to make the runner framework reusable. It turns out that it's probably pretty useful just for generic test environments and maybe should even be its own package. But for now this works for me in Frisbii; see https://github.com/ipld/frisbii/pull/36.

### DIFF
--- a/test/testindexer.go
+++ b/test/testindexer.go
@@ -1,0 +1,200 @@
+package test
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	IndexerReadyMatch    = "Indexer is ready"
+	ProviderHasPeerMatch = "connected to peer successfully"
+	ProviderReadyMatch   = "admin http server listening"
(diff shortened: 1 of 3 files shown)


test/testindexer.go
```diff
@@ -0,0 +1,200 @@
...
+			// but helps see what's happening, especially when the test fails.
+			tr.t.Logf("%s: %s", name, line)
+
+			for _, watcher := range tr.watchers {
+				if watcher.Prog == name && strings.Contains(line, strings.ToLower(watcher.Match)) {
+					watcher.Signal <- struct{}{}
+				}
+			}
+		}
+	}()
```
gammazero, line 172: I know that is what it did previously, but...

It seems strange to watch every command and see if its output contains something a watcher is looking for.  Would it make more sense if `...watchers` were a parameter to `Start`?

gammazero (approved):
See comments

test/testindexer.go
rvagg: A bit tricky because that's two lots of varargs, but if you have a look at the latest commit you'll see my attempt to resolve the problem by introducing a new `Execution` type that lets us build an "execution" and attach watchers to it d2bb69c1

gammazero (approved):
Awesome!

PR #2316, opened 2023-10-04
bump pvc to 2gb

willscott:


deploy/manifests/dev/us-east-2/tenant/storetheindex/heyfil/pvc.yaml
```diff
@@ -9,5 +9,5 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 1Gi
+      storage: 2Gi
```
masih, line 12: This change is for the dev instance. checking that it's intended.

masih (approved):

deploy/manifests/dev/us-east-2/tenant/storetheindex/heyfil/pvc.yaml
willscott: yes - that's where we've promoted the recent change / are seeing disk full errors

PR #2342, opened 2023-10-17
Try deploying dhstore-seka on c6a.12xlarge

gammazero:


masih:
No longer needed. Root cause was cluster autoscaler OOM. Fixed in #2343 and dhstore is back up

PR #2350, opened 2023-10-23
Reduce max ASG size for prod worker node group

masih:
The size was increased last week as part of investigating deployment issues.

Since then the issue has been resolved. Reduce to avoid excessively large number of nodes.


gammazero (approved):

gammazero:
This will need to be increased in 3-4 months as we need more storage nodes.

PR #2491, opened 2024-01-23
Implement the core federation protocol

masih:
First pass at implementing the IPNI federation protocol:
* core data structures/functionality
* periodic snapshot
* basic vector clock reconciliation
* HTTP serve mux for `/ipni/v1/fed`

See: https://github.com/ipni/specs/pull/27

### DIFF
--- a/federation/federation.go
+++ b/federation/federation.go
@@ -301,6 +301,11 @@ func (f *Federation) reconcile(ctx context.Context, t time.Time) error {
 			snapshotCid := head.Head.(cidlink.Link).Cid
 			logger = logger.With("snapshotCid", snapshotCid)
 
+			if cid.Undef.Equals(snapshotCid) {
+				logger.Debug("Head link is empty")
+				continue
+			}
+
 			if seenCid, ok := f.lastSeenSnapshotByMember[member.ID]; ok && seenCid.Equals(snapshotCid) {
 				logger.Debugw("Snapshot already seen; skipping reconciliation")
 				continue
--- a/federation/federation.go
+++ b/federation/federation.go
@@ -119,7 +119,9 @@ func (f *Federation) snapshot(ctx context.Context, t time.Time) error {
 		return err
 	}
 	newSnapshot := &Snapshot{
-		Epoch:       uint64(t.UTC().Unix()), // TODO: consider time dilution?
+		// TODO: consider reducing the granularity of epoch. Millisecond increments in terms of unix clock may be too
+		//       fine for the purposes of IPNI federation.
+		Epoch:       uint64(t.UTC().Unix()),
 		VectorClock: f.vc.tick(f.host.ID()),
 		Providers: ProvidersIngestStatusMap{
 			Values: make(map[string]IngestStatus),
--- a/federation/federation.go
+++ b/federation/federation.go
@@ -1,0 +1,411 @@
+package federation
+
+import (
+	"context"
+	"errors"
+	"net"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/codec/dagjson"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/multicodec"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+	"github.com/ipni/go-libipni/find/client"
+	"github.com/ipni/go-libipni/find/model"
+	"github.com/ipni/go-libipni/maurl"
+	"github.com/ipni/go-libipni/mautil"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+var (
+	headSnapshotKey = datastore.NewKey("head-snapshot")
+	logger          = logging.Logger("federation")
+)
+
+type (
+	Federation struct {
+		*options
+		vc vectorClock
+
+		mu            sync.Mutex
+		headNodeCache ipld.Node
+		httpServer    http.Server
+		shutdown      chan struct{}
+
+		// lastSeenSnapshotByMember is the cache of already reconciled snapshots by federation member.
+		// An in-memory data structure is enough, because upon indexer restart we want to check if
+		// the local node has missed anything since reconciliation process is cheap enough.
+		lastSeenSnapshotByMember map[peer.ID]cid.Cid
+	}
+)
+
+func New(o ...Option) (*Federation, error) {
+	opts, err := newOptions(o...)
+	if err != nil {
+		return nil, err
+	}
+	return &Federation{
+		options:                  opts,
+		vc:                       newVectorClock(),
+		shutdown:                 make(chan struct{}),
+		lastSeenSnapshotByMember: make(map[peer.ID]cid.Cid),
+	}, nil
+}
+
+func (f *Federation) Start(_ context.Context) error {
+
+	listen, err := net.Listen("tcp", f.httpListenAddr)
+	if err != nil {
+		return err
+	}
+	go func() {
+		f.httpServer.Handler = f.ServeMux()
+		switch err := f.httpServer.Serve(listen); {
+		case errors.Is(err, http.ErrServerClosed):
+			logger.Infow("Federation HTTP server stopped")
+		default:
+			logger.Errorw("Federation HTTP server stopped unexpectedly")
+		}
+	}()
+
+	go func() {
+		snapshotTicker := time.NewTicker(f.snapshotInterval)
+		reconciliationTicker := time.NewTicker(f.reconciliationInterval)
+		defer func() {
+			snapshotTicker.Stop()
+			reconciliationTicker.Stop()
+		}()
+		for {
+			select {
+			case <-f.shutdown:
+				logger.Warnw("Stopping federation event loop")
+				return
+			case t := <-snapshotTicker.C:
+				if err := f.snapshot(context.TODO(), t); err != nil {
+					logger.Errorw("Failed to take snapshot", "err", err)
+				}
+			case t := <-reconciliationTicker.C:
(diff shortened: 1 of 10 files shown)


federation/federation.go
```diff
@@ -0,0 +1,195 @@
...
+		options:              opts,
+		snapshotTicker:       time.NewTicker(opts.snapshotInterval),
+		reconciliationTicker: time.NewTicker(opts.reconciliationInterval),
+		vc:                   newVectorClock(),
+		ctx:                  ctx,
+		cancel:               cancel,
+	}, nil
+}
+
+func (f *Federation) Start(_ context.Context) error {
```
willscott, line 57: Why take and not use cancel here?

federation/federation.go
```diff
@@ -0,0 +1,195 @@
...
+		*options
+		members []peer.ID
+		vc      vectorClock
+
+		mu                   sync.Mutex
+		snapshotTicker       *time.Ticker
+		reconciliationTicker *time.Ticker
+		headNodeCache        ipld.Node
+		ctx                  context.Context
+		cancel               context.CancelFunc
```
willscott, line 37: Prefer letting the caller cancel rather than storing on the struct?

federation/schema.go
```diff
@@ -0,0 +1,161 @@
...
+	Prototypes struct {
+		Head                     schema.TypedPrototype
+		Snapshot                 schema.TypedPrototype
+		ProvidersIngestStatusMap schema.TypedPrototype
+		IngestStatus             schema.TypedPrototype
+
+		link ipld.LinkPrototype
+	}
+
+	//go:embed schema.ipldsch
```
willscott, line 36: This is an interesting pattern - does it impact startup time to do this on startup rather than a compile time step?

federation/schema.ipldsch
```diff
@@ -0,0 +1,20 @@
+type FederationHead  struct {
+    Head            Link
+    Topic  optional String
+    PublicKey       Bytes
+    Signature       Bytes
+}
+
+type FederationSnapshot struct {
+    Epoch             Int
+    VectorClock       Int
```
willscott, line 10: This is this instances clock, not the full vector clock, right?

gammazero (commented):
There are a few minor coding issues commented on, but the more important thing is to establish an understanding of how to do:
1) Discovery of new indexers within the federation.
2) Authentication to allow membership
3) Reconsiliation process.

As for discovery, an indexer could be given a list of other indexer peer IDs. Or, indexers could periodically announce their presence on a libp2p topic, where the topic is the name of the federation. Other indexers subscribed to that same topic/federation whould choose whether to accept another indexer as a federation peer.

If discovery is limited to a configured list, where is this list published? If using libp2p pubsub, is there some bootstrap node somewhere? Is there a registry of "trusted" indexers for a given federation?

federation/federation.go
masih: this is to cancel the event loop started internally by Start. 

We can use the context passed into start. but that context might be only used for starting the event loop and could be surprising to the users if it is tied to the entire lifecyle of the federation.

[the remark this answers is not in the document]
masih: The issue with closing channel i see is that if we decide to separate the goroutines that handle reconciliation and snapshotting then we need a channel for each. 

I am not fussed. both work.

federation/federation.go
masih: Because it might be surprising to the user that start context is tied to the entire lifecycle of federation rather than just starting it.

federation/federation.go
```diff
@@ -0,0 +1,195 @@
...
+		ctx:                  ctx,
+		cancel:               cancel,
+	}, nil
+}
+
+func (f *Federation) Start(_ context.Context) error {
+	go func() {
+		for {
+			select {
+			case <-f.ctx.Done():
```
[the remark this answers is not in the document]
masih: I see a future where we most likely need to separate the reconciliation and snapshot goroutines to avoid them blocking eachother in a case where provider list is long or the process takes too long.

Using channels would then mean one channel for each goroutine which seems like an overkill for just stopping the goroutines?

federation/federation.go
```diff
@@ -0,0 +1,195 @@
...
+
+	headLink, err := f.getHeadLink(ctx)
+	switch {
+	case errors.Is(err, datastore.ErrNotFound):
+		// Set headNodeCache to an empty node to avoid hitting datastore every time head is fetched.
+		f.headNodeCache = basicnode.Prototype.Any.NewBuilder().Build()
+		return f.headNodeCache, nil
+	case err != nil:
+		return nil, err
+	default:
```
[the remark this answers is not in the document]
masih: The suggested code seems to be checking `err` the same number of times as switch. What am I missing?

federation/federation.go
```diff
@@ -0,0 +1,195 @@
...
+func (f *Federation) Shutdown(_ context.Context) error {
+	f.snapshotTicker.Stop()
+	f.cancel()
+	return nil
+}
+
+func (f *Federation) reconcile(ctx context.Context, t time.Time) error {
+	// TODO: For each f.peers, get head, get snapshot, f.vc.reconcile, aggregate providers.
+	return nil
+}
```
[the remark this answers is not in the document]
masih: 1. Right.
2. Correct. Vector clocks are monotonically increasing.

federation/schema.go
masih: I have not measured it. I also have not heard complains about its current usage elsewhere being slow.

Happy to measure it if you think it's a blocker.

federation/schema.go
```diff
@@ -0,0 +1,161 @@
...
+		Head      ipld.Link
+		Topic     *string
+		PublicKey []byte
+		Signature []byte
+	}
+	Snapshot struct {
+		Epoch       uint64
+		VectorClock uint64
+		Providers   ProvidersIngestStatusMap
+		Previous    ipld.Link
```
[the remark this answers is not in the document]
masih: > A previous snapshot does not seem necessary for reconciliation.

Right. 

The rationale for having a chain of snapshot is to provide opportunity for reputation measurements in the future since IPNI will need some sort of reputation system atop a permissionless federation setup. 

Side note: It might be worth bringing up this question on the spec PR as it relates more to protocol design itself than the implementation. I think it is worth writing a rationale for its existence in the spec at least, if we reach the consensus to keep it.

federation/schema.ipldsch
masih: Right

masih:
> establish an understanding of how to do:
> 
> 1. Discovery of new indexers within the federation.
> 2. Authentication to allow membership
> 3. Reconsiliation process.
> 
> As for discovery, an indexer could be given a list of other indexer peer IDs. Or, indexers could periodically announce their presence on a libp2p topic, where the topic is the name of the federation. Other indexers subscribed to that same topic/federation whould choose whether to accept another indexer as a federation peer.
> 
> If discovery is limited to a configured list, where is this list published? If using libp2p pubsub, is there some bootstrap node somewhere? Is there a registry of "trusted" indexers for a given federation?

1. Discovery of indexers is strictly considered out of scope for the fist pass to 1) reduce scope, and 2) be continued as part of the ambient discovery proposal in IPFS specs. The list is configured as a static config when indexers start up, much like the bootstrap peers.

2. This makes the federation permission-ful in the initial implementation.

3. What can I help clarify re reconciliation process that's not covered by the spec?

federation/federation.go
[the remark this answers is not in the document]
gammazero: In the non-error usual path error is checked only once (`if err != nil`) in suggestion, whereas `switch` checks `case errors.Is(err, datastore.ErrNotFound)` and then `case err != nil`.

federation/vector_clock.go
```diff
@@ -0,0 +1,35 @@
...
+	return clock, found
+}
+
+func (vc vectorClock) tick(id peer.ID) uint64 {
+	vc[id] = vc[id] + 1
+	return vc[id]
+
+}
+
+func (vc vectorClock) untick(id peer.ID) uint64 {
```
willscott, line 24: can you add comments on what these methods do.

i'm uncertain why we want to run the clock backwards - that's generally a much harder state machine to model when allowed

federation/federation.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+		}
+	}()
+	return nil
+}
+
+func (f *Federation) ServeMux() *http.ServeMux {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/ipni/v1/fed/head", f.handleV1FedHead)
+	mux.HandleFunc("/ipni/v1/fed/", f.handleV1FedSubtree)
```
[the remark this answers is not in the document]
masih: I strongly advise to have a unified prefix (plus version) for IPNI URIs. I think it makes it clear what the grouping is as a coherent protocol for content discovery.

federation/federation.go
```diff
@@ -0,0 +1,411 @@
...
+func (f *Federation) snapshot(ctx context.Context, t time.Time) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	previous, err := f.getHeadLink(ctx)
+	if err != nil {
+		return err
+	}
+	newSnapshot := &Snapshot{
+		Epoch:       uint64(t.UTC().Unix()), // TODO: consider time dilution?
```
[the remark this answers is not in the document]
masih: By dilution in the comment i mean something that doesn't increment at millisecond granularity. 

Please let me know if this is a blocker and something to resolve prior to further progress.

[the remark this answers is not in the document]
willscott: not a blocker - but more clarity in the comment would be nice :)

federation/vector_clock.go
masih: Will add comments. 

The reason for this is to handle error cases correctly, where taking a snapshot might fail before it is stored in the snapshot chain, and if that happens we don't want to leave the clock incremented. 

Please let me know if you have any alternative suggestions on how to atomically record a snapshot without untick.

willscott: the other pattern might be to make a "copy" at the higher tick, but not commit it back to the snapshot until reaching consensus/finality. - it's nice to have the assumption that the 'consensus' / 'committed' vector clock will monotonically increase

federation/federation.go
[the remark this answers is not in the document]
masih: Re mounting, the ingress reverse proxy would do the rewrite and it really makes no difference what is here. Except for the fact that the current mux makes it possible to run a fully self contained federation server separate from an indexer itself.

PR #2504, opened 2024-02-08
Simplify Dockerfile for a minimal non-root image

masih:
Remove user switch and additional binary installations from Dockerfile for a minimal non-root image that still contains a shell, at entrypoint `sh`.


willscott (approved):
this looks good to me

gammazero (commented):
We will need to run the `assigner` subcommand when running this as an assigner. This means there is some additional change that needs to be made to deployment.

gammazero (approved):

PR #2512, opened 2024-02-13
Separate read and write storage backend for mirror

masih:
Allow district read side and write side configuration for ad mirror.

This enables a storage backend to be passively populated from a retrieval backend since ingest will attempt to store all processed ads regardless of their retrieval origin.


filestore/s3.go
```diff
@@ -115,6 +121,7 @@ func (s *S3) Get(ctx context.Context, relPath string) (*File, io.ReadCloser, err
 		file.Modified = *rsp.LastModified
 	}
 
+	s3logger.Debugw("Successfully performed GET", "key", relPath)
```
gammazero, line 124: I think this one is unnecessary.

filestore/s3.go
```diff
@@ -149,7 +160,7 @@ func (s *S3) Head(ctx context.Context, relPath string) (*File, error) {
 	if rsp.LastModified != nil {
 		file.Modified = *rsp.LastModified
 	}
-
+	s3logger.Debugw("Successfully performed HEAD", "key", relPath)
```
gammazero, line 163: Not needed IMO.

gammazero (approved):

filestore/s3.go
masih: Without these we have no visibility into s3 success path. They are already at debug level which is disabled at runtime in favour of info level logs. I see no harm in having these considering the other detailed debug level logs.

gammazero: That's fine, especially since there is no run-time indicator if s3 writing is enabled or not. Generally, I avoid logging in the success path as long as non-success paths have logging or error returns, so this was those reflexes.

gammazero (approved):

PR #2526, opened 2024-02-20
Allow configuration of per-provider publishers policy

gammazero:
This allows each provider to have its own policy for which publishers are allowed to publish advertisements on behalf of that provider. The previous publisher policy serves as the default when there is no policy configured for a specific provider.

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -47,6 +47,7 @@ type PublishersPolicy struct {
 // NewPolicy returns Policy with values set to their defaults.
 func NewPolicy() Policy {
 	return Policy{
-		Allow: true,
+		Allow:   true,
+		Publish: true,
 	}
 }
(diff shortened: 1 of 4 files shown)


config/policy.go
```diff
@@ -21,22 +21,32 @@ type Policy struct {
...
+	// Except is a list of peer IDs that are exceptions to the Allow
+	// policy.
+	Except []string
 }
 
 // NewPolicy returns Policy with values set to their defaults.
 func NewPolicy() Policy {
 	return Policy{
-		Allow:   true,
-		Publish: true,
```
gammazero, line 40: Removing `Publish: true` means that the default policy is to block publishers that are not explicitly allowed. Consider whether the default should be open, requiring work to secure, or closed, requiring work to remove restrictions.

gammazero:
Note: This leaves the default publisher policy config (`Discovery.Policy.Publish = true`) to allow any peer to publish/sign advertisements for a provider that does not have its own specific publisher policy. If we want to make the default behavior to block any publisher (with peerID different from provider), that is not explicitly allowed, then this should be changed to false.

masih (approved):

PR #2527, opened 2024-02-20
Block unknown publishers with per-provider publisher policy

gammazero:
For providers that already have a known publisher, create a publisher policy that allows only the known publisher to publish advertisements for the provider.

Depends on #2526

This is draft for now, to delay deploying it for until IPNI future is somewhat clearer.

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/config.json
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/config.json
@@ -42,7 +42,104 @@
       "Allow": true,
       "Except": null,
       "Publish": true,
-      "PublishExcept": null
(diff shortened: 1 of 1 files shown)


deploy/manifests/prod/us-east-2/tenant/storetheindex/instances/inga/config.json
```diff
@@ -42,7 +42,104 @@
       "Allow": true,
       "Except": null,
       "Publish": true,
```
gammazero, line 44: Note: ` "Publish": true,` was not changed so that new providers with different publishers would be indexed by default. We can change this to ` "Publish": false,` to block any publishers that are not explicitly allowed.

masih (approved):

PR #2531, opened 2024-02-26
Clean up old users and permissions

masih:
Remove redundant access to resources and clean up permissions.

This is in prep for phasing out the AWS environment.


masih:
Cc @gammazero @willscott  I have removed redundant access to resources. This should help flag anything that we might have missed in terms of cross project collaboration when the time comes to shut down AWS.

PR #2607, opened 2024-04-25
Scale up old dhstore in prep for making it primary

gammazero:
The current primary dhstore, where new data is written to, is nearly out of space. The oldest dhstore is only 61% full, so prepare it to become the primary.

Storage IOPS and throughput will be scaled up administratively with AWS volume management.

This also scales down dhstore-seka. dhstore-tetra will be scaled down after prod is using the new (refurbished) dhstore.

### DIFF
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-seka/deployment.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-seka/deployment.yaml
@@ -20,10 +20,10 @@ spec:
               mountPath: /data
           resources:
             limits:
-              cpu: "28"
+              cpu: "7"
               memory: 58Gi
             requests:
-              cpu: "28"
+              cpu: "7"
               memory: 58Gi
       affinity:
         nodeAffinity:
--- a/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore/pvc-gp3.yaml
+++ b/deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore/pvc-gp3.yaml
@@ -8,4 +8,8 @@ spec:
   resources:
     requests:
       storage: 16Ti
+  dataSource:
+    name: dhstore-20230427
+    kind: VolumeSnapshot
+    apiGroup: snapshot.storage.k8s.io
   storageClassName: gp3-iops5k-t300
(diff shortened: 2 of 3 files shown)


deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-seka/deployment.yaml
```diff
@@ -20,10 +20,10 @@ spec:
               mountPath: /data
           resources:
             limits:
-              cpu: "28"
+              cpu: "7"
               memory: 58Gi
```
masih, line 24: I am curious why memory is not being reduced?

deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore/pvc-gp3.yaml
```diff
@@ -8,8 +8,4 @@ spec:
   resources:
     requests:
       storage: 16Ti
-  dataSource:
-    name: dhstore-20230427
-    kind: VolumeSnapshot
-    apiGroup: snapshot.storage.k8s.io
-  storageClassName: gp3-iops5k-t300
```
masih, line 15: What's the rationale for removing this? I am not sure if pvc will be modifiable this way. If unrelated to pr i recommend restoring these changes.

masih (changes requested):
The instance type needs to be updated for the resource increase to be schdulable. Otherwise, the pod cannot run since node selection cannot be met. Take a look at the deployment which has the resources reduced and use the same instance type for the other one. 

Otherwise, LGTM. Left some questions/suggestions.

deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore-seka/deployment.yaml
gammazero: Only because others with the same CPU reduction are still at 58Gi.

deploy/manifests/prod/us-east-2/tenant/storetheindex/dhstore/pvc-gp3.yaml
gammazero: I removed it because I thought we could stop using the snapshot here, and that would allow us to delete it later. I will put it back to make sure that it does not cause a problem.

masih (approved):
I recommend changing the seka instance type to r6a now that its resource demand is reduced.

PR #2630, opened 2024-06-18
feat: decode multihash string for B58 and Hex

LexLuthr:
## Context

Currently, find only support decoding B58 multihash strings. We are adding support for Hex strings as well.

## Proposed Changes

Currently, find only support decoding B58 multihash strings. We are adding support for Hex strings as well.

## Tests

NA

## Revert Strategy

Yes

### DIFF
--- a/server/find/server.go
+++ b/server/find/server.go
@@ -184,8 +184,9 @@ func (s *Server) findMultihash(w http.ResponseWriter, r *http.Request) {
 		var hexErr error
 		m, hexErr = multihash.FromHexString(mhVar)
 		if hexErr != nil {
-			log.Errorw("error decoding multihash", "multihash", mhVar, "b58Err", err, "hexErr", hexErr)
-			httpserver.HandleError(w, err, "find")
+			msg := "find: input is not a valid base58 or hex encoded multihash"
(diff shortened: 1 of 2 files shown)


server/find/server.go
```diff
@@ -516,3 +516,15 @@ func createExtendedProviderResult(epInfo registry.ExtendedProviderInfo, iVal ind
 		},
 	}
 }
+
+func caseMHString(s string) (multihash.Multihash, error) {
+	mh, err := multihash.FromHexString(s)
```
masih, line 521: My sense is that this diverging from the IPFS standards of always using base-58 encoded string value of mutlihash in URLs (for good reason). 

If the demand for this feature is popular enough, ipni spec update is the place to start. Otherwise my vote would be to keep it consistent and allow lookup by base-58 encoding only, similar to pretty much every IPFS api out there?

gammazero: Ideally, I agree with keeping it as base58. However there are many tools that show multihashes as hex-encoded strings, and allowing these into the API will increase interoperability, even if just pasting multihashes in to curl.

Added logging to log both decoding errors, and unit test.

masih: OK 👍 

Could we make sure spec is also updated please?

PR #2664, opened 2024-09-06
Support Ipni-Cid-Schema-Type HTTP header

### LINKED ISSUE #2662: Add HTTP header for Ad sync
As of now, all requests for CID are made on a single path /ipni/v1/ad/{CID}. The request can be for advertisement or entrylink. This basically means user need to figure out which and then try to find them. This is a waste of time and resource on provider side if they don't have the requested CID. 

To make this easier on provider implementation, can we send a HTTP header specifying this is a advertisement CID?

This will not impact any current provider implementation and would newer implementation to use this feature. Others can also upgrade if they wish. This will be limited to HTTP providers.

gammazero:
The Ipni-Cid-Schema-Type HTTP header serves as an indication to advertisement publishers what type of data is being requested and is identified by the CID. This may help some publishers more quickly lookup the data.

- Remove `loadtest synthetic` command that was never used

The `synthetic` command is not used as there is not longer a way to import the generated data. This command was originally used to populate storetheindex with artificial data for testing. Test data is readily available from providers and CAR mirrors, and can be made available by a reference index provider, so there is not a need to generate data synthetically withing a storetheindex admin command.

Other changes:
- Update go-libp2p
- Update boxo
- Update README to remove references to comands that no longer exist

Closes #2662


assigner/core/assigner_test.go
```diff
@@ -137,11 +136,10 @@ func TestAssignerAll(t *testing.T) {
 			}
 			t.Log("Publisher", peer2IDStr, "assigned to indexer", assignNum)
 			assigns = append(assigns, assignNum)
-		case <-timeout.C:
+		case <-time.After(3 * time.Second):
```
masih, line 139: What was the rationale for these changes? Would it be possible to use `require.Eventually` at the time of assertion to avoid fixed value waits?

masih (approved):

assigner/core/assigner_test.go
gammazero: The rationale for this was because now (in go1.23) timers can be GC'd even if they are still running, so `time.After` means less code and no need to stop and reset timers and worry about semantics before and after go1.23.

PR #2665, opened 2024-09-23
Stop using archived `go-homedir` repo

gammazero:
The `go-homedir` repo is unnecessary now that there is go support for getting a user's home directory.

- Add `fsutil.Expand` function to replace "~" at start of path with user's home directory
- Config calls expand in additional places to ensure that config paths with "~" are handled.


gammazero:
Replaced by #2666

PR #2671, opened 2024-10-12
Unfreeze even if freezing is disabled.

### LINKED ISSUE #2670: Unfreeze should work even if freezing is disabled
Freezing may be disabled after an indexer becomes frozen to prevent the indexer from freezing again. Unfreezing the indexer should still work even if freezing is disabled.

gammazero:
Freezing may be disabled after an indexer becomes frozen to prevent the indexer from freezing again. Unfreezing the indexer should still work even if freezing is disabled. This PR makes this possible, whereas previously an indexer could not be unfrozen if freezing was disabled.

Closes #2670

This PR also allows disk usage to be queried even when freezing is disabled.


masih (approved):
Thank you for picking this up so quickly 🙏

PR #2680, opened 2024-10-28
feat: make public the ingest and registry

alanshaw:
## Context
I wanted to run an IPNI node locally, programmatically, for testing.

## Proposed Changes
Make `ingest` and `registry` packages public.

## Tests
No code added.

## Revert Strategy
Change is safe to revert.


masih (approved):

gammazero (approved):

gammazero:
@masih, @alanshaw 
storetheindex is meant to be an application and not a library.  I recommend that the ingestion/registry code be moved into another repo that is meant as a library, so that it can be maintained as such and not treated like an application for our own use.

I think the code should either move into `ipni/go-libipni` or into another new repo `ipni/ingest`

alanshaw:
I don't have time to do that right now but it would be great if this could be merged in the mean time.

masih:
> I don't have time to do that right now but it would be great if this could be merged in the mean time.

Happy to merge it once the conflicts are resolved.

PR #2711, opened 2025-03-25
Cache delegated routing paths and add sf2 origin

masih:
Cache the delegated rouding paths in cloudfront using the default cache policy.

Update cloudfront origin to add the new sf2 endpoint.


willscott (commented):
👍

PR #2808, opened 2026-03-13
update to use prometheus metrics in app and core

gammazero:
Use of prometheus metrics is cleaner than using opencensus.

Requires: https://github.com/ipni/go-indexer-core/pull/251

There is no urgency to accept this PR and it is not blocking any other feature or functionality. This is only a suggested improvement.


internal/metrics/server.go
```diff
@@ -3,137 +3,132 @@ package metrics
...
+		FindLatency, AdIngestLatency, AdIngestErrorCount, AdIngestActive, AdIngestSuccessCount,
+		AdIngestSkippedCount, ProviderCount, EntriesSyncLatency, PercentUsage,
+		NonRemoveAdCount, RemoveAdCount,
+	)
+}
+
 // Measures
 var (
-	FindLatency          = stats.Float64("find/latency", "Time to respond to a find request", stats.UnitMilliseconds)
-	IngestChange         = stats.Int64("ingest/change", "Number of syncAdEntries started", stats.UnitDimensionless)
```
byo: This one is removed - is that intentional?

internal/metrics/server.go
```diff
@@ -3,137 +3,132 @@ package metrics
...
-	AdIngestSuccessCount = stats.Int64("ingest/adingestSuccess", "Number of successful ad ingest", stats.UnitDimensionless)
-	AdIngestSkippedCount = stats.Int64("ingest/adingestSkipped", "Number of ads skipped during ingest", stats.UnitDimensionless)
-	AdLoadError          = stats.Int64("ingest/adLoadError", "Number of times an ad failed to load", stats.UnitDimensionless)
-	ProviderCount        = stats.Int64("provider/count", "Number of known (registered) providers", stats.UnitDimensionless)
-	EntriesSyncLatency   = stats.Float64("ingest/entriessynclatency", "How long it took to sync an Ad's entries", stats.UnitMilliseconds)
-	PercentUsage         = stats.Float64("ingest/percentusage", "Percent usage of storage available in value store", stats.UnitDimensionless)
-	NonRemoveAdCount     = stats.Int64("ingest/nonremoveadcount", "Number of non-removal advertisements", stats.UnitDimensionless)
-	RemoveAdCount        = stats.Int64("ingest/removeadcount", "Number of removal advertisements", stats.UnitDimensionless)
-)
+	FindLatency = prometheus.NewGaugeVec(
```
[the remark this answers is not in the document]
byo: I think changing that to gauge will not work as expected - from prometheus point of view it will only sample that once per 30 seconds (by default?) so we would grab only an instant snapshot of some latency value - maybe as an approximation of the average latency. But in such cases I'd still see a histogram as a useful tool to watch the distribution over time and a grab specific percentile value.

internal/metrics/server.go
gammazero: Yes, I believe that is correct.

PR #2830, opened 2026-04-13
chore: Add direct origins to CloudFront configuration

byo:
In the infrastructure deployment configuration, directly reference main indexer machines instead of going through EKS instance that forwards the traffic.

## Context

Before this change, the traffic handled by cloudfront was forwarded to EKS instance. That EKS instance was doing TLS termination and forwarded the traffic to physical indexer machines. Intermediate EKS is an unnecessary hop in this setup. This PR configures cloudfront in such a way that the traffic goes directly to physical machines.

## Proposed Changes

Add new CloudFront origins that point to physical machines and change behaviors using the default machine.

## Tests

Watching crucial endpoints while deployment is in progress.

## Revert Strategy

Terraform apply old version.


willscott:
happy to try this and see if it works

byo:
> happy to try this and see if it works

All looks good after deployment, @willscott please approve if no issues found.

willscott (approved):

PR #2832, opened 2026-04-14
Backward synce

willscott:
Proof of concept from @parkan on evaluating a 'double cursor' or backward sync to allow for faster startup


parkan:
pushed a fix for transient failures that kill the backwards loop

PR #2833, opened 2026-04-15
Sub paths in local mirror

byo:
## Context

Advertisement mirroring using a local path can create a directory with huge amount of entries causing filesystem issues.

## Proposed Changes

Introduce PathSplit configuration to the local filestore settings that will instruct the system on how to spread car files into a sub-directory structure.

## Tests

- [x] Dedicated unit 
- [x] Manual tests with a local indexer instance

## Revert Strategy

Change is backwards-compatible if no additional filestore config options are added.

To revert the change after pathSplit settings are added, manual file move is required to flatten the directory structure.

### DIFF
--- a/config/mirror.go
+++ b/config/mirror.go
@@ -26,6 +26,16 @@ type Mirror struct {
 func NewMirror() Mirror {
 	return Mirror{
 		Compress: "gzip",
+		Retrieval: filestore.Config{
+			Local: filestore.LocalConfig{
+				DefaultPathSplit: []int{11, 2},
+			},
+		},
+		Storage: filestore.Config{
+			Local: filestore.LocalConfig{
+				DefaultPathSplit: []int{11, 2},
+			},
+		},
 	}
 }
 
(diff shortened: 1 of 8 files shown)


willscott (approved):
this seems like a reasonable way to configure it.

worth having a manifest at the root of a file-store explaining how that filestore is laid out, so that rather than configuring it as an option in the software, you point to a path and then the config of that path is picked up from disk?

byo:
> this seems like a reasonable way to configure it.
> 
> worth having a manifest at the root of a file-store explaining how that filestore is laid out, so that rather than configuring it as an option in the software, you point to a path and then the config of that path is picked up from disk?

@willscott:

Added local metadata file, the filestore still needs to be configured somehow so I renamed PathSplit to DefaultPathSplit. Manifest file takes precedence over that default value.

There was also a question how to initialize new local filestore using default path split - to keep it backwards compatible I'm only creating the metadata file if the directory is empty. If it has any entry then it will be treated as the legacy, flat setup.

willscott (approved):
this looks good to me

PR #2840, opened 2026-05-27
Expose carmirror

byo:

### DIFF
--- a/config/addresses.go
+++ b/config/addresses.go
@@ -19,6 +19,9 @@ type Addresses struct {
 	ReverseIndexer string
 	// NoResourceManager disables the libp2p resource manager when true.
 	NoResourceManager bool
+	// CarMirror is the CAR mirror HTTP listen address. Set to "none" to
+	// disable.
+	CarMirror string
 }
 
 // NewAddresses returns Addresses with values set to their defaults.
@@ -29,6 +32,7 @@ func NewAddresses() Addresses {
 		Ingest:         "/ip4/0.0.0.0/tcp/3001",
 		P2PAddr:        "/ip4/0.0.0.0/tcp/3003",
 		ReverseIndexer: "0.0.0.0:3004",
+		CarMirror:      "/ip4/0.0.0.0/tcp/3005",
 	}
 }
 
--- a/internal/ingest/mirror.go
+++ b/internal/ingest/mirror.go
@@ -14,9 +14,10 @@ import (
 )
 
 type adMirror struct {
-	carReader *carstore.CarReader
-	carWriter *carstore.CarWriter
-	rdWrSame  bool
+	carReader          *carstore.CarReader
+	carWriter          *carstore.CarWriter
+	rdWrSame           bool
+	exposableFilestore filestore.Interface
 }
 
 func (m adMirror) canRead() bool {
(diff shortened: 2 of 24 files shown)


internal/ingest/mirror.go
```diff
@@ -30,8 +34,50 @@ func (m adMirror) cleanupAdData(ctx context.Context, adCid cid.Cid, skipEntries
...
+			return nil, 0, err
+		}
+
+		return adBlock2, adDataSourceFallback, nil
+	}
+	if err != nil {
+		return nil, adDataSourceNone, err
+	}
+
+	// Block is writable of read and write storages are different
```
nymd: Comment is reversed; should be "if read and write storages are the same"

internal/ingest/mirror.go
```diff
@@ -30,8 +34,50 @@ func (m adMirror) cleanupAdData(ctx context.Context, adCid cid.Cid, skipEntries
...
+	if m.rdWrSame {
+		return adBlock, adDataSourceWriter, nil
+	}
+
+	return adBlock, adDataSourceReader, nil
 }
 
 func (m adMirror) write(ctx context.Context, adCid cid.Cid, skipEntries, noOoverwrite bool) (*filestore.File, error) {
```
nymd: Typo: one too many "o's" in NoOoverwrite. Not sure if you want to do a quick find/replace to fix.

nymd (approved):
GTG; changes noted are ornamental.

internal/ingest/mirror.go
byo: Updated the comment as it was outdated (prior to some in-PR refactor). But the comment is actually correct (ignoring the of -> if typo ;)). We can write the car file if it was read from a mirror other than the write storage itself. If it was read from the write storage (which would happen if the main read and write storages are same) then it would already be there.

internal/ingest/mirror.go
byo: Fixed

server/carmirror/server.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (s *Server) Start() error {
+	log.Infow("carmirror http server listening", "listen_addr", s.listener.Addr())
+	return s.server.Serve(s.listener)
+}
+
+func (s *Server) Close() error {
+	log.Info("carmirror http server shutdown")
+	return s.server.Shutdown(context.Background())
```
nymd: A stuck stream or slow download will block the daemon shutdown; maybe wrap it in a context.WithTimeout() ; defer cancel() ?

config/addresses.go
```diff
@@ -19,6 +19,9 @@ type Addresses struct {
 	ReverseIndexer string
 	// NoResourceManager disables the libp2p resource manager when true.
 	NoResourceManager bool
+	// CarMirror is the CAR mirror HTTP listen address. Set to "none" to
+	// disable. Used when Ingest.AdvertisementMirror.ExposeHTTP is true.
```
nymd: I can't find ExposeHTTP as a config knob in the commit. Comment is wrong?

nymd (changes requested):

config/addresses.go
byo: Goot catch, some outdated comment from earlier iterations.

server/carmirror/server.go
byo: Added the timeout for shutdown to car server and other servers as well (none had this set up).

Currently this is only set to default values - 30 sec on most servers, for car mirror I've set 120 sec so that there's a time to download larger CARs upon server restart.

I'll create separate PR to make shutdown timeout and other server settings (for car and other servers) configurable through env / config file.

nymd (approved):

PR #2861, opened 2026-07-28
Endpoint for advertisement id status

### LINKED ISSUE #2859: Endpoint for Advertisement ID status
Add an advertisement status endpoint that allows providers to check the processing state of a given advertisement CID, following the pattern established by the recently added provider sync status endpoint.

Proposed Endpoint

Public: GET `/sync/status/ad/{cid}`

Example response:

```
{
  "Ad": "baguqeera...",
  "Provider": "Qmdv6yNikmUWUWXufLJLRNkv6Y9sY5cmgeX5RVWA4WNMz4",
  "Status": "indexed",
  "SeenAt": "2026-07-03T21:18:55Z",
  "ProcessedAt": "2026-07-03T22:45:00Z"
}
```
Possible statuses: unknown, pending (seen but not processing), processing, indexed

If it isn't indexed yet, could it also determine how far away it is in the ad chain from the current processing item?

byo:
## Context

There's no easy way for a publisher to understand whether indexer has processed a given advertisement or not. The only exposed API deals with provider info - that exposes the last processed AD id which is OK for newest AD but in case of long chains of advertisements to process, figuring out whether a given AD has already been processed may require a linear scan of the history.

## Proposed Changes

Create a simple API (with potential to be extended in the future) exposing information about a given AD to quickly check whether is has already been processed or not.

## Tests

Dedicated unit tests to check the behavior.

## Revert Strategy

This is a change exposing API only, no internal logic was changed, no new data structures added.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -537,6 +537,51 @@ func (ing *Ingester) adAlreadyProcessed(adCid cid.Cid) (bool, bool, error) {
 	return processed, resync, nil
 }
 
+type AdState struct {
+	Processed bool
+	Resync    bool
+	Frozen    bool
+}
+
+// Indexed reports whether the advertisement's content should be considered
+// available from this indexer.
+//
+// An ad is indexed only when it was fully processed while the indexer was not
+// frozen. Ads marked for resync are treated as not indexed because a resync
+// invalidates the previous processing result until the ad is processed again.
+//
+// Note that Processed (and therefore Indexed) does not distinguish successful
+// content indexing from permanent skips such as malformed ads.
+func (s AdState) Indexed() bool {
+	return s.Processed && !s.Resync && !s.Frozen
+}
+
+// GetAdState returns the known state for an advertisement.
+//
+// Processed is true when the ad is marked fully processed (/adProcessed/ value
+// 1). Resync is true when the ad is marked for resync (value 2). Frozen is true
+// when the ad was processed while the indexer was in frozen mode (/adF/ key
+// present); frozen processing updates provider metadata but does not index
+// entry multihashes.
+func (ing *Ingester) GetAdState(adCid cid.Cid) (state AdState, err error) {
+	state.Processed, state.Resync, err = ing.adAlreadyProcessed(adCid)
+	if err != nil {
+		return state, err
+	}
+
+	_, err = ing.ds.Get(context.Background(), datastore.NewKey(adProcessedFrozenPrefix+adCid.String()))
+	switch {
+	case errors.Is(err, datastore.ErrNotFound):
+		state.Frozen = false
+	case err != nil:
+		return state, err
+	default:
+		state.Frozen = true
+	}
+
+	return state, nil
+}
+
 // MarkAdProcessed explicitly marks an advertisement as processed. This is used
 // to avoid ingesting this and previous advertisements.
 func (ing *Ingester) MarkAdProcessed(publisher peer.ID, adCid cid.Cid) error {
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -525,17 +525,16 @@ func (ing *Ingester) markAdUnprocessed(adCid cid.Cid, forResync bool) error {
 	return ing.ds.Put(context.Background(), datastore.NewKey(adProcessedPrefix+adCid.String()), data)
 }
 
-func (ing *Ingester) adAlreadyProcessed(adCid cid.Cid) (bool, bool) {
+func (ing *Ingester) adAlreadyProcessed(adCid cid.Cid) (bool, bool, error) {
 	v, err := ing.ds.Get(context.Background(), datastore.NewKey(adProcessedPrefix+adCid.String()))
-	if err != nil {
-		if err != datastore.ErrNotFound {
-			log.Errorw("Failed to read advertisement processed state from datastore", "err", err)
-		}
-		return false, false
+	if errors.Is(err, datastore.ErrNotFound) {
+		return false, false, nil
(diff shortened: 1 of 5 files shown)


internal/ingest/ingest.go
```diff
@@ -987,7 +1009,11 @@ func (ing *Ingester) processRawAdChain(ctx context.Context, syncFinished dagsync
 		// only publish Ads for one provider, but it's possible that an ad
 		// chain can include multiple providers.
 
-		processed, resync := ing.adAlreadyProcessed(c)
+		processed, resync, err := ing.adAlreadyProcessed(c)
+		if err != nil {
+			log.Errorw("Failed to read advertisement processed state from datastore", "err", err)
+			return
```
nymd: This will hang until a syncTimeout surfaces on an error. Should it also force an ing.inEvents adProcessedEvent or a break?

internal/ingest/ingest.go
```diff
@@ -525,17 +525,39 @@ func (ing *Ingester) markAdUnprocessed(adCid cid.Cid, forResync bool) error {
...
+// Note that processed does not distinguish successful indexing from permanent
+// skip (for example malformed ads); both write the processed marker.
+func (ing *Ingester) GetAdProcessedState(adCid cid.Cid) (processed, resync, frozen bool, err error) {
+	processed, resync, err = ing.adAlreadyProcessed(adCid)
+	if err != nil {
+		return false, false, false, err
+	}
+	_, err = ing.ds.Get(context.Background(), datastore.NewKey(adProcessedFrozenPrefix+adCid.String()))
+	if err != nil {
+		if err == datastore.ErrNotFound {
```
nymd: I believe this should match the above

if errors.Is(err, datastore.ErrNotFound)

server/ingest/server.go
```diff
@@ -77,6 +78,7 @@ func New(listen string, indexer indexer.Interface, ingester *ingest.Ingester, re
 	mux.HandleFunc("/announce", s.putAnnounce)
 	mux.HandleFunc("/health", s.getHealth)
 	mux.HandleFunc("/register", s.postRegisterProvider)
+	mux.HandleFunc("/sync/status/ad/", s.getAdStatus)
```
nymd, line 81: Is this conflicting with the other handler owning "/sync/status/..." ?  Maybe its better at "/ad/status/" ?

server/ingest/server.go
```diff
@@ -140,6 +142,58 @@ func (s *Server) getHealth(w http.ResponseWriter, r *http.Request) {
...
 }
 
+const (
+	adStatusUnknown = "unknown"
+	adStatusIndexed = "indexed"
+)
+
+type adStatusResponse struct {
+	Ad        string `json:"Ad"`
+	Status    string `json:"Status"`
```
nymd: Status and Processed seem currently redundant. I think we only need one of them unless we give status more data like the frozen or resync states that it knows about.

internal/ingest/ingest.go
```diff
@@ -525,17 +525,39 @@ func (ing *Ingester) markAdUnprocessed(adCid cid.Cid, forResync bool) error {
...
+
+// GetAdProcessedState returns the durable processing state for an advertisement.
+//
+// processed is true when the ad is marked fully processed (value 1). resync is
+// true when the ad is marked for resync (value 2). frozen is true when the ad
+// was processed while the indexer was in frozen mode (/adF/ key present).
+//
+// Note that processed does not distinguish successful indexing from permanent
+// skip (for example malformed ads); both write the processed marker.
+func (ing *Ingester) GetAdProcessedState(adCid cid.Cid) (processed, resync, frozen bool, err error) {
```
nymd: Probably better for future proofing to make this a result struct rather than a list.

nymd (changes requested):

server/ingest/server.go
byo: That shouldn't be an issue. Would be resolved on the kubernetes ingress level - that would select longest matching path so /sync/status/ad/ would be selected if that's the prefix (https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches)

I also want to move the /sync/status to the ingest service as IMO it fits here better which I realized working on this PR - but then it would also work as expected (https://pkg.go.dev/net/http#hdr-Precedence-ServeMux).

We can also consider renaming /sync/status -> /sync/status/provider, or use /sync/adstatus ?

server/ingest/server.go
byo: I started with more fields and removed those that are not relevant to the current cases we want to solve - for that I think I'll remove the `Status` field as the `Processed` is easier to interpret - e.g. doesn't have to know what value to compare against.

byo: Took a closer look and completely changed the result, now it will only have `Ad` and `Indexed` fields - the Indexed one is calculated from the status of the ad and it's meaning is that if it is true then this means that this ad has been fully indexed. Old "Processed" flag was misleading - e.g. if the ad was ingested in a frozen state it would be marked as "Processed" but it's CID changes would not have been persisted yet.

internal/ingest/ingest.go
byo: Done

internal/ingest/ingest.go
byo: Good catch, changed the implementation

internal/ingest/ingest.go
byo: Removed the return at this place - it's safe to ignore here as it will be re-checked later when we process ads of a specific provider. That should then end up with a message sent over to adProcessedEvent potentially causing better visibility for the error

nymd (approved):

PR #2862, opened 2026-07-29
Prevent rescan of whole history on old ad reannouncement

byo:
## Context

The scan phase of ingestion will traverse the ad chain until reaching an ad without previous ad, or until reaching the latest processed ad. This will cause issues when the scan starts at a node that is preceeding the latest processed one in the chain. The result is that the scan will continue running until the genesis ad (without previous one) is found.

For large chains this may result with a time-consuming scans that prevent newest ads of the affected providers from being ingested. In a production environment there were observed cases of such a scan taking more than 26 hours.

## Proposed Changes

Stop the scan if we reach already processed ads.

## Tests

The issue can be best exposed using split-chain situation - announcing a new AD that only partially shares the history. In such case, only the new part of the chain should be scanned and the shared part should be skipped. Without the fix, the scan would not stop until the genesis ad is reached.

## Revert Strategy

Only code change revert is needed.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -249,6 +249,20 @@ func (ing *Ingester) Skip500EntriesError(skip bool) {
 }
 
 func (ing *Ingester) generalDagsyncBlockHook(publisher peer.ID, c cid.Cid, actions dagsync.SegmentSyncActions) {
+	// If this ad is already fully processed, all older ads are too.
+	// Stop advancing without loading the ad data.
+	processed, _, err := ing.adAlreadyProcessed(c)
+	if err != nil {
+		ing.reg.EndScan(publisher, err)
+		actions.FailSync(err)
+		return
+	}
+	if processed {
+		log.Debugw("Reached already processed advertisement; stopping scan", "publisher", publisher, "adCid", c)
+		actions.SetNextSyncCid(cid.Undef)
+		return
+	}
+
 	// The only kind of block we should get by loading CIDs here should be
 	// Advertisement.
 	//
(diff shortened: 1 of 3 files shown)


internal/ingest/ingest.go
```diff
@@ -249,6 +249,19 @@ func (ing *Ingester) Skip500EntriesError(skip bool) {
 }
 
 func (ing *Ingester) generalDagsyncBlockHook(publisher peer.ID, c cid.Cid, actions dagsync.SegmentSyncActions) {
+	// If this ad is already fully processed, all older ads are too.
+	// Stop advancing without loading the ad data.
+	processed, _, err := ing.adAlreadyProcessed(c)
+	if err != nil {
+		ing.reg.EndScan(publisher, err)
+		actions.FailSync(err)
```
nymd, line 257: Is there a reason this fails the sync immediately at this point but an adAlreadyProcessed found below (line 1047) will continue on?

nymd (approved):

internal/ingest/ingest.go
byo: Announcing the error here is pretty straightforward, well-supported in the existing code. Continuing the execution if we don't know whether the ad has been processed or not would be a mistake as it could end up with the whole chain scan.

In the code on line 1047 I couldn't find a good way to propagate the error without touching too much code (while working on adding error to the adAlreadyProcessed - it was silently ignoring errors before). It's the publisher vs provider situation where the error happens when checking the publisher but error should be attached to providers. At that place I ignored the error as adAlreadyProcessed will be called again in a context of a specific provider where it can be properly propagated.

PR #2865, opened 2026-07-29
Simplify AdvertisementMirror configuration

byo:
## Context

Current mirror configuration, extended with the fallback source, seems like overly complicated.

## Proposed Changes

Change the configuration from Storage and Source to Local and External.

## Tests

Unit tests, manual tests

## Revert Strategy

Code revert


gammazero (commented):
I think this is also confusing, just in a different way.

Maybe there is a better way. How about getting rid of the 'Read' and 'Write' modes that are used in conjunction with the `Main` and `External` (previously Retrieval, Storage, and FallbackRetrieval). And instead have 0 or more ReadMirrors, order by precedence, and 0 or 1 WriteMirror.

```go
type Mirror struct {
   ReadMirrors []filestore.Config
   WriteMirror *filestore.Config
}

func (c *Mirror) CanRead() bool {
    return len(c.ReadMirrors) != 0
}

func (c *Mirror) CanWrite() bool {
    return c.WriteMirror != nil
}
```
The `Compress` config value should be inside each specific mirror's config. Maybe even consider auto-detecting compression for read-mirrors.

byo:
Thanks @gammazero for the comments :+1: 

> I think this is also confusing, just in a different way.
> 
> Maybe there is a better way. How about getting rid of the 'Read' and 'Write' modes that are used in conjunction with the `Main` and `External` (previously Retrieval, Storage, and FallbackRetrieval). And instead have 0 or more ReadMirrors, order by precedence, and 0 or 1 WriteMirror.
> 
> ```go
> type Mirror struct {
>    ReadMirrors []filestore.Config
>    WriteMirror *filestore.Config
> }
> 
> func (c *Mirror) CanRead() bool {
>     return len(c.ReadMirrors) != 0
> }
> 
> func (c *Mirror) CanWrite() bool {
>     return c.WriteMirror != nil
> }
> ```

IMO the biggest concern with this approach is that the separation of read and write mirror configs does not fully align with the logic.

I.e. in the current code in the `main` branch a local disk mirror must have separate config entries for read / write. This results in two distinct filestore instances even though those point to exactly the same data. And then there are different paths in the code depending on whether read and write mirrors are the same.

I was thinking before about just having an array of mirrors (with the requirement of all being distinct ones), each one of them being marked as either read / write or both:

```go
type Mirror struct {
  Stores []StoreConfig
}

type StoreConfig {
    filestore.Config

    Read bool
    Write bool
    Compress string
}
```

But that comes with its own problems - i.e. should we restrict to only one writer? If not should the car be written to multiple mirrors upon successful ingestion or to just one?

Plus in case of read mirrors scanning them one-by-one could actually add additional latency overhead. Originally it was that the local mirror was expected to be fast - so it would have to be checked first, then we test the fallback one - the fallback could be extended to a list of mirrors but then to avid increasing the latency all the fallback mirrors would be queried at once, and the first one to serve the data would win.

In case of a generic list of mirrors we'd have to add some priorities, constraints - so the question is whether this would be a feature worth implementing. My assumption was that there will usually be one local car mirror - so that we can have the car locally if resyncing and  we can expose this through http for other indexers, plus there would be one or more fallback car mirrors that we connect to to avoid getting the data straight from the provider. @gammazero do you see some other use-cases worth considering here as well?

> The `Compress` config value should be inside each specific mirror's config. Maybe even consider auto-detecting compression for read-mirrors.

Yup, that sounds reasonable, will add in this PR

byo:
@gammazero updated the PR so that compression is now applied per store, leaving detection for some other time, also the updatemirror command only copies data between mirrors of the same compression, conversion between compression methods would add too much to this PR

gammazero:
### Compression
Since no mirror has ever operated without compression, we can safely eliminate compression as an option and just make it always enabled, so that mirrored CAR files are always compressed.

### Multiple mirrors

Multiple read mirrors are useful when bringing a new indexer up-to-date. The old read-only read mirror is read to pull existing, and still valid, index content from. The new read-write mirror is used to save all valid content. After the new indexer is caught up, then the old mirror can be deleted, along with all the expired/removed content it contains. However, there are other ways that garbage collection can be done, so this is not critical.
 
> in the current code in the main branch a local disk mirror must have separate config entries for read / write. This results in two distinct filestore instances even though those point to exactly the same data. And then there are different paths in the code depending on whether read and write mirrors are the same.

In retrospect, I think this was a poor design choice. It would be better for a mirror to have a single filestore, and that mirror could be `WRONLY`, `RDONLY` or `RDWR`. A separate filestore would require a separate mirror.

byo:
> ### Compression
> Since no mirror has ever operated without compression, we can safely eliminate compression as an option and just make it always enabled, so that mirrored CAR files are always compressed.

I'll see if it causes too much trouble, if so then I'll remove it. Currently it seems to work fine if restrictions to mirrorupdate tools are in place. I still find it useful in the future if we want to introduce a better compression, so that the code doesn't have to be excessively refactored.

> ### Multiple mirrors
> Multiple read mirrors are useful when bringing a new indexer up-to-date. The old read-only read mirror is read to pull existing, and still valid, index content from. The new read-write mirror is used to save all valid content. After the new indexer is caught up, then the old mirror can be deleted, along with all the expired/removed content it contains. However, there are other ways that garbage collection can be done, so this is not critical.

Car mirror may also be beneficial in case of many independent indexers. E.g. in the current case: sf and sf2 could exchange car mirrors reducing the dependency on provider's quality. Once the first indexer gets the data in, all the others could sync up using its mirror.

I've extended the implementation so that there could be multiple external mirrors (selecting the fastest one that has the data).

> In retrospect, I think this was a poor design choice. It would be better for a mirror to have a single filestore, and that mirror could be `WRONLY`, `RDONLY` or `RDWR`. A separate filestore would require a separate mirror.

That's exactly how the Main mirror currently works - configuration + mode (`read` / `write` / `readwrite`), external mirrors are read-only by design.

byo:
@gammazero I've tested updated config (but limited to a single external car mirror only) on sf (current standby) and my local machine. Locally I was able to test all the use-cases: no mirror anywhere, data in local mirror, data in external mirror - and all worked as expected.

Please let me know if you still have any concerns regarding the current configuration schema.

I'd like to close this PR soon as it is now blocking a new release.

nymd (approved):
I think you two agree more than you think or at least more than you did at the start. @gammazero 's point on Aug 7 was that a mirror should be one storage device with an access mode rather than separate r/w entries and that is what Bart's MainMode implements. Both options of where the read role lives work and as this isn't user facing and is blocking release, I say we go with it as is.

Also it needs an update in config.md for this merge.

gammazero (approved):
I do not have any other use cases that need to be addressed. I thing this PR is good to merge. Mirror functionality may evolve in the future, but no need to do more than this for now.

PR #2876, opened 2026-08-07
Move sync status to indexer

byo:
## Context

The /sync/status API is currently handled through the find server where it is reflecting the ingester status. That could be problematic with future updates, the API would also become non-functional if ingester and finder are converted to separate processes.

## Proposed Changes

Move the /sync/status API to ingester server, simplify the state tracking by storing the  status directly in the ingester instead of provider registry (that is shared among multiple server instances).

## Tests

Convert existing tests for /sync/status from the find server to the ingestion server.

## Revert Strategy

Code revert.


internal/ingest/syncstatus.go
```diff
@@ -428,3 +428,101 @@ func (t *syncTracker) IncError() {
...
+	if !ok {
+		st = &syncTracker{}
+		ing.syncStatus[publisherID] = st
+	}
+	return st
+}
+
+// SyncStatusJSONFor returns a JSON snapshot of the sync status for the given
+// publisher, or nil if no sync status is tracked.
+func (ing *Ingester) SyncStatusJSONFor(publisherID peer.ID) json.RawMessage {
```
byo: Note to self - should this be in this layer of the code?  It seems to be a part of the presentation layer.

server/ingest/header.go
```diff
@@ -0,0 +1,48 @@
...
+	"net/http"
+	"strings"
+)
+
+const (
+	mediaTypeJson = "application/json"
+	mediaTypeAny  = "*/*"
+)
+
+func enableCors(w http.ResponseWriter) {
```
nymd: These two functions are duplicated; could be merged to a more top level file so they don't diverge

https://github.com/ipni/storetheindex/pull/2876/changes#diff-82ecd46351079d2b22d668eea57df3be514882f202343785cf5a2d3d943a0377L151

and in server/find/header.go (line 16-45)

nymd (approved):

server/ingest/header.go
byo: done, also generalized Accept header checking and content-type consts

PR #2886, opened 2026-08-12
2885 ad status update

### LINKED ISSUE #2885: Add more detailed status info to the ad status endpoint
The `/sync/status/ad/<adCid>` endpoint reports misleading state.

Two issues to fix:

- **False positive.** Permanently skipped ads (malformed, decode failure, missing entry chunk) are marked with the same byte as successfully indexed ones, so a rejected ad reports `Indexed: true`.
- **No unknown state.** A never-seen CID and an enqueued-but-unprocessed ad both report `Indexed: false`. An external team passed a content CID (the only CID an uploading client holds), got `Indexed: false`, and reasonably read indexed content as not indexed.

Root cause: a single boolean carrying a three-way distinction.

Fix: record skipped as a distinct persisted state, mark ads pending at admission, add a `State` field alongside `Indexed`, and reject non-advertisement CIDs with a 400 pointing at `/cid/`.

Reported by @jennijuju and the Filecoin Onchain Cloud team, 2026-08-05.

nymd:
## Context

The `/sync/status/ad/<adCid>` endpoint collapses ad state into one boolean. Permanently skipped ads (malformed, decode failure, missing entry chunk) report `Indexed: true`, and a never-seen CID is indistinguishable from an enqueued one. Callers passing content CIDs get a silently wrong answer. Reported by @jennijuju and the Filecoin Onchain Cloud team. Closes #2885.

## Proposed Changes

- New persisted marker byte `3` for processed-but-skipped ads, with a skip reason sidecar; skipped ads now report `Indexed: false`
- Pending marker written at chain admission, so enqueued ads are distinguishable from unknown, surviving restarts
- Response keeps `Indexed` unchanged and adds `State` (`unknown`, `pending`, `indexed`, `skipped`, `resyncing`), `Reason` (skipped only), and `Frozen`
- Non dag-json CIDs return 400 with a pointer to `/cid/`; nil ingester returns 503; zero-length marker values return 500 instead of panicking

Ads processed before this change cannot be distinguished from skipped and continue to report `indexed`. Pending marking adds one datastore write per admitted ad. Test helpers `blockableLinkSys` and `initIngest` changed signature, which is why some pre-existing tests appear in the diff.

## Tests

Raw marker byte table (including legacy byte `1` and undefined bytes), end-to-end permanent-error ingest writing the skipped marker and reason, skipped ads not reprocessed on later syncs, pending marker at admission with restart persistence and no resync downgrade, endpoint assertions for all five states, codec guard, empty marker 500, and nil ingester 503.

## Revert Strategy

Change is safe to revert. Persisted `3` and `0` markers read as unprocessed under the old code, so previously skipped ads would be re-attempted once and skipped again; skip reason sidecar keys are orphaned but harmless.

### DIFF
--- a/doc/ingestion.md
+++ b/doc/ingestion.md
@@ -478,7 +478,7 @@ The response includes:
     skipped ads and will report `indexed` regardless of their actual outcome
   - `"skipped"` - the ad was permanently skipped (malformed, decode error, etc.)
   - `"resyncing"` - the ad is marked for resync (previous processing invalidated)
-- `Reason` - the skip reason string, non-empty only when `State` is `"skipped"`
+- `SkipReason` - the skip reason string (truncated to 256 bytes), non-empty only when `State` is `"skipped"`
 - `Frozen` - true when the ad was processed while the indexer was in frozen mode
 
 `Indexed` is false when:
--- a/doc/ingestion.md
+++ b/doc/ingestion.md
@@ -459,12 +459,25 @@ status.
 
 Whether a single advertisement's content is available from this indexer is
 exposed on the same ingest HTTP API as `GET /sync/status/ad/<adCid>`. The
-response includes:
+endpoint only accepts dag-json CIDs (codec 0x0129); requests with other codecs
+return HTTP 400. If the ingester is not available, the endpoint returns HTTP
+503.
+
+The response includes:
 
 - `Ad` - the requested advertisement CID
 - `Indexed` - true only when the ad was fully processed while the indexer was
   not frozen, and is not currently marked for resync
   (`Processed && !Resync && !Frozen` from `AdState`)
+- `State` - one of `"unknown"`, `"pending"`, `"indexed"`, `"skipped"`, or
+  `"resyncing"`:
+  - `"unknown"` - the ad is not known to the ingester
+  - `"pending"` - the ad is known but not yet fully processed
+  - `"indexed"` - the ad was fully processed and entries were indexed
+  - `"skipped"` - the ad was permanently skipped (malformed, decode error, etc.)
+  - `"resyncing"` - the ad is marked for resync (previous processing invalidated)
+- `Reason` - the skip reason string, non-empty only when `State` is `"skipped"`
+- `Frozen` - true when the ad was processed while the indexer was in frozen mode
 
 `Indexed` is false when:
 
--- a/doc/ingestion.md
+++ b/doc/ingestion.md
@@ -321,9 +321,13 @@ Ingestion markers (`ingest.go`):
 
 - `syncPrefix` (`/sync/<publisherID>`) -> CID bytes of the latest fully
   processed ad for a publisher (`GetLatestSync` / `getLastKnownSync`).
-- `adProcessedPrefix` (`/adProcessed/<adCid>`) -> marks an ad as processed
+- `adProcessedPrefix` (`/adProcessed/<adCid>`) -> marks an ad state
   (used to stop chain walks and skip re-processing). Value is a single byte:
-  `1` = processed, `0` = explicitly unprocessed, `2` = marked for resync.
+  `0` = unprocessed, `1` = processed, `2` = marked for resync,
+  `3` = permanently skipped (will not be retried).
+- `adSkipReasonPrefix` (`/adSkipReason/<adCid>`) -> sidecar string stored
+  alongside the `3` marker, explaining why the ad was permanently skipped
+  (e.g. decoding error, malformed, entry chunk error, content not found).
 - `adProcessedFrozenPrefix` (`/adF/<adCid>`) -> marks ads processed while
   frozen, used to roll back on unfreeze (`Unfreeze` / `removeProcessedFrozen`).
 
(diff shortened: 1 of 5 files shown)


doc/ingestion.md
```diff
@@ -455,12 +459,27 @@ status.
...
+- `State` - one of `"unknown"`, `"pending"`, `"indexed"`, `"skipped"`, or
+  `"resyncing"`:
+  - `"unknown"` - the ad is not known to the ingester
+  - `"pending"` - the ad is known but not yet fully processed
+  - `"indexed"` - the ad was fully processed and entries were indexed; note that
+    ads processed before this change was deployed cannot be distinguished from
+    skipped ads and will report `indexed` regardless of their actual outcome
+  - `"skipped"` - the ad was permanently skipped (malformed, decode error, etc.)
+  - `"resyncing"` - the ad is marked for resync (previous processing invalidated)
+- `Reason` - the skip reason string, non-empty only when `State` is `"skipped"`
```
byo: Can this be `SkipReason` ?

internal/ingest/ingest.go
```diff
@@ -616,6 +653,28 @@ func (ing *Ingester) MarkAdProcessed(publisher peer.ID, adCid cid.Cid) error {
...
+}
+
+// MarkAdSkipped explicitly marks an advertisement as permanently skipped. This
+// is used to record that an ad was processed but could not be indexed due to a
+// permanent error (decode failure, malformed, entry chunk error, etc.).
+func (ing *Ingester) MarkAdSkipped(publisher peer.ID, adCid cid.Cid, reason string, frozen bool) error {
+	return ing.markAdSkipped(publisher, adCid, reason, frozen)
+}
+
+func (ing *Ingester) markAdSkipped(publisher peer.ID, adCid cid.Cid, reason string, frozen bool) error {
```
byo: Those signatures look identical and the public one only calls the private one. Let's only use the public one then.

internal/ingest/ingest.go
```diff
@@ -1258,11 +1324,13 @@ func (ing *Ingester) ingestWorkerLogic(ctx context.Context, provider, publisher
 					// error will happen. So log and drop this error.
 					log.Errorw("Skipping ad because of a permanent error", "adCid", ai.cid, "err", err, "errKind", adIngestErr.state)
 					stats.Record(context.Background(), metrics.AdIngestSkippedCount.M(1))
+					skipReason = string(adIngestErr.state)
```
byo: How about using adIngestErr.Error() ? That would contain both the state and error message

server/ingest/server.go
```diff
@@ -243,17 +246,43 @@ func (s *Server) getAdStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	resp := adStatusResponse{
-		Ad: adCid.String(),
+	if adCid.Prefix().Codec != cid.DagJSON {
+		msg := fmt.Sprintf("this endpoint expects an advertisement CID (dag-json codec), got %d; use /cid/%s for content lookups", adCid.Prefix().Codec, cidStr)
+		log.Debugw(msg, "cid", cidStr)
+		http.Error(w, msg, http.StatusBadRequest)
+		return
```
byo, line 253: Why this limitation? I recall Will mentioning that we should support DagCBOR for advertisements

server/ingest/server.go
```diff
@@ -222,6 +222,9 @@ func (s *Server) getSyncStatus(w http.ResponseWriter, r *http.Request) {
 type adStatusResponse struct {
 	Ad      string `json:"Ad"`
 	Indexed bool   `json:"Indexed"`
+	State   string `json:"State"`
+	Reason  string `json:"Reason,omitempty"`
```
byo: Let's rename to SkipReason

internal/ingest/ingest.go
```diff
@@ -1111,6 +1170,12 @@ func (ing *Ingester) processRawAdChain(ctx context.Context, syncFinished dagsync
 			ai.skip = true
 		}
 
+		if stateErr == nil && !adState.Known {
+			if err := ing.markAdUnprocessed(c, false); err != nil {
```
byo, line 1180: Technically this place seems fine, but it's only once we start processing the ad chain. I wonder whether we could mark ads already in the sync phase so that we mark ads as already noticed much earlier. Of course if not too much work or the code doesn't become too messy.

doc/ingestion.md
nymd: Yes that actually makes more sense, thanks.

internal/ingest/ingest.go
nymd: Agreed, dropping the private one. It was mirroring the MarkAdProcessed pattern which doesn't apply here.

server/ingest/server.go
nymd: Great note, thanks. The CODEC guard was in place due to a comment from Jenni about it silently failing when passed an incorrect format but not giving a real reason for the failure. I'll expand to CBOR.

byo: I'm still thinking about this - so if there's an advert with a wrong format - erroring on the API side may not be the correct place. We should in fact detect that when doing the sync phase - and that IMO should populate skip marker and the skip reason. Then from the API it would only be about fetching the information that's already there.

byo (approved):
Still some concerns about CID codec filtering, but other than that LGTM

server/ingest/server.go
nymd: That would also be preferred behaviour but I think this check is still valid just to prevent bad queries to this API from returning invisible failures.

byo: I wonder then if it wouldn't be better to use `multicodec.LookupDecoder(codec)` (e.g. as in https://github.com/ipni/go-libipni/blob/4e628c7771ce4d4863f9622aa9db829e9406d827/ingest/schema/types.go#L148) instead of a hardcoded list - just in case a new better codec is added, or someone uses some legacy one such as dag-pb

nymd: I think the check we want is "is this an advertisement" not "can this be decoded". I believe there are cases that would pass the lookup but not be an ad. If an additional codec is added in go-libipni for publishers, we can add it here in a one-line PR.

PR #2887, opened 2026-08-13
2881 batch ad status

### LINKED ISSUE #2881: Extend the /sync/status/ad API to support multiple AD IDs at once
Current API allows querying a single AD ID at a time. This could be a significant bottleneck on both the caller and the server side.

We should extend the API so that multiple IDs can be submitted at once, returning indexing status for all those IDs.

Internally this could be an optimized call to the underlying database.

There should be a limit of the number of AD IDs to query at once.

nymd:
## Context

The `/sync/status/ad/<cid>` endpoint only supports single-advertisement lookups. This closes #2881 which asks for the endpoint to accept multiple advertisement CIDs in a single call.

## Proposed Changes

- **New `POST /sync/status/ad` batch endpoint**: accepts `{"Ads": ["<cid>", ...]}` and returns `{"Statuses": [...]}` with one entry per submitted position, in request order, duplicates answered per position. Batch size is limited to 128 CIDs and the body to 1 MiB. 

- **Per-item validation**: an undecodable CID or an unsupported codec produces an entry carrying an `Error` field with `State` omitted, and the remaining entries are still answered. 

- **`GetAdStates` batch method on Ingester**: looks up state for multiple CIDs, sharing the single-ad read path exactly so the two cannot diverge, and deduplicating datastore reads for repeated CIDs.

- **Context threading**: `GetAdState`, `GetAdStates`, and `adAlreadyProcessed` now accept `context.Context`, and the HTTP handler passes the request context through, so a client disconnect cancels the datastore reads.

- **Updated `doc/ingestion.md`**: documented the batch endpoint, the limits, the per-item versus request-level error split, and the GET preference for single-ad checks.

## Tests

- `TestAdStatusBatch`: covers all five states in a single request asserted positionally against the whole response body, frozen ads, reversed order with a duplicated CID, a single-element batch, and dag-cbor accepted alongside dag-json. 

- `TestGetAdStates` and companions (`TestGetAdStatesOrderAndDedup`, `TestGetAdStatesFrozen`, `TestGetAdStatesEmpty`, `TestGetAdStatesErrorNoPartial`): prove that every batch entry equals the corresponding single `GetAdState` call, plus order preservation, duplicate handling, frozen state, empty input, and that a corrupt marker value returns an error with no partial results.

- Existing assertions in `TestAdStatus` are unchanged, which is the evidence that the single-CID endpoint behaves identically to before.

## Revert Strategy

Change is safe to revert.

### DIFF
--- a/internal/ingest/ingest_test.go
+++ b/internal/ingest/ingest_test.go
@@ -1221,7 +1221,7 @@ func TestPermanentErrorRoutesToSkippedMarker(t *testing.T) {
 	require.Equal(t, adCid, endCid)
 
 	// Verify the ad was marked as skipped with marker byte 3.
-	state, err := i.GetAdState(adCid)
+	state, err := i.GetAdState(t.Context(), adCid)
 	require.NoError(t, err)
 	require.True(t, state.Known)
 	require.True(t, state.Processed)
@@ -2413,31 +2413,31 @@ func TestGetAdState(t *testing.T) {
 	publisher := te.pubHost.ID()
 	ads := random.Cids(4)
 
-	state, err := te.ingester.GetAdState(ads[0])
+	state, err := te.ingester.GetAdState(t.Context(), ads[0])
 	require.NoError(t, err)
 	require.Equal(t, AdState{}, state)
 	require.False(t, state.Indexed())
 
 	require.NoError(t, te.ingester.MarkAdProcessed(publisher, ads[0]))
-	state, err = te.ingester.GetAdState(ads[0])
+	state, err = te.ingester.GetAdState(t.Context(), ads[0])
 	require.NoError(t, err)
 	require.Equal(t, AdState{Known: true, Processed: true}, state)
 	require.True(t, state.Indexed())
 
 	require.NoError(t, te.ingester.markAdUnprocessed(ads[1], true))
-	state, err = te.ingester.GetAdState(ads[1])
+	state, err = te.ingester.GetAdState(t.Context(), ads[1])
 	require.NoError(t, err)
 	require.Equal(t, AdState{Known: true, Resync: true}, state)
 	require.False(t, state.Indexed())
 
 	require.NoError(t, te.ingester.markAdUnprocessed(ads[2], false))
-	state, err = te.ingester.GetAdState(ads[2])
+	state, err = te.ingester.GetAdState(t.Context(), ads[2])
 	require.NoError(t, err)
 	require.Equal(t, AdState{Known: true}, state)
 	require.False(t, state.Indexed())
 
 	require.NoError(t, te.ingester.markAdProcessed(publisher, ads[3], true, false))
-	state, err = te.ingester.GetAdState(ads[3])
+	state, err = te.ingester.GetAdState(t.Context(), ads[3])
 	require.NoError(t, err)
 	require.Equal(t, AdState{Known: true, Processed: true, Frozen: true}, state)
 	require.False(t, state.Indexed())
@@ -2473,7 +2473,7 @@ func TestAdStateByMarkerByte(t *testing.T) {
 
 	// Test marker byte 0 (unprocessed) via markAdUnprocessed
 	require.NoError(t, te.ingester.markAdUnprocessed(ads[0], false))
-	state, err := te.ingester.GetAdState(ads[0])
+	state, err := te.ingester.GetAdState(t.Context(), ads[0])
 	require.NoError(t, err)
 	require.True(t, state.Known)
 	require.False(t, state.Processed)
@@ -2483,7 +2483,7 @@ func TestAdStateByMarkerByte(t *testing.T) {
 
 	// Test marker byte 1 (processed) via MarkAdProcessed
 	require.NoError(t, te.ingester.MarkAdProcessed(publisher, ads[1]))
-	state, err = te.ingester.GetAdState(ads[1])
+	state, err = te.ingester.GetAdState(t.Context(), ads[1])
 	require.NoError(t, err)
 	require.True(t, state.Known)
 	require.True(t, state.Processed)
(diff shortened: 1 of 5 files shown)


internal/ingest/ingest_test.go
```diff
@@ -2900,6 +2901,100 @@ func (vs *errorValueStore) Flush() error
...
+	require.NoError(t, err)
+	require.NotNil(t, states)
+	require.Len(t, states, 0)
+}
+
+func TestGetAdStatesErrorNoPartial(t *testing.T) {
+	te := setupTestEnv(t, false)
+
+	badAd := random.Cids(1)[0]
+	require.NoError(t, te.ingester.ds.Put(context.Background(), datastore.NewKey(adProcessedPrefix+badAd.String()), []byte{}))
```
byo: Nit:


```suggestion
	require.NoError(t, te.ingester.ds.Put(t.Context(), datastore.NewKey(adProcessedPrefix+badAd.String()), []byte{}))
```

server/ingest/protocol_test.go
```diff
@@ -327,6 +328,222 @@ func TestAdStatus(t *testing.T) {
...
+	require.NoError(t, err)
+	data, err = io.ReadAll(res.Body)
+	res.Body.Close()
+	require.NoError(t, err)
+	require.Equal(t, http.StatusBadRequest, res.StatusCode)
+	require.Contains(t, string(data), "malformed")
+
+	// Body over 1 MiB returns 400 with distinct message
+	// Send valid-looking JSON prefix + padding so decoder reads past limit before failing
+	pad := bytes.Repeat([]byte("x"), 1024*1024)
```
byo: Nit: could use a space instead of 'x' so that it is still a valid json, and actually add a single CID at the end of the "Ads" field - so that we can be sure the error is not caused by json unmarshal failure but the request size limit.

server/ingest/server.go
```diff
@@ -30,6 +30,10 @@ var log = logging.Logger("indexer/ingest")
 // clearly in error.
 const maxBodySize = 1024 * 1024
```
byo, line 31: For ad status I think this is way too large, I'd prefer to have a separate const here, 

```bash
$ python3 -c 'import json; print(len(json.dumps({"Ads":["baguqeera6m6t4ixb6sjhaetutcof4gnyh3cotrzxf2yuvdy5m6pehmzti6hq"]*128})))'
8329
```

I think 16kb would be sufficient, or we could derive the const from maxAdStatusBatch, something like:

```go
const maxAdStatusBodySize = 128*maxAdStatusBatch+64
```

server/ingest/server.go
```diff
@@ -80,6 +84,7 @@ func New(listen string, indexer indexer.Interface, ingester *ingest.Ingester, re
 	mux.HandleFunc("/register", s.postRegisterProvider)
 	mux.HandleFunc("/sync/status", s.listSyncStatus)
 	mux.HandleFunc("/sync/status/", s.getSyncStatus)
+	mux.HandleFunc("/sync/status/ad", s.batchAdStatus)
```
byo, line 96: Nit: I wonder if we could switch to registering for specific method here already (that was introduced in 1.22 so we can use it):


```suggestion
	mux.HandleFunc("POST /sync/status/ad", s.batchAdStatus)
```

The handler would be a bit cleaner then as it wouldn't have to handle method check, but I'm not sure if options work then.

server/ingest/server.go
```diff
@@ -258,36 +288,134 @@ func (s *Server) getAdStatus(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			log.Debugw("cannot decode advertisement cid in batch", "raw", raw, "err", err)
+			results[i] = adStatusResponse{
+				Ad:    raw,
+				Error: "cannot decode advertisement cid",
+			}
+			continue
+		}
+
+		if adCid.Prefix().Codec != cid.DagJSON && adCid.Prefix().Codec != cid.DagCBOR {
```
byo: This verification repeats in two places now - let's create a separate method so that it's easy to update when needed (plus we could think of using it for ingestion as well to be fully consistent)

server/ingest/server.go
```diff
@@ -258,36 +288,134 @@ func (s *Server) getAdStatus(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "ingester not available", http.StatusServiceUnavailable)
+		return
+	}
+
+	results := make([]adStatusResponse, len(req.Ads))
+	type indexedCid struct {
+		idx int
+		cid cid.Cid
+	}
+	var validCids []indexedCid
```
byo: Nit: Why not store this as a map + build the list of CIDs on the fly:

```suggestion
	validCids := make([]cid.Cid, 0, len(req.Ads))
	validCidIdx := make(map[cid.Cid]int, len(req.Ads))
```

and at the end of the loop below:

```go
validCids = append(validCids, adCid)
validCidIds[adCid] = i
```

byo (approved):
Added some comments, mostly nits (so I'm OK with what is there in the PR already), LGMT once addressed

server/ingest/server.go
```diff
@@ -258,36 +288,134 @@ func (s *Server) getAdStatus(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	bodyReader := http.MaxBytesReader(w, r.Body, maxBodySize)
+	defer r.Body.Close()
+
+	var req batchAdStatusRequest
+	if err := json.NewDecoder(bodyReader).Decode(&req); err != nil {
+		var maxErr *http.MaxBytesError
+		if errors.As(err, &maxErr) {
```
byo: One more nit: as we're not interested in the value of the error, a bit more concise vesion:


```suggestion
		if _, ok := errors.AsType[*http.MaxBytesError](err); ok {
```

server/ingest/server.go
nymd: I'd rather not touch ingestion at this point to avoid breakage while fixing a non-issue, but good point. I'm centralizing it.

PR #2891, opened 2026-08-18
build: ship cross-announce in the published image

nymd:
Builds `scripts/cross_announce` in the Dockerfile builder stage and copies the binary into the runtime image alongside `storetheindex`, so deployed cronjobs can run it from the published image instead of cloning and compiling from source on every run.

`ENTRYPOINT` and `CMD` are unchanged; `cross-announce` is run by overriding the entrypoint. Also adds a `make cross-announce` target, a `.gitignore` entry so a local build is not swept into the image build context by `COPY . .`, and a README for the tool.

Packaging and docs only. No Go source changed.

### Notes

- The `cross-announce` build uses `CGO_ENABLED=0`, which differs from the `storetheindex` build above it, so the two steps share no build cache and the dependency graph is compiled twice. Verified the tool builds without cgo.
- `distroless/cc` ships libc, so either linking mode works in the runtime image.
- The README documents that this tool has previously triggered full advertisement chain resyncs on targets, and that it must not be pointed at a production indexer without first reading the current flag set.

### Verification

- `docker build .` succeeds.
- `docker run --rm --entrypoint /usr/local/bin/cross-announce <image> --help` prints the usage text, exit 0.
- Default entrypoint still starts `storetheindex daemon`.
- `make cross-announce` produces a working binary; `make clean` removes it.
- `git diff --stat` confirms no Go source changed.

The full `go test ./...` suite was not run; the diff is packaging and docs only and cannot affect it.

### Follow-up

The usage text in `scripts/cross_announce/main.go` still points at `go run ./scripts/cross_announce/main.go`, a path that does not exist inside the image. The README notes this; correcting the string itself needs a Go change and is left to the follow-up branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,13 @@ RUN go mod download
 COPY . .
 
 RUN CGO_ENABLED=1 go build
+# CGO_ENABLED differs from the line above, so this step shares no build cache and recompiles the dependency graph.
+RUN CGO_ENABLED=0 go build -o cross-announce ./scripts/cross_announce
 
 # Debug non-root image used as base in order to provide easier administration and debugging.
 FROM gcr.io/distroless/cc:debug-nonroot
 COPY --from=builder /storetheindex/storetheindex /usr/local/bin/
+COPY --from=builder /storetheindex/cross-announce /usr/local/bin/
 
 # Default port configuration:
 #  - 3000 Finder interface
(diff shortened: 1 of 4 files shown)


Dockerfile
byo, line 1: Wouldn't it be better to build two separate docker images? That way we can keep them slim and easier to use. This I believe can still be done with a single dockerfile - with two separate targets for each of the resulting images - so that we can keep the build process mostly the same.

byo (approved):

PR #2898, opened 2026-08-25
Store and report indexing time

byo:
## Context

It would be good for the ad status API to return the time when the ad was considered indexed.

## Proposed Changes

Store the timestamp in the datastore in a compact form next to the ad processing status within the same database key.

## Tests

Dedicated unit tests for this feature, including backwards-compatibility ones.

## Revert Strategy

Just code revert, database format is backwards-compatible, the old code only checks the first byte of the datastore value.

### DIFF
--- a/internal/ingest/ingest.go
+++ b/internal/ingest/ingest.go
@@ -618,6 +618,7 @@ type AdState struct {
 	Resync      bool
 	Frozen      bool
 	IndexedTime *time.Time
+	SkippedTime *time.Time
 }
 
 // Indexed reports whether the advertisement's content should be considered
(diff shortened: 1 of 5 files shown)


internal/ingest/ingest.go
```diff
@@ -649,6 +671,14 @@ func (ing *Ingester) GetAdState(ctx context.Context, adCid cid.Cid) (state AdSta
 		state.Frozen = true
 	}
 
+	// IndexedTime is only reported for ads whose content is currently
+	// considered indexed. Frozen must be loaded first so Indexed() is
+	// accurate. Legacy 1-byte markers have ModifiedAt == 0.
+	if state.Indexed() && adState.ModifiedAt != 0 {
```
nymd: Couldn't we also store this for ads that are skipped, giving us a "SkippedTime" timestamp as well?  Is that useful?  If not, disregard.

nymd (approved):

LexLuthr (approved):
Looks good for our initial use case. Thank you!

internal/ingest/ingest.go
byo: Added, was pretty straightforward to implement

PR #2903, opened 2026-08-28
Assigner: JSON announces, Prometheus metrics, and k8s base cleanup

byo:
## Context
The assigner is being used as a live HTTP announce fan-out (empty indexer pool, `ForwardAnnounceURLs` only). Two gaps showed up in that setup:

1. HTTP `PUT /announce` was CBOR-only, while indexer ingest already accepts JSON. Publishers and tools that send `Content-Type: application/json` were rejected by the assigner.
2. There was no metrics server, so forwarded announces, decode failures, and in-memory CID-cache drops were only visible in logs. During restarts we also want `/metrics` to stay up after announce HTTP has stopped, so scrapes still work through that window.

The k8s base still pointed at an unused indexer identity env (`STORETHEINDEX_PRIV_KEY_PATH`) and did not expose the metrics port on the Service. Overlays still pin `v0.8.19` (tini/`start_assigner`); after this lands and the overlay `ref` is bumped, `patch-command.yaml` can be dropped. Config remains overlay-owned.

## Proposed Changes
- Accept JSON announce bodies on `/announce` and `/ingest/announce` the same way ingest does (`application/json`, including charset). CBOR remains the default. Request body is capped at 1 MiB.
- Add a metrics HTTP server (`Daemon.MetricsAddr`, default `/ip4/0.0.0.0/tcp/8081`, disable with `"none"`). Counters: `assigner_announce_received_total` (encoding + result), `assigner_announce_accepted_total` (passed the 64-CID cache), `assigner_announce_forwarded_total` (per target host). Metrics starts before announce HTTP and shuts down last. Shutdown errors are logged and joined with `errors.Join`.
- K8s base: drop unused `STORETHEINDEX_PRIV_KEY_PATH`, expose Service port 8081 for metrics. No bundled ConfigMap.

## Tests
- `assigner/server`: JSON, JSON+charset, invalid JSON, and CBOR announces (including received-counter increments).
- `assigner/metrics`: metrics server serves `/metrics`.
- `assigner/core`: `TestForwardAnnounce` asserts a forwarded `ok` counter.
- `go test ./assigner/server/ ./assigner/metrics/ ./assigner/core/` (including `-race` on those packages during development).

## Revert Strategy
Change is safe to revert. After revert, assigner HTTP is CBOR-only again and the metrics server is gone; bump any overlay that already scrapes `:8081` or sends JSON announces before reverting.


assigner/server/server.go
```diff
@@ -85,6 +98,14 @@ func (s *Server) Close() error {
 	return s.server.Shutdown(ctx)
 }
 
+func announceEncoding(contentType string) string {
```
nymd: This should be moved to internal/httpserver so it can be shared with server/ingest/server.go as they currently diverge in how they determine a valid JSON header.

nymd (changes requested):

assigner/server/server.go
byo: Moved the whole announcement data decoding into a shared code

nymd (approved):

PR #2911, opened 2026-09-03
ingest: Do not trust broken CAR mirror files

byo:
## Context

The ingestion path trusted a CAR file from the mirror unconditionally.
`ingestEntriesFromCar` read the first entry block, checked only that its CID
matched the advertisement's entries CID, and then indexed every remaining block
in the file without ever following the `Next` chain. A truncated, reordered, or
padded CAR was therefore accepted as a complete advertisement, silently
under-indexing the provider's data. Filling the CAR mirror surfaced this: many
files already on Main were incomplete.

While testing that, a second problem appeared. `EntryBlock.EntryChunk()`
returned `(nil, nil)` when a block was not an EntryChunk but still decoded as
generic IPLD (for example a raw blob appended to the CAR): the failed decode
error was shadowed by the successful probe decode. The caller only checked
`err` and dereferenced the nil chunk, panicking the ingest worker.

## Proposed Changes

- `carstore`: `EntryChunk()` keeps the original decode error instead of
  returning `(nil, nil)`. Only a successful HAMT probe changes the result.
- `carstore`: add `ErrUnusable`, and verify every block's CID against its bytes
  while reading a CAR. A wrong root, a first block that is not the
  advertisement, or a hash mismatch is now `ErrUnusable`.
- `internal/ingest`: entry streaming walks the chain strictly. Each block must
  be the next expected `Next` CID, and no extra blocks may follow the end of the
  chain. Anything else is `ErrUnusable`.
- Source preference is unchanged: Main first (the resync fast path), then
  External, then the publisher. What changes is that an unusable Main CAR now
  falls through to the next source instead of being indexed.
- A writable Main is always rewritten when the data came from External or the
  publisher, so a broken Main CAR gets replaced. `OverwriteMirrorOnResync`
  still governs the case where Main itself was the source.
- Docs and config comments updated: `doc/ingestion.md`, `doc/config.md`,
  `config/ingest.go`.

## Tests

- `carstore`: `TestEntryChunkRejectsNonEntryBlock` covers the panic;
  `TestVerifyCID`, `TestReadRejectsWrongFirstBlock`, and
  `TestReadRejectsAdvertisementHashMismatch` cover CID verification.
- `internal/ingest`: repairing a truncated Main CAR from the publisher; using a
  complete External CAR to rewrite a truncated Main; falling back when the
  External CAR is also incomplete; falling back when CAR entries are in the
  wrong order; falling back when a CAR contains unrelated blocks.
- `go test ./...` and `go vet ./...` pass.

## Revert Strategy

Change is safe to revert. There are no data format or config schema changes;
reverting restores the previous behavior of trusting mirror CARs.

### DIFF
--- a/carstore/carreader.go
+++ b/carstore/carreader.go
@@ -60,17 +60,18 @@ type EntryBlock struct {
 
 func (e EntryBlock) EntryChunk() (*schema.EntryChunk, error) {
 	chunk, err := decodeEntryChunk(e.Data, e.Cid)
-	if err != nil {
-		node, err := decodeIPLDNode(bytes.NewBuffer(e.Data), e.Cid.Prefix().Codec, basicnode.Prototype.Any)
-		if err != nil {
-			return nil, err
-		}
-		if isHAMT(node) {
-			return nil, ErrHAMT
-		}
-		return nil, err
+	if err == nil {
+		return chunk, nil
 	}
-	return chunk, nil
+	// Probe for HAMT so callers can distinguish that case. Keep the original
+	// decode error for any other IPLD that is not an EntryChunk; a successful
+	// generic decode must not hide that error (returning (nil, nil) panics
+	// callers that only check err).
+	node, nodeErr := decodeIPLDNode(bytes.NewBuffer(e.Data), e.Cid.Prefix().Codec, basicnode.Prototype.Any)
+	if nodeErr == nil && isHAMT(node) {
+		return nil, ErrHAMT
+	}
+	return nil, err
 }
 
 type gzipReadCloser struct {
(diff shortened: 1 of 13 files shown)


carstore/carreader.go
```diff
@@ -133,39 +134,46 @@ func (cr CarReader) Location() string {
...
+			_ = rc.Close()
+		}
+	}()
+
 	if cr.compAlg == Gzip {
-		gzr, err := gzip.NewReader(r)
+		gzr, err := gzip.NewReader(rc)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("%w: %w", ErrUnusableReasonDecompress, err)
```
byo: Note: r.Close was missing here - that's why I moved it to defer

